PR #35, opened 2014-01-12
lwt_unix: sub-second precision for stat results

gfxmonk:
This patch does make lwt_Unix.stat results unequal to the corresponding Unix.stat results, since Unix.stat still only uses the second-level fields. I'd like to fix that as well, but forking and compiling ocaml itself is harder than lwt.

(there's an open bug if anyone else wants to jump in: http://caml.inria.fr/mantis/view.php?id=6285)

**note**: The code is copied from Jane street core (https://github.com/janestreet/core/blob/master/lib/unix_stubs.c#L426). The code is under the APL, although I don't know if copyright attribution is required, since the code is driven almost entirely by maths and system APIs.

### DIFF
--- a/discover.ml
+++ b/discover.ml
@@ -259,6 +259,25 @@ CAMLprim value lwt_test(value u)
 }
 "
 
+let struct_ns_code conversion = "
+#define _GNU_SOURCE
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <caml/mlvalues.h>
+
+#define NANOSEC" ^ conversion ^ "
+
+CAMLprim value lwt_test() {
+  struct stat *buf;
+  double a, m, c;
+  a = (double)NANOSEC(buf, a);
+  m = (double)NANOSEC(buf, m);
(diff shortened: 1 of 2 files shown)


vbmithr:
@gfxmonk 

Thanks, but we need checks at runtime like http://caml.inria.fr/mantis/view.php?id=6285 in order to merge this.

gfxmonk:
@vbmithr I don't follow. Do you mean a compile-time check? I can probably just port the final ocaml patch (https://github.com/ocaml/ocaml/pull/148/files) to the lwt codebase, if that's what's needed?

gasche:
I think @vbmithr was just asking for a configure-time test that the feature is present, as done in Jérémie's patch.

vbmithr:
Yes exactly. Because it seems it is not so simple to know what system uses what, and cover all the systems, so a configure-time check is the most sensible way to do this.

gfxmonk:
I've now made this a compile-time check now, and rebased against current `master`.

PR #67, opened 2014-05-22
Lwt_queue, for blocking message passing

c-cube:
I had this piece of code in one of my repositories, and today I heard someone else complain that there lacks some way of communicating between long-running `Lwt` threads (unless I missed something). In case it might be useful, here is a very simple blocking queue (for mailboxes, message passing between threads...).

The semantics of `push` is that it might block if the queue has a limited capacity and is full.

### DIFF
--- a/src/core/lwt_queue.mli
+++ b/src/core/lwt_queue.mli
@@ -1,0 +1,42 @@
+(* Lightweight thread library for OCaml
+ * http://www.ocsigen.org/lwt
+ * Module Lwt_stream
+ * Copyright (C) 2014 Simon Cruanes
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, with linking exceptions;
+ * either version 2.1 of the License, or (at your option) any later
+ * version. See COPYING file for details.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


hcarty:
I also have uses for this or @c-cube's suggestion to add a blocking version of `Lwt_stream.bounded_push`.  In most or all of the instances where I've wanted something like this I've had a fixed/limited set of threads which need to push so the extra allocation isn't a problem.

src/core/lwt_queue.mli
```diff
@@ -0,0 +1,42 @@
...
+(** {1 Blocking queue}
+This queue can be used for message passing between threads *)
+
+type 'a t
+(** Queue containing elements of type 'a *)
+
+val create : unit -> 'a t
+(** New queue with unbounded capacity *)
+
+val create_bounded : int -> 'a t
```
hcarty: Should the bound be an optional argument rather than a separate function?

    val create : ?max:int -> unit -> 'a t

hcarty:
A `get_bound` or `get_max` function may be useful.

    val get_max : 'a t -> int option

rgrinberg:
@c-cube Could you please create an opam package for this module? I'd like to use it as well.

c-cube:
@rgrinberg this is a bit old now; I've been playing with a more powerful interface (or so I hope) in [containers.lwt.Lwt_pipe](https://github.com/c-cube/ocaml-containers/blob/7bacac2c986c462b02a083436b88ba38afd17f99/src/lwt/lwt_pipe.mli). It's not heavily tested though.

If you want, I can move this to a separate opam package.

PR #107, opened 2014-12-19
Deprecated functions; will be removed

whitequark:


zoggy:
What should be used instead of Lwt_unix.run, then ?

avsm:
Lwt_main.run

whitequark:
Will this be merged?

aantron:
Marked these deprecated more clearly. Can't use `[@@ocaml.deprecated]` because
Lwt still supports 4.01. Also marked `Lwt_stream.on_terminate` and `Lwt_unix.execute_job`. `Lwt_io.of_string` was removed in 6049279b3c1d0ecc14220876a522916c3572e1b9.

At least the following packages are still using `Lwt_unix.run` (some in tests
and examples, but nonetheless):

> `bistro`, `cowabloga`, `csvprovider`, `dog`, `git`, `git-unix`, `http2https`,
`imaplet-lwt`, `iocaml`, `irmin`, `irmin-unix`, `mbr-format`,
`merge-queues`, `merge-ropes`, `message-switch`, `mirage-git`, `mirage-irmin`,
`mirage-profile`, `mirage-unix`, `ppx_json_types`, `riak`, `sqlexpr`, `tuntap`,
`typerex-lldb`

Eliom recommends `Lwt_unix.run` in its documentation
(`doc/manual-wiki/server-services.wiki`).

At least the following packages are using `Lwt_stream.on_terminate`:

> `cohttp`

At least the following packages are using `Lwt_unix.execute_job`:

> `baardseeker`, `lambda-term`

No public package is apparently using `Lwt_unix.has_wait4` or `Lwt_sys.windows`,
but there may be private code.

attn @pveber @avsm @rgrinberg @samoht @nv-vn @vasilisp @gregtatcam @andrewray
@talex5 @djs55 @mfp @lefessan

aantron:
attn @toolslive

rgrinberg:
@aantron What should we use instead of `Lwt_stream.on_terminate`

aantron:
`Lwt_stream.on_termination` :)

metadave:
@aantron I appreciate the ping! I've only used this in the Riak client, which I haven't touched in years.

nv-vn:
Thanks for pinging, I've gone ahead and patched that in all projects that were using them.

toolslive:
concerning `Lwt_unix.execute_job`. I don't see this pull request removing it. 
Am I missing something ?

As an aside, I recently worked on the ordma package and needed to implement a new Lwt engine, based on librdmacm's rselect. There were some problems:
- Lwt checks for readability (`Lwt_unix.readable`) on a `Lwt.file_descr`  _on the side_ using a separate call to poll [here](https://github.com/ocsigen/lwt/blob/master/src/unix/lwt_unix_unix.c#L40), iso going through the engine. This breaks down if the resource represented by this `Lwt.file_descr` is not a resource that can be `poll`-ed (in this case an rsocket).

- `Lwt.file.descr` has some internal state (the state attribute )that cannot be manipulated through the interface. This forced me into using a private copy the definition, and a "cast" via `Obj.magic`, 
which is really ugly. A typical example is the [close](https://github.com/toolslive/ordma/blob/master/lwt_rsocket.ml#L109) call.

Basically, what I'm saying is that there is an API to plug in your own lwt engine, but that it doesn't really work when you actually try to do this (or I wasn't smart enough to figure out how to do this cleanly).

aantron:
@toolslive This PR didn't originally seek to remove `execute_job`, but it does now. A commit to do this hasn't been posted yet, because it's too early at this point.

> As an aside, ...

Would you be willing to turn this into a separate issue? Copy/paste from here is fine, but reproduction cases and/or patches are also very welcome :)

toolslive:
the aside is continued [here](https://github.com/ocsigen/lwt/issues/237)

aantron:
Thanks. [`Lwt_unix.run_job`](http://ocsigen.org/lwt/2.5.1/api/Lwt_unix#VALrun_job) is the alternative.

aantron:
Now using deprecation warnings (`[@ocaml.deprecated]`). See also #293 ("Semantic versioning; safely breaking Lwt").

PR #117, opened 2015-01-13
additional dummy functions for windows

fdopen:
The pre-processor was replaced with runtime detection.  Now you have to define a lot of functions, that are never used on windows, just to make the linker happy. Not sure, if I've got them all.


vouillon:
Thanks!

PR #118, opened 2015-01-13
use Bytes.of_string in tests

fdopen:
the test cases don't compile out of the box without this,...


vouillon:
Thanks!

This fixes #124.

PR #122, opened 2015-02-11
Add ghost locations to the ppx.

Drup:
Requested by @def-lkb to help merlin.

Wait for his approval before merging.

### DIFF
--- a/ppx/ppx_lwt_ex.ml
+++ b/ppx/ppx_lwt_ex.ml
@@ -13,6 +13,11 @@ let with_loc f { txt ; loc } =
 let def_loc txt =
   { txt; loc = !default_loc }
 
+(** All generated name should be marked with ghost location.
+    It helps merlin, among other things.
+*)
+let ghost t = {t with Location.loc_ghost = true}
+
 (** {3 Internal names} *)
 
 let lwt_prefix = "__ppx_lwt_"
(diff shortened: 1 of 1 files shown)


Drup:
@whitequark : you might want to add it everywhere too.

def-lkb:
I need a few more to check it. Slightly too many expressions are made ghost by the patch in its current state, and merlin is not yet interpreting all locations as it should.

ppx/ppx_lwt_ex.ml
```diff
@@ -337,18 +344,18 @@ let lwt_mapper args =
         else
           default_mapper.expr mapper expr
       | { pexp_desc = Pexp_apply (fn, args); pexp_attributes; pexp_loc } when !log ->
-        default_loc := pexp_loc;
+        default_loc := ghost pexp_loc;
```
def-lkb: It's mapping an expression from the input tree, it is a bit too early to set it ghost.

Drup:
Fixed, other issues ?

vbmithr:
@drup @def-lkb Happy with this ? Should we merge it ?

Drup:
It's not ready and it's not important enough to delay the relase.

vbmithr:
@drup: what is the state of this now?

Drup:
I don't know, @def-lkb ?

def-lkb:
Good to me.

def-lkb:
(sorry for previous comment)
Good to me. I updated merlin to handle these cases better (by interpreting ghost_loc differently).

Unless you encountered buggy cases with merlin 2.3.1 when using lwt-ppx, this can be closed.

PR #123, opened 2015-02-17
add reset_rules to logging core

toolslive:
saw a typo too.


src/logger/lwt_log_core.ml
```diff
@@ -208,6 +208,9 @@ let append_rule pattern level =
   rules := !rules @ [(split pattern, level)];
   Section.recompute_levels ()
 
+let reset_rules () =
+  rules := []
```
Drup: I think you should add `Section.recompute_levels ()`.

PR #125, opened 2015-02-25
Avoids problem of multiple definitions winsock.

modlfo:
See http://stackoverflow.com/questions/1372480/c-redefinition-header-files-winsock2-h

### DIFF
--- a/src/unix/lwt_process_stubs.c
+++ b/src/unix/lwt_process_stubs.c
@@ -23,7 +23,7 @@
 #include <lwt_config.h>
 
 #if defined(LWT_ON_WINDOWS)
-
+#define _WINSOCKAPI_
 #include <windows.h>
 
 #include <lwt_unix.h>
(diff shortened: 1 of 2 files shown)


Drup:
Can someone who know its way in the C bindings take a look ?

fdopen:
Why not remove the the inclusion of 'winsock2.h' and 'windows.h' altogether?
`<caml/unixsupport.h>` is included later through `lwt_unix.h` (in both files).
And `<caml/unixsupport.h>` already includes everything neccessary (in the right order, see https://github.com/ocaml/ocaml/blob/trunk/otherlibs/win32unix/unixsupport.h).

If you include it before `<caml/unixsupport.h>`, the inclusions and defintions there don't have the intended effect.

diml:
Indeed, now you might be able to just remove `winsock2.h`. It was included in lwt stubs as `unixsupport.h` didn't include it before OCaml 4.01. `windows.h` is not included in the latter thought, so you might want to check it still builds fine before removing it. If you remove `winsock2.h` but keep `windows.h` you might need to move it after `unixsupport.h` to preserve the order.

vouillon:
Thanks! This is fixed by #129.

Issue #136, opened 2015-03-15
segfault in `getservbyname`?


samoht:
While running some stress tests for cohttp, I got a segfault on OSX 10.10.2:

```
* thread #1: tid = 0x10f742, 0x00000001001c06b9 foo`caml_alloc_array + 99, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00000001001c06b9 foo`caml_alloc_array + 99
foo`caml_alloc_array + 99:
-> 0x1001c06b9:  cmpq   $0x0, 0x10(%r13,%rbx,8)
   0x1001c06bf:  leaq   0x1(%rbx), %rbx
   0x1001c06c3:  jne    0x1001c06b9               ; caml_alloc_array + 99
   0x1001c06c5:  cmpq   $-0x1, %rbx
(lldb) bt
* thread #1: tid = 0x10f742, 0x00000001001c06b9 foo`caml_alloc_array + 99, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00000001001c06b9 foo`caml_alloc_array + 99
    frame #1: 0x00000001001b1c7e foo`alloc_servent + 126
    frame #2: 0x00000001001af099 foo`result_getservbyname + 57
    frame #3: 0x00000001001b0f01 foo`lwt_unix_self_result + 49
    frame #4: 0x00000001001024b8 foo`.L573 + 20
```

samoht:
I've also have some other weird errors on that stress test:

```
* thread #1: tid = 0x110b88, 0x00007fff904b0c22 libsystem_kernel.dylib`write + 10, queue = 'com.apple.main-thread', stop reason = signal SIGPIPE
    frame #0: 0x00007fff904b0c22 libsystem_kernel.dylib`write + 10
libsystem_kernel.dylib`write + 10:
-> 0x7fff904b0c22:  jae    0x7fff904b0c2c            ; write + 20
   0x7fff904b0c24:  movq   %rax, %rdi
   0x7fff904b0c27:  jmp    0x7fff904aac78            ; cerror
   0x7fff904b0c2c:  retq
(lldb) bt
* thread #1: tid = 0x110b88, 0x00007fff904b0c22 libsystem_kernel.dylib`write + 10, queue = 'com.apple.main-thread', stop reason = signal SIGPIPE
[...]

PR #139, opened 2015-03-27
Fix issue #136

fdopen:
See discussion at https://github.com/ocsigen/lwt/issues/136

- The test code for netdb_reentrant_code was too weak. E.g. it compiled on netbsd, although 'gethostbyname_r' etc. aren't available in the necessary form (link information is available for backward compatiblity, but the prototype is not defined in any header). I've added more type information and assign the result to a variable, so the compiler will complain.

- non-reentrant functions like gethostbyname are protected by lwt mutexes and the result is copied to the common c heap. The functions are often implemented thread-safe even without the *_r suffix (however, there are currently no #ifdefs for this case). The returned pointer could point to thread local storage.

- htons was missing in getservbyport.

### DIFF
--- a/discover.ml
+++ b/discover.ml
@@ -234,6 +234,31 @@ CAMLprim value lwt_test(value Unit)
 }
 "
 
+let hostent_reentrant_code = "
+#define _GNU_SOURCE
+#include <stddef.h>
+#include <caml/mlvalues.h>
+#include <caml/config.h>
+/* Helper functions for not re-entrant functions */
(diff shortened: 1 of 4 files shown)


samoht:
Thanks very much for the fix, it is indeed solving the segmentation fault I see when running Irmin stress-tests. Any rough idea on when this could be released?

vbmithr:
I've merged it, since you look happy with the patch. Releasing this soon.

PR #140, opened 2015-03-31
Wrong default logger documentation.

haesbaert:
Default logger is to send everything to stderr, not stdout, I believe this is a documentation error and not a code error, since logging to stderr makes more sense than stdout, especially in applications that output real data to stdout.


Drup:
The issue is that it's not exactly true: 
 - if lwt.unix is not loaded, only Lwt_log_core is present, and the default logger is null
 - if lwt.unix is loaded, then it's the one you documented.

Can you fix the doc, and I'll merge ?

haesbaert:
Ack, fixed.

Drup:
I fixed with a more specific wording.

PR #142, opened 2015-04-02
Remove temp files properly

planar:
It does not make any sense to call `Filename.basename` at this place and indeed it doesn't work.


vbmithr:
Merged, thanks!

planar:
Wait a second, it doesn't work.

PR #143, opened 2015-04-03
Lwt_unix: add support for joining/leaving multicast groups

nojb:
This patch adds bindings to be able to join and leave IPv4 multicast groups.  As far as I know currently the only libraries where this functions are bound are JS `core` and `ocamlnet`.  Both are inconvenient to use with `lwt`.

### DIFF
--- a/src/unix/lwt_unix.ml
+++ b/src/unix/lwt_unix.ml
@@ -1507,6 +1507,36 @@ let getsockopt_error ch =
   Unix.getsockopt_error ch.fd
 
 (* +-----------------------------------------------------------------+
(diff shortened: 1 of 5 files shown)


vbmithr:
Awesome, much appreciated! (at least by me ;) )

Can you  please add a test that covers this code (like the ones in `tests/unix`).
Thanks

nojb:
Added the tests in e6ae14cd07dcffffea867332f258f7cd297fa4ad.  Thanks!

vbmithr:
Amazing, thanks.

PR #144, opened 2015-04-04
Lwt_unix: lwt_unix_mcast_* not available on Windows

nojb:
This is a follow-up to #143, marking the `lwt_unix_mcast_*` primitives as not available on Windows.


vbmithr:
Merged, thanks.

PR #145, opened 2015-04-08
Lwt_io: ?buffer argument for user-supplied buffer

gdsfh:
This patch allows Lwt user to reuse buffers for Lwt_io channels instead of allocating them on each call to functions that create Lwt_io channels.
We (at Ahrefs) create tonns of short-living Lwt_io channels to serve http requests, and allocating bigarray for every channel is too heavy for GC (caml_alloc_custom -> caml_adjust_gc_speed -> ... -> major slices happen too often -> everything is slow).
Using this patch, we reuse buffers (allocating them it if none available), and our code runs much faster.
This patch can be helpful for other Lwt users too, so I send this pull request.

### DIFF
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -35,6 +35,9 @@ let check_buffer_size fun_name buffer_size =
   else
     ()
 
+let check_buffer fun_name buffer =
+  check_buffer_size fun_name (Lwt_bytes.length buffer)
+
 let default_buffer_size = ref 4096
 
 (* +-----------------------------------------------------------------+
@@ -470,21 +473,22 @@ let no_seek pos cmd =
 
 let make :
   type m.
-    ?buffer_size : int ->
+    ?buffer : Lwt_bytes.t ->
     ?close : (unit -> unit Lwt.t) ->
     ?seek : (int64 -> Unix.seek_command -> int64 Lwt.t) ->
     mode : m mode ->
     (Lwt_bytes.t -> int -> int -> int Lwt.t) ->
-    m channel = fun ?buffer_size ?(close=Lwt.return) ?(seek=no_seek) ~mode perform_io ->
-  let size =
-    match buffer_size with
+    m channel = fun ?buffer ?(close=Lwt.return) ?(seek=no_seek) ~mode perform_io ->
+  let (buffer, size) =
+    match buffer with
+      | Some buffer ->
+          check_buffer "Lwt_io.make" buffer;
+          (buffer, Lwt_bytes.length buffer)
       | None ->
-          !default_buffer_size
-      | Some size ->
-          check_buffer_size "Lwt_io.make" size;
-          size
+          let size = !default_buffer_size in
+          (Lwt_bytes.create size, size)
   in
-  let buffer = Lwt_bytes.create size and abort_waiter, abort_wakener = Lwt.wait () in
+  let abort_waiter, abort_wakener = Lwt.wait () in
   let rec ch = {
     buffer = buffer;
     length = size;
@@ -535,13 +539,13 @@ let of_bytes ~mode bytes =
   } in
   wrapper
 
-let of_fd : type m. ?buffer_size : int -> ?close : (unit -> unit Lwt.t) -> mode : m mode -> Lwt_unix.file_descr -> m channel = fun ?buffer_size ?close ~mode fd ->
+let of_fd : type m. ?buffer : Lwt_bytes.t -> ?close : (unit -> unit Lwt.t) -> mode : m mode -> Lwt_unix.file_descr -> m channel = fun ?buffer ?close ~mode fd ->
   let perform_io = match mode with
     | Input -> Lwt_bytes.read fd
     | Output -> Lwt_bytes.write fd
   in
   make
-    ?buffer_size
+    ?buffer
     ~close:(match close with
               | Some f -> f
               | None -> (fun () -> Lwt_unix.close fd))
@@ -549,8 +553,8 @@ let of_fd : type m. ?buffer_size : int -> ?close : (unit -> unit Lwt.t) -> mode 
     ~mode
     perform_io
 
-let of_unix_fd : type m. ?buffer_size : int -> ?close : (unit -> unit Lwt.t) -> mode : m mode -> Unix.file_descr -> m channel = fun ?buffer_size ?close ~mode fd ->
-  of_fd ?buffer_size ?close ~mode (Lwt_unix.of_unix_file_descr fd)
+let of_unix_fd : type m. ?buffer : Lwt_bytes.t -> ?close : (unit -> unit Lwt.t) -> mode : m mode -> Unix.file_descr -> m channel = fun ?buffer ?close ~mode fd ->
+  of_fd ?buffer ?close ~mode (Lwt_unix.of_unix_file_descr fd)
 
 let buffered : type m. m channel -> int = fun ch ->
   match ch.channel.mode with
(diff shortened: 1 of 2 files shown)


vbmithr:
Cool, looks very useful :)

Why aren't those propagated to _all_ functions using channels, like i.e. open_connection / with_connection / establish_server ?

gdsfh:
`{open,with}_connection` create two channels (so, two buffers are required).  I'll add them, like `?send_buffer` and `?recv_buffer`, if you think it would be good.  (in the current version of patch I've tried to be conservative.  btw, what do you think, do we need specific `?buffer_size`'s for send/receive buffers?  I think we don't: users who need it can use other functions, which create one channel on each call.)

As for `establish_server`, I don't see how it could work: we can't pass fixed buffers there, as we don't know how many connections will be served simultaneously.  `establish_server` should have (ideally) some argument that decides how buffers should be allocated and freed, but it seems too complex to do right, so I'd prefer to not touch `establish_server` at all.  Users that need custom buffers for servers are surely able to write their own server-function with their custom rules of buffers reuse.

vbmithr:
On 08/04/2015 11:21, Dmitry Grebeniuk wrote:
> |{open,with}_connection| create two channels (so, two buffers are
> required). I'll add them, like |?send_buffer| and |?recv_buffer|, if you
> think it would be good. (in the current version of patch I've tried to
> be conservative. btw, what do you think, do we need specific
> |?buffer_size|'s for send/receive buffers? I think we don't: users who
> need it can use other functions, which create one channel on each call.)

I am not sure. As with |establish_server|, we can consider that those
are high-level functions and that they should remain simple.

> As for |establish_server|, I don't see how it could work: we can't pass
> fixed buffers there, as we don't know how many connections will be
> served simultaneously. |establish_server| should have (ideally) some
> argument that decides how buffers should be allocated and freed, but it
> seems too complex to do right, so I'd prefer to not touch
> |establish_server| at all. Users that need custom buffers for servers
> are surely able to write their own server-function with their custom
> rules of buffers reuse.

Yes, definitely. You are right here. |establish_server| should indeed
just stay simple.


So basically I'm quite glad to accept your patch as it is.

@drup @diml @vouillon @rgrinberg @avsm @samoht @dbuenzli

any opinions ?

-- 
Vincent

fdopen:
If it's really that faster, what about a solution similar to `Lwt_io.default_buffer_size`?

The user could specify default functions for buffer allocation ( `int -> Lwt_bytes.t` ) and deallocation (`Lwt_bytes.t -> unit`, called at close). These functions are then used by `Lwt_io.make` and `Lwt_io.resize_buffer` (if ~buffer is not specified).

This way, one could use this feature without updating existing code and functions like `establish_server` are also covered.

gdsfh:
I'd prefer to not set buffer [de]allocators globally (shared mutable state, probable mistakes when using Lwt_io for more than one task (for http sockets and for files, for example)).
But if we could pass [de]allocators to functions that create channels, it would be better than my solution with `~buffer`.
My current proposal is: ``?allocation : ([ `In | `Out ] -> int -> Lwt_bytes.t) * ([ `In | `Out ] -> Lwt_bytes.t -> unit)`` with default `((fun _in_out size -> Lwt_bytes.create size), (fun _in_out _buffer -> ()))`
(``[ `In | `Out ]`` may be helpful for `Lwt_io.pipe` or like, to be able to specify which buffers Lwt_io can reuse/allocate for input and output channels separately.)
Are you ok with this proposal?

dbuenzli:
I never used that API so excuse any misunderstandings. On the PR itself having both `buffer` and `buffer_size` is needlessly complex (this already shows up in the documentation). Keep only `buffer`, if the client wants to specify a size he should provide a `buffer`. I know this will break backward compatibility but that's what you have to do at specific points in time to avoid bloated APIs that are harder to use and understand that it should be.

I certainly wouldn't do anything along like @fdopen proposes, globals like this lead to compositionality problems. Finally it seems to me that @gdsfh final proposal is overkill and less easy to use in practice, but I'm not sure I understand the scope of his proposal.

vbmithr:
On 08/04/2015 16:23, Daniel Bünzli wrote:
> I never used that API so excuse any misunderstandings. On the PR itself
> having both |buffer| and |buffer_size| is needlessly complex (this
> already shows up in the documentation). Keep only |buffer|, if the
> client wants to specify a size he should provide a |buffer|. I know this
> will break backward compatibility but that's what you have to do at
> specific points in time to avoid bloated APIs that are harder to use and
> understand that it should be.

I have the same opinion here.

> I certainly wouldn't do anything along like @fdopen
> <https://github.com/fdopen> proposes, globals like this lead to
> compositionality problems. Finally it seems to me that @gdsfh
> <https://github.com/gdsfh> final proposal is overkill and less easy to
> use in practice, but I'm not sure I understand the scope of his proposal.
> 

Yes, definitely. I think the best action here is to replace buffer_size
by buffer. Or nothing at all. But indeed _not_ trying to preserve
compatibility at the cost of clarity.

-- 
Vincent

avsm:
If you do break the API, please do bump the version number to 2.5.x to reflect this.

gdsfh:
"nothing at all" is a bad choice here, performance-wise.  Let's do something constructive.

> @gdsfh final proposal is overkill and less easy to use in practice

Really.  But it's very rarely needed, so I think it's acceptable.  (and it solves `establish_server` problem, which is nice.)

Anyway, I'll code any interface that will allow me to use preallocated buffers when I need to do it.  Just decide on it, and I'll code it with a pleasure.

vbmithr:
On 08/04/2015 16:50, Anil Madhavapeddy wrote:
> If you do break the API, please do bump the version number to 2.5.x to
> reflect this.

Yes, promise. I won't ever break an API without noticing everybody first
anymore :p

-- 
Vincent

vbmithr:
On 08/04/2015 17:04, Dmitry Grebeniuk wrote:
> "nothing at all" is a bad choice here, performance-wise. Let's do
> something constructive.

I suggested that "nothing at all" was better than something bad, that's
all. I'm not saying your feature is not useful.

>     @gdsfh <https://github.com/gdsfh> final proposal is overkill and
>     less easy to use in practice
> 
> Really. But it's very rarely needed, so I think it's acceptable. (and it
> solves |establish_server| problem, which is nice.)

I think the API should be clear, regardless of the fact that it is
rarely needed or not.

> Anyway, I'll code any interface that will allow me to use preallocated
> buffers when I need to do it. Just decide on it, and I'll code it with a
> pleasure.

You coded it already. Thanks for that. I just want to reach some
consensus here, for now as it turns out, there is me + Daniel thinking
that we should break the API and have only a "buffer" argument, and you
on the other side that likes the alleged "overkill" option. Anyone else
ideas ?


-- 
Vincent

rneswold:
>  I just want to reach some consensus here, for now as it turns out, there is me + Daniel thinking that we should break the API and have only a "buffer" argument, and you on the other side that likes the alleged "overkill" option. Anyone else ideas?

I have no problems with API changes that break code as long as the new interface prevents programming mistakes, is clearer, and/or is arguably better. Passing buffers *and* their length seem a step backwards in a language like OCaml.

ygrek:
I would vote for both the original patch and for the original+removing buffer_size, unfortunately ocaml doesn't provide [@ocaml.deprecated] for optional arguments %)

vbmithr:
Ok, please rewrite your patch removing buffer_size. It seems we reached consensus here.

gdsfh:
Good, I'll do it.  What should I do with functions that create two channels (like `Lwt_io.open_connection`), and with `Lwt_io.establish_server` (which creates channels on each request)?

vbmithr:
On 10/04/2015 18:53, Dmitry Grebeniuk wrote:
> Good, I'll do it. What should I do with functions that create two
> channels (like |Lwt_io.open_connection|), 

Replace buffer_size by two optional arguments like in_buffer/out_buffer

and with
> |Lwt_io.establish_server| (which creates channels on each request)?

Do nothing for this one. It should stay simple, to provide an idea to
the user how to implement ad-hoc/optimized version of such functionality.

-- 
Vincent

gdsfh:
Done.

vbmithr:
Merged, thanks.

Issue #98, opened 2014-11-26
GC pressure caused by channels created by `Lwt_ssl`


mfp:
`write_bytes` and `read_bytes` allocate temporary string buffers in order to perform the IO using `SSL.read` and `SSL.write`, which work with strings (and not `Lwt_bytes.t`).

A preliminary fix would be to allocate one string buffer per channel (same size as the Lwt_bytes one) and reuse it.

In the long-term, it'll be nice to have a `Ssl.write_bigarray` able to work with `Lwt_bytes.t` directly.
It would also get rid of not one but two(!) copies (Ssl mallocs a buffer & copies the data there before releasing the runtime).

Drup:
Ssl has been released with that, it seems. Can you do a PR with the fix, @mfp ?

PR #154, opened 2015-05-23
Depessimize lwt ssl

mfp:
Addresses #98.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@
 
   * API CHANGE: Functions in Lwt_io that were previously using a
     ~buffer_size argument now takes a ~buffer argument.
+  * Accept ?buffer argument in Lwt_ssl.{in,out}_channel_of_descr.
   * Fix non-reentrant function calls (#136)
   * IPv4 multicast support
   * Minor fixes + documentation improvements
--- a/opam
+++ b/opam
@@ -31,10 +31,11 @@ depopts: [
   "base-unix"
   "conf-libev"
   "camlp4"
-  "ssl" {>= "0.5.0"}
+  "ssl"
   "lablgtk"
   "react"
 ]
 conflicts: [
   "react" {< "1.0.0" }
+  "ssl" {< "0.5.0"}
 ]
--- a/opam
+++ b/opam
@@ -31,7 +31,7 @@ depopts: [
   "base-unix"
   "conf-libev"
   "camlp4"
-  "ssl"
+  "ssl" {>= "0.5.0"}
   "lablgtk"
   "react"
 ]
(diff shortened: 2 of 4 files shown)


mfp:
The builder reports a failure because it uses ssl 0.4.7, and the functions this PR builds upon were added in the recently released 0.5.0.

Note that Ssl.read_into_bigarray & friends have not been tested (but [they've been reviewed](https://github.com/savonet/ocaml-ssl/pull/15#issuecomment-65264830)).

I think that it should be safe to use `Ssl.read_into_array_blocking` and `Ssl.write_into_array_blocking`, because Lwt_ssl should be operating with non-blocking sockets. I've however used the variants that release the runtime to keep it safe. If nobody sees a reason not to do it, the code can be changed trivially to the slightly faster variants.

Drup:
If you rely on a specific version of ssl, please add a constraint in the opam file.

mfp:
Added to opam. Note that the ssl 0.5.0 PR for the OPAM repository is yet to be merged https://github.com/ocaml/opam-repository/pull/4132

vbmithr:
@drup could you give it a spin of CI ? If not, I'll check manually.

mfp:
On Thu, Jul 02, 2015 at 08:36:29AM -0700, Vincent Bernardoff wrote:
> @drup could you give it a spin of CI ? If not, I'll check manually.

FWIW, it "builds for me", and I just gave it a quick test with this:

    (* invoke as   ./t www.facebook.com   *) 
    (* ocamlfind ocamlopt -package lwt.ssl -o t t.ml -linkpkg *)

    open Lwt

    let request = "GET / HTTP/1.0\r\n\r\n"

    let run () =
      Ssl.init ();
      let addr = match
                   Unix.getaddrinfo Sys.argv.(1) "443"
                     [Unix.AI_FAMILY Unix.PF_INET; Unix.AI_PROTOCOL 6]
                 with
                   | [] -> failwith "can't resolve"
                   | { Unix.ai_addr; _ } :: _ -> ai_addr in
      let c    = Ssl.create_context Ssl.SSLv23 Ssl.Client_context in
      let sock = Lwt_unix.socket Unix.PF_INET Unix.SOCK_STREAM 0 in
        Lwt_unix.connect sock addr >>= fun () ->
        Lwt_ssl.ssl_connect sock c >>= fun ssl ->
        let ich = Lwt_ssl.in_channel_of_descr ssl in
        let och = Lwt_ssl.out_channel_of_descr ssl in
        Lwt_io.write och request >>= fun () ->
        Lwt_io.flush och >>= fun () ->
        Lwt_io.read ich >|= print_endline

    let () = Lwt_main.run (run ())


Note that running it against e.g.  www.google.com  results in a

    Ssl.Read_error Ssl.Error_syscall

exception because the other end does not perform a proper TLS shutdown. I
have verified this happens also without this PR. This probably deserves a
separate bug report, but I'm not sure where (probably ocaml-ssl, sigh).

-- 
Mauricio Fernández

PR #156, opened 2015-06-01
Disable the logging syntax extension by default in lwt.ppx.

Drup:
The logging syntax extension was not enabled by default in the camlp4 version and, more importantly, can make compilation fail for partially applied log functions. It should really not be enabled by default, it's just an optimisation.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,8 @@
   * Minor fixes + documentation improvements
   * Add support for if%lwt in ppx extension
   * Add Lwt.return_some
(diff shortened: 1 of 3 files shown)


vbmithr:
@drup Seems good to me, maybe you should remove the "no-log" option since the default is now not to log.
I would do that because it may confuse users about what is the default behaviour.

After that I think we should merge this too.

Drup:
No, I'll keep it for backward compat.

PR #157, opened 2015-06-08
Lwt_mvar.t's writers, readers need not be mutable

rgrinberg:
`writers`, `readers` fields are never assigned so they don't need to be mutable.


rgrinberg:
Bump? This is trivial.

PR #169, opened 2015-07-05
Add Lwt_unix.file_exists and Large_file version

hcarty:
I'm not sure if two versions are actually required but it seems like a safer approach since `stat` comes in two forms.

### DIFF
--- a/tests/unix/test_lwt_io_non_block.ml
+++ b/tests/unix/test_lwt_io_non_block.ml
@@ -35,12 +35,18 @@ let open_and_read_filename () =
   return ()
 
 let suite = suite "lwt_io non blocking io" [
+  test "file does not exist"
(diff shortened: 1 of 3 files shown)


tests/unix/test_lwt_io_non_block.ml
```diff
@@ -35,6 +35,9 @@ let open_and_read_filename () =
   return ()
 
 let suite = suite "lwt_io non blocking io" [
+  test "file does not exist"
+    (fun () -> Lwt_unix.file_exists test_file >>= fun r -> Lwt.return (not r));
```
Drup: You can use `>|=`

hcarty: Changed, thanks.

PR #185, opened 2015-09-25
Document opam package `conf-libev`

smondet:
I tested the wiki syntax against
<http://ocsigen.org/js_of_ocaml/2.6/files/wiki/index.html>.


Drup:
I feel like you should fix `conf-libev`'s depext instead, and just advise to use opam depext.

aantron:
What is wrong with `conf-libev`'s `depext` (besides missing `macports`)?

aantron:
Merged as b78389441760fbbd69787625e8c69afe431fcab0, thank you!

Issue #179, opened 2015-08-21
Could not compile lwt 2.4.5 on cygwin


Chimrod:
Hello,

I've got an error when compiling lwt on cygwin (installation done with opam).

The trace is here :

```
/home/dailly/.opam/4.02.1/bin/ocamlfind ocamlopt -g -I src/unix -linkpkg -package bigarray -package camlp4 -package camlp4.extend -package camlp4.quotations.o -package unix syntax/lwt-syntax.cmxa src/core/lwt.cmxa src/logger/lwt-log.cmxa src/unix/lwt-unix.cmxa examples/unix/logging.cmx -o examples/unix/logging.native
** Cannot resolve symbols for src/unix/liblwt-unix_stubs.a(lwt_unix_stubs.o/
lwt_libev_stubs.o/
lwt_process_stubs.o/
lwt_unix_job_access.o/
lwt_unix_job_chdir.o/
lwt_unix_job_chmod.o/
lwt_unix_job_chown.o/
lwt_unix_job_chroot.o/
lwt_unix_job_close.o/
lwt_unix_job_fchmod.o/
lwt_unix_job_fchown.o/
lwt_unix_job_fdatasync.o/
lwt_unix_job_fsync.o/
lwt_unix_job_ftruncate.o/
lwt_unix_job_link.o/
lwt_unix_job_lseek.o/
lwt_unix_job_mkdir.o/
lwt_unix_job_mkfifo.o/
lwt_unix_job_rename.o/
lwt_unix_job_rmdir.o/
lwt_unix_job_symlink.o/
lwt_unix_job_tcdrain.o/
lwt_unix_job_tcflow.o/
lwt_unix_job_tcflush.o/
lwt_unix_job_tcsendbreak.o/
lwt_unix_job_truncate.o/
lwt_unix_job_unlink.o/
):
 _mincore
File "caml_startup", line 1:
Error: Error during linking
Command exited with code 2.
```

Chimrod:
I've referenced lwt 2.4.5, but the same issue appear with 2.5.0 :

```
[...]

PR #190, opened 2015-11-10
no mincore on cygwin as well

fdopen:
`mincore` is not available on cygwin, see https://github.com/ocsigen/lwt/issues/179


CumpsD:
Seems to compile properly on cygwin on first sight

rgrinberg:
@Drup @vouillon @vbmithr merge this perhaps? Seems like the original bug reporter tested this and is satisfied.

PR #201, opened 2015-12-20
Lwt_stream: bounded_push#close needs to wake reader

edwintorok:
If at the time the `bounded_push#close` method is called the stream is empty then any attempt to process the entire stream will never finish (`Lwt_stream.iter`, `Lwt_stream.map`, etc.).
Calling the `close` method should wake any waiting threads.

See this test, it prints just `test`, but never prints `stream done`.
```ocaml
open Lwt;;

let () =
  Lwt_main.run (
    let stream, push = Lwt_stream.create_bounded 4 in
    Lwt.async( fun () ->
        Lwt_stream.iter (fun s -> print_endline s) stream >>= fun () ->
        print_endline "stream done";
        flush stdout;
        return ()
      );
    push#push "test1" >>= fun () ->
    push#close;
    Lwt_unix.sleep 2. >>= fun () ->
    return ()
  )
```

After the patch it prints:
```
test1
stream done
```

The patch also includes a testcase 'create_bounded close' that fails without the patch to `lwt_stream.ml` and passes otherwise.

### DIFF
--- a/src/core/lwt_stream.ml
+++ b/src/core/lwt_stream.ml
@@ -326,6 +326,14 @@ class ['a] bounded_push_impl (info : 'a push_bounded) wakener_cell last hooks = 
         info.pushb_pending <- None;
         Lwt.wakeup_later_exn info.pushb_push_wakener Closed
       end;
+      (* Send a signal if at least one thread is waiting for a new
+         element. *)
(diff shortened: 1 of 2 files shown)


vbmithr:
Thanks.

PR #202, opened 2015-12-22
Avoid deadlock when canceling a thread blocked on a mutex

mark-christiaens:
Problem scenario:
- Thread 1 locks mutex
- Thread 2 tries to lock mutex but is blocked and moved to the Lwt_sequence.
- Now, somehow Thread 2 is canceled (and therefor unblocked).
- Thread 1 is unlocked.  The Lwt_sequence still contains the wakener of Thread 2 so the mutex code thinks Thread 2 should get the lock.
- Thread 2 never unlocks the lock since it was canceled and therefor assumes that it does not hold the lock.

To test I:
- ocaml setup.ml -configure --enable-tests
- ocaml setup.ml -build
- ocaml setup.ml -test

### DIFF
--- a/src/core/lwt_mutex.ml
+++ b/src/core/lwt_mutex.ml
@@ -19,21 +19,13 @@
  * License along with this program; if not, write to the Free Software
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA
  * 02111-1307, USA.
- *)
+*)
 
 open Lwt.Infix
 
-type t = { mutable locked : bool; mutable waiters : unit Lwt.u Lwt_sequence.t  }
+type t = { mutable locked : bool; mutable waiters : (unit Lwt.u * bool ref) Lwt_sequence.t  }
 
 let create () = { locked = false; waiters = Lwt_sequence.create () }
-
-let rec lock m =
-  if m.locked then
-    Lwt.add_task_r m.waiters
-  else begin
-    m.locked <- true;
-    Lwt.return_unit
-  end
 
 let unlock m =
   if m.locked then begin
@@ -42,12 +34,35 @@ let unlock m =
     else
       (* We do not use [Lwt.wakeup] here to avoid a stack overflow
          when unlocking a lot of threads. *)
-      Lwt.wakeup_later (Lwt_sequence.take_l m.waiters) ()
+      let wakener, is_locked_by_us = Lwt_sequence.take_l m.waiters in
+      let () = is_locked_by_us := true in
+      Lwt.wakeup_later wakener ()
   end
 
+let rec lock m =
+  if m.locked then
+    let waiter, wakener = Lwt.task () in
+    let is_locked_by_us = ref false in
+    let node = Lwt_sequence.add_r (wakener, is_locked_by_us) m.waiters in
+    let () = Lwt.on_cancel waiter (fun () -> Lwt_sequence.remove node) in
+    Lwt.catch
+      (fun () -> waiter)
+      (fun e ->
+         let () = if !is_locked_by_us then unlock m in
+         Lwt.fail e
+      )
+  else
+    let () = m.locked <- true in
+    Lwt.return ()
+
 let with_lock m f =
-  lock m >>= fun () ->
-  Lwt.finalize f (fun () -> unlock m; Lwt.return_unit)
+  Lwt.finalize
+    (fun () -> Lwt.bind (lock m) (fun () -> f ()))
+    (fun () ->
+       let () = unlock m in
+       Lwt.return_unit
+    )
 
 let is_locked m = m.locked
+
 let is_empty m = Lwt_sequence.is_empty m.waiters


mfp:
AFAICS point (4) does not hold: the wakener for the sleeping thread t2 waits on is added to the sequence with `Lwt.add_task_r`, whose documentation states:

> [add_task_r seq] creates a sleeping thread, adds its wakener to the right of [seq] and returns its waiter. When the thread is canceled, it is removed from [seq].

So the wakener should be removed from the sequence when t2 is canceled, and on completion of t1 the state of the mutex should be set to 'unlocked'.

aantron:
> ocaml setup.ml -configure --enable-tests
ocaml setup.ml -build
ocaml setup.ml -test

Does this refer to some test case(s) that is(are) missing from the attached commit?

stijn-devriendt:
@mfp: The issue is not with add_task_r. The issue is with the "wakeup_later".
When unlock is called, the task is removed from the mutex waiters queue and executed.
If we're already in a wakeup, however, then it is queued.
If the cancel happens when this task is in the queue, then this lock is not released.

This patch protects against that.

stijn-devriendt:
It looks as if this commit (https://github.com/ocsigen/lwt/commit/b30865ce18720fccbc7e82552f465b364af82edd) already fixed the underlying issue.

aantron:
Ok, I wrote a test (3fa49497cb96fd24dc659226d61770871e5a71e4) that, AFAICT, follows the sequence in the original comment. It doesn't cause a deadlock. As a basic sanity check, commenting out the [call to `cancel`](https://github.com/ocsigen/lwt/commit/3fa49497cb96fd24dc659226d61770871e5a71e4#diff-4d0184b468d21a13994e4fdd8a941735R43) does cause a deadlock, as expected.

I'm going to look at the involved code in detail, but it would be helpful if you could comment on how to reproduce this deadlock (if still possible) in the meantime. Thanks!

aantron:
> The issue is with the "wakeup_later".
When unlock is called, the task is removed from the mutex waiters queue and executed.
If we're already in a wakeup, however, then it is queued.
If the cancel happens when this task is in the queue, then this lock is not released.

When the task is removed from the waiters queue and queued in the global Lwt queue, its state is already set to `Return ()`, so it can't be canceled, so the mutex is not released – but the task doesn't stop running. This seems to be confirmed by testing – I ran a variant of the above test:

As before:

- Thread 1 locks mutex.
- Thread 2 tries to lock it, and ends up queued.

New:

- Thread 3 is a waiting thread. The following operations are performed inside a bind on it, to make sure they run "inside" a wakeup.
  - Thread 1 unlocks the mutex. Because we are already in a wakeup, this puts thread 2 into Lwt's global queue, instead of running it immediately (confirmed by print in `lwt.ml` and code not being run in thread 2).
  - Call `Lwt.cancel` on thread 2, but this does not cancel the thread. The mutex is now locked by thread 2.
- Wake thread 3 to trigger the above.

If I add an `unlock` at the end of thread 2 and wait on thread 2, it unlocks the mutex correctly, despite the attempt to cancel thread 2.

I'm not sure if this is what the behavior *should* be, but it doesn't seem completely unreasonable. It also seems like the safest way to cancel a thread waiting on a mutex is to do it while holding that mutex.

aantron:
Cleaned the code for the latter test up, see the [updated commit](https://github.com/ocsigen/lwt/commit/c52d8248a56344c927ff12aec6703a5d04dfd9a6#diff-4d0184b468d21a13994e4fdd8a941735R63).

stijn-devriendt:
I've tried doing many similar things over the course of the last 2 days, in an attempt to reproduce this issue. Like you, I was unable to reproduce this issue.
While going through the git history, I stumbled across [this one](https://github.com/ocsigen/lwt/commit/b30865ce18720fccbc7e82552f465b364af82edd) which seems to pretty much fix the underlying issue.

If you have a look at the test case, there's this snippet:

```
ignore (
        try_lwt
          t
        with _ ->
          (* Take the first thread from the queue and send it a value. *)
          wakeup (Lwt_sequence.take_l queue) 42;
          return 0
);

```
and transform this to:

```
ignore (
        take_mutex_here
        try_lwt
          do_something under mutex
        with _ ->
          unlock mutex (which performs a wakeup under the hood)
          return ()
);

```

Then it becomes clear that it can happen that the unlock does not trigger.

My last comment probably should have been a bit clearer, but this was a shortly-before-release investigation, so the in-house development process got a bit prioritized...

We're fine with this merge request being dropped.

aantron:
I plan to write more tests for `Lwt_mutex`, since it obviously has interesting corner cases and interactions. Let's keep this PR open for now as an issue. Maybe there is a problem lurking in there somewhere :) I can let you know what I find.

Are you experiencing actual symptoms from `Lwt_mutex` in your wider program(s)?

stijn-devriendt:
We used to see the deadlock mentioned above and applied the provided patch, but other than that, things are quiet as they should be. This issue only popped up under load, though. We'll definitely let you know if we see issues in the future.

PR #206, opened 2016-01-06
Make Lwt compatible with CPS.

aantron:
Adds a version of `on_any` that performs tail calls on `f` and `g`. The following test program:

``` ocaml
let repeat : int -> (unit -> unit) -> unit = fun limit k ->
  let rec loop n =
    if n >= limit then k ()
    else Lwt.on_any Lwt.return_unit (fun () -> loop (n + 1)) ignore
  in
  loop 0

let () = repeat 100000 ignore
```

Fails with a stack overflow, when compiled and run as:

```
ocamlfind opt -linkpkg -package lwt test.ml && ulimit -s 128 && ./a.out
```

This is because `Lwt.on_any` wraps the calls to `f` and `g` in exception handlers, resulting in a growing stack. Replacing `Lwt.on_any` with `Lwt.on_any_unsafe` fixes the problem.

I am working with CPS code, and this helps to interface it to Lwt. I think this can also be accomplished by exposing a function in Lwt that can tail-call a function on the completion of any given thread. I could then use that function together with a reference to get the same result. I think the solution in the current diff is more sensible, however.

### DIFF
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -439,6 +439,14 @@ val on_any : 'a t -> ('a -> unit) -> (exn -> unit) -> unit
       or [g] raises an exception it is given to
       {!async_exception_hook}. *)
 
+val on_any_unsafe : 'a t -> ('a -> unit) -> (exn -> unit) -> unit
+  (** [on_any_unsafe t f g] is the same as [on_any t f g], except that
+      the calls to [f] and [g] are tail calls.
+
+      As a consequence, [on_any_unsafe] does not attempt to catch
+      exceptions raised by [f] and [g], and therefore does not call
+      {!async_exception_hook}. *)
+
 (** Infix operators. You should open only this module. *)
 module Infix : sig
 
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -765,6 +765,22 @@ let on_any t f g =
     | Repr _ ->
         assert false
 
+let on_any_unsafe t f g =
+  match (repr t).state with
+    | Return v ->
+        f v
+    | Fail exn ->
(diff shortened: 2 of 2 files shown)


src/core/lwt.mli
```diff
@@ -439,6 +439,14 @@ val on_any : 'a t -> ('a -> unit) -> (exn -> unit) -> unit
       or [g] raises an exception it is given to
       {!async_exception_hook}. *)
 
+val on_any_unsafe : 'a t -> ('a -> unit) -> (exn -> unit) -> unit
+  (** [on_any_unsafe t f g] is the same as [on_any t f g], except that
+      the calls to [f] and [g] are tail calls.
+
+      As a consequence, no [on_any_unsafe] does not attempt to catch
```
Drup: The `no` seems superfluous.

aantron:
Oops, fixed.

I am not sure what the best name for this function is, or what to focus on in the documentation.

aantron:
FYI this results in the need for an ugly workaround in Markup.ml: http://aantron.github.io/markup.ml/Markup_lwt.html#VALensure_tail_calls

aantron:
It's possible to get tail calls of `f` and `g` in what would otherwise be `Lwt.on_any t f g` like so:

```ocaml
match Lwt.state t with
| Lwt.Return v -> f v
| Lwt.Fail exn -> g exn
| Lwt.Sleep -> Lwt.on_any t f g
  (* Can't get tail calls on a sleeping thread anyway, use Lwt.wakeup_later to
     control stack growth for sleeping threads. *)
```

So this is closed as not necessary right now.

Will add some kind of mention of this in the API documentation.

Issue #209, opened 2016-01-15
Documentation, Documentation, Documentation and backtraces


fxfactorial:
I'm having a hard time getting a backtrace from Lwt. 

The main page of Lwt's documentation http://ocsigen.org/lwt/2.5.1/manual/manual has just this

"Backtrace support

When using Lwt, exceptions are not recorded by the ocaml runtime, and so you don't get backtraces. However it is possible to get them when using the syntax extension. All you have to do is to pass the -lwt-debug switch to camlp4:

$ ocamlfind ocamlc -syntax camlp4o -package lwt.syntax \
    -ppopt -lwt-debug -linkpkg -o foo foo.ml"

Okay, but I'm using PPX which is being pushed by the community and I'm using oasis which is also pushed heavily by the community. 

1) Assume I used camlp4, where is the mention of how to turn on backtraces with oasis? 
2) If camlp4 is mentioned at all, why isn't PPX mentioned? 
3) Okay, let's go to http://ocsigen.org/lwt/2.5.1/api/Ppx_lwt

which says:

"Debug

By default, the debug mode is enabled. This means that the backtrace versions of the bind, finalize and catch functions are used, enabling proper backtraces for the Lwt exceptions.

The debug mode can be disabled with the option -no-debug."

And yet I'm using lwt.ppx and get an exception, for reasons unknown, in my  async_exception_hook of Unix.Unix_error(Unix.ENOTCONN, "shutdown", "") with no backtrace.

Also, docs have an example of write stdio "Hello, " >> write stdio "world!" but >> doesn't come up in ocp-browser or in Lwt.Infix.

I understand that documentation is hard, but some consistency in the docs would be nice. 

And since the OCaml community tells everyone to "Don't write your own makefiles, use oasis", it would be nice to have build commands including all the esoteric flags expressed in oasis as well as command line usage.

(Yes, I understand the maintainers are probably overworked or don't necessarily care about what I need out of Lwt, so I'll be looking at fixing the documentation as well.)

PR #210, opened 2016-01-18
Documentation improvements

Drup:
Should fix https://github.com/ocsigen/lwt/issues/209
cc @fxfactorial

For the oasis part of the question, I oppened https://github.com/ocaml/ocamlbuild/issues/45

### DIFF
--- a/manual/manual.wiki
+++ b/manual/manual.wiki
@@ -462,14 +462,28 @@ expr
 
 === Backtrace support ===
 
-  When using {{{Lwt}}}, exceptions are not recorded by the ocaml runtime, and so you don't
-  get backtraces. However it is possible to get them when using the syntax extension. All you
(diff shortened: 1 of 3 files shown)


aantron:
You might even want to contrast `>>` with `>>=`, since this is what is setting up the expectation in the first place. `Note that unlike [>>=], [>>] is not an OCaml value`..., or something.

Drup:
@aantron They are recorded, but it's going to be so mangled it might as well not be the case. I added a slight precision. I also added more pointers to the manual.
Your formulation for `>>` is better, thanks!

aantron:
Strange, why aren't there links to comments on outdated diffs?

aantron:
Looks good, apart from the semicolon in the message for https://github.com/ocsigen/lwt/commit/533ef6f89cad7639bf9fd4f575ccb3ae30201af8 :)

PR #215, opened 2016-02-15
When using -lpthread, modifying the library search path is harmful

dsheets:
When using -lpthread, setting the library search path is not required and
results in general dysfunction for later library lookups. For instance, on
OS X, Apple ships a very old version of libffi but most users install a
modern libffi via homebrew. Due to a longstanding OCaml compiler front-end
bug (mantis report forthcoming), the ordering of cmxa and -cclib arguments
is not preserved. This results in the OS-supplied libffi being selected
instead of the homebrew installed libffi despite downstream developers'
best efforts. Either pthread will be in the default library search path
in a system directory with low precedence (e.g. /usr/lib) or the developer
can explicitly control the library search path order downstream to correctly
determine precedence between multiple shared library dependencies. Given
this situation, I believe it is incorrect to ever specify -L flags in cmxa
files. I have not modified the libev dependency lookup but will if desired.

/cc @yallop

### DIFF
--- a/discover.ml
+++ b/discover.ml
@@ -557,13 +557,7 @@ let () =
   let test_pthread () =
     let opt, lib =
       if !android_target then ([], []) else
-      lib_flags "PTHREAD"
-        (fun () ->
-          match search_header "pthread.h" with
-            | Some (dir_i, dir_l) ->
-                (["-I" ^ dir_i], ["-L" ^ dir_l; "-lpthread"])
-            | None ->
-                ([], ["-lpthread"]))
+      lib_flags "PTHREAD" (fun () -> ([], ["-lpthread"]))
     in
     setup_data := ("pthread_opt", opt) :: ("pthread_lib", lib) :: !setup_data;
     test_code (opt, lib) pthread_code


dsheets:
The mantis issue is [7150](http://caml.inria.fr/mantis/view.php?id=7150).

Drup:
I have no idea if the patch is correct and I don't have the capacity to test. Can anyone comment ?

vasilisp:
It looks correct with respect to how the `-l` flag should behave in a modern, properly set-up toolchain. The compiler knows where to look, and you shouldn't be linking to stuff that you find in arbitrary other paths.

That said, `search_header` is not accidental; it must be catering for some system's weirdness.

diml:
I added `search_header` in 2011 to get lwt to build on Windows  and OSX (32d1a4bb19c21b9a2e444cb310269465d260a980). The list was later modified by #79. If you think it's not needed anymore you should feel free to remove it

dsheets:
I don't know about Windows but `pthread` is fine without the `-L` on OS X.

aantron:
The patch looks to be fine under regular setups with [Ubuntu](https://travis-ci.org/ocsigen/lwt/jobs/139373886#L372), [Homebew](https://travis-ci.org/ocsigen/lwt/jobs/139373887#L445), [MacPorts](https://travis-ci.org/ocsigen/lwt/jobs/139373892#L744), and [Cygwin](https://ci.appveyor.com/project/aantron/lwt/build/81/job/0wnrrxl7cvkjiko8#L350). I think I should test on MinGW, which will probably happen in some weeks. It will probably work on MinGW too, and then it can be merged.

samoht:
Any progress on this? This is making the compilation of [osx-fsevents](https://github.com/dsheets/ocaml-osx-fsevents) fails, which forces us to ship a patch version of lwt in [Docker for Mac](https://blog.docker.com/2016/07/docker-for-mac-and-windows-production-ready/). This is clearly not optimal :-)

(and thanks for the hard work on maintaining lwt!!)

yallop:
This does seem like a fairly safe change: if it does cause build problems on some platforms it's always possible to override the discovered paths by setting the `PTHREAD_CFLAGS` and `PTHREAD_LIBS` environment variables.

aantron:
Thanks, this is in. [Confirmed](https://ci.appveyor.com/project/aantron/lwt/build/129/job/n6bqu3rtv2ouoeg1) that it doesn't affect the MinGW build (as expected).

PR #219, opened 2016-02-28
Lwt_preemptive: work around Event.sync perf issue.

### LINKED ISSUE #218: Lwt_preemptive.detach forces a full major GC run every 2500 calls at most
`Lwt_preemptive` uses `Event` channels which under the hood allocate one condition variable per `Event.sync`. The parameters given to `caml_alloc_custom` when creating the condition variable indicate that at most 5000 values can be left unreclaimed. Heavy GC load ensues. 

See https://github.com/mfp/ocaml-sqlexpr/issues/13

Also, should a (performance) bug be filed against OCaml's `Event` module?

mfp:
Refer to

   http://caml.inria.fr/mantis/view.php?id=7158
   https://github.com/ocsigen/lwt/issues/218.
   https://github.com/mfp/ocaml-sqlexpr/issues/13

Fix #218.

### DIFF
--- a/src/preemptive/lwt_preemptive.ml
+++ b/src/preemptive/lwt_preemptive.ml
@@ -55,8 +55,46 @@ let threads_count = ref 0
    | Preemptive threads management                                   |
    +-----------------------------------------------------------------+ *)
 
+module CELL :
+sig
+  type 'a t
+
+  val make : unit -> 'a t
+  val get : 'a t -> 'a
+  val set : 'a t -> 'a -> unit
+end =
+struct
+  type 'a t = {
+    m  : Mutex.t;
+    cv : Condition.t;
+    mutable cell : 'a option;
+  }
+
+  let make () = { m = Mutex.create (); cv = Condition.create (); cell = None }
+
+  let get t =
+    let rec await_value t =
+      match t.cell with
+        | None ->
+            Condition.wait t.cv t.m;
+            await_value t
+        | Some v ->
+            t.cell <- None;
+            Mutex.unlock t.m;
+            v
+    in
+      Mutex.lock t.m;
+      await_value t
+
+  let set t v =
+    Mutex.lock t.m;
+    t.cell <- Some v;
+    Mutex.unlock t.m;
+    Condition.signal t.cv
+end
+
 type thread = {
-  task_channel: (int * (unit -> unit)) Event.channel;
+  task_cell: (int * (unit -> unit)) CELL.t;
   (* Channel used to communicate notification id and tasks to the
      worker thread. *)
 
@@ -76,7 +114,7 @@ let waiters : thread Lwt.u Lwt_sequence.t = Lwt_sequence.create ()
 
 (* Code executed by a worker: *)
 let rec worker_loop worker =
-  let id, task = Event.sync (Event.receive worker.task_channel) in
+  let id, task = CELL.get worker.task_cell in
   task ();
   (* If there is too much threads, exit. This can happen if the user
      decreased the maximum: *)
@@ -89,7 +127,7 @@ let rec worker_loop worker =
 let make_worker () =
   incr threads_count;
   let worker = {
-    task_channel = Event.new_channel ();
+    task_cell = CELL.make ();
     thread = Thread.self ();
     reuse = true;
   } in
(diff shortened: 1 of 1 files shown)


mfp:
Extra eyeballs appreciated, it's very late over here :-/

aantron:
@mfp: Do you have some idea, from your discussions and experience, what the impact of resolving the two Mantis issues ([1](http://caml.inria.fr/mantis/view.php?id=7158), [2](http://caml.inria.fr/mantis/view.php?id=7198)) will be? As in, is there a nice way to resolve them, such that `Event` creating all those `Condition.t`s won't be a performance issue anymore?

If performance will still be a problem, we should definitely merge this. If it won't be a problem anymore, we can still merge this, but I'm only about 75% convinced in that case – because it's a vanishing problem given the target of 4.03.1, and because it adds code. In that case, further arguments are welcome :)

aantron:
cc @hcarty since he is working on a [potential replacement or alternative](https://github.com/hcarty/mwt) to `Lwt_preemptive`.

@hcarty: would be interested about your experience/opinion on `Mwt` or `Lwt_preemptive` being subjected to the kind of usage that causes problems here.

hcarty:
I haven't run into this largely because the cases where I've used `Lwt_preemptive` have been for heavier calls. So some extra GC overhead would be lost in the noise.

That is likely to change soon though so I would be happy to see performance improvements here! My main question is if this should live in Lwt or as a separate module/library in opam, used by Lwt.

mfp:
On Tue, Jun 21, 2016 at 04:59:08PM -0700, Anton wrote:
> @mfp: Do you have some idea, from your discussions and experience, what the impact of resolving the two Mantis issues ([1](http://caml.inria.fr/mantis/view.php?id=7158), [2](http://caml.inria.fr/mantis/view.php?id=7198)) will be? As in, is there a nice way to resolve them, such that `Event` creating all those `Condition.t`s won't be a performance issue anymore?
> 
> If performance will still be a problem, we should definitely merge this. If it won't be a problem anymore, we can still merge this, but I'm only about 75% convinced in that case – because it's a vanishing problem given the target of 4.03.1, and because it adds code. In that case, further arguments are welcome :)

The moment these issues are resolved, this PR will become unnecessary.
However, I'm not sure MPR7158 can be readily solved in a direct way (TBH, I
haven't made any serious attempt to understand that code, which is at any rate
much trickier than the present PR), and addressing MPR7198 would require work
on the GC and/or a new API, which could take a long time to happen (regardless
of the target version... I'm not sure it means it *will* positively be fixed
by that time).

Note that this PR addresses the performance issue as far as `detach` is
concerned, but `run_in_main` will still be affected (this could be solved
easily by pooling the mutexes/conditions or cells, at the cost of some more
code).

I'd say there are two arguments for merging this, and one and a half against:

pros:

* merging this solves the `detach` issue *now*
* and *for all OCaml versions*

cons:

* I might have introduced a bug (the code looks easy to inspect to me, though)
* this adds code, but given Lwt_preemptive's history (no significant changes
  in the last 4 years), it would be easy to revert at a later point in time if
  the MPRs are addressed directly and older OCaml targets become irrelevant

-- 
Mauricio Fernández

mfp:
On Fri, Jul 01, 2016 at 06:39:53PM -0700, Hezekiah M. Carty wrote:
> I haven't run into this largely because the cases where I've used `Lwt_preemptive` have been for heavier calls. So some extra GC overhead would be lost in the noise.
> 
> That is likely to change soon though so I would be happy to see performance improvements here! My main question is if this should live in Lwt or as a separate module/library in opam, used by Lwt.

The fix is specific to Lwt_preemptive's use of Event -- it does not solve the
overall Event/GC issue. The "cell" implementation is small enough to live
comfortably in Lwt_preemptive and avoid both the extra work to release it as a
separate OPAM package and adding a new dependency to Lwt.

-- 
Mauricio Fernández

aantron:
Ok, I am convinced :)

> My main question is if this should live in Lwt or as a separate module/library in opam, used by Lwt.

Assuming you mean module `CELL`, I don't think it's worth it at this point – this is just a regular data structure + CV + mutex. Plus, I don't think stdlib threads are used that much.

<br/>

The change is not so trivial actually, because it replaces a queue (channel) with a single cell (the mutable option), but it appears to be correct.

Rebased on master, committed.

> but `run_in_main` will still be affected

I'm adding a comment, in a [follow-on commit](b6aa378ef0cb4e112a621a1002c81b184f16980f), about the potential performance issue, above the implementation of `run_in_main`, so that if someone is ever debugging that, they might be able to save some time by quickly finding this PR and #218.

Thank you!

PR #227, opened 2016-03-29
Update for 4.03.0

Drup:
It uses the new module in ppx_tools to avoid any preprocessing.


Drup:
For 4.03 early adopters that have issues with oasis, I have a version with oasis pre-generated that you can pin like this: 

```
opam pin add lwt git+https://github.com/Drup/lwt.git#4.03.0-no-oasis
```

whitequark:
merge it?

PR #238, opened 2016-05-04
Rebased and reorganized #223 for Jenkins

aantron:
Discussion in #223. Resolves #223.

### DIFF
--- a/src/core/lwt_stream.ml
+++ b/src/core/lwt_stream.ml
@@ -101,14 +101,13 @@ type 'a source =
 type 'a t = {
   source : 'a source;
   (* The source of the stream. *)
+  close : unit Lwt.u;
+  (* A waiter for a thread that sleeps until the stream is closed. *)
   mutable node : 'a node;
   (* Pointer to first pending element, or to [last] if there is no
      pending element. *)
   last : 'a node ref;
   (* Node marking the end of the queue of pending elements. *)
-  hooks : (unit -> unit) list ref;
-  (* Functions called when the end of stream is reached. Hooks are
-     shared between all clones. *)
 }
 
 class type ['a] bounded_push = object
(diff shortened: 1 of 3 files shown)


seliopou:
Why were my commits rebased and modified? These commits now contain messages that I did not author, even though I'm listed as the author. Also, why not actually include a merge commit so there's a pointer back to discussion in the commit history?

This is also going to complicate merging #239.

aantron:
Everything in the commit messages is something you authored, only some commits were squashed into one, which has trivial sentence changes to make the sentences flow together after concatenation. The only sentence I added is the overall title, which is a natural consequence of a squash. I think you as author and me as committer reflects this situation accurately.

I did this while reviewing the code, because the original order, together with hidden comments in #223, made it rather difficult to clearly see what was going on, especially given that I was not yet as familiar with the surrounding code. Given that I had already done this squash for review, I decided to post and then merge it, without asking you to go through the exercise, in potentially multiple iterations with me, of cleaning up the messy history of #223 to achieve basically the same end result. While I now realize I should have clearly told you my intentions in a comment, there also was a link to this PR from #223 that was readily visible. Perhaps the title "...for Jenkins" was misleading? Anyway, I assumed that there had been time to object, so I apologize.

There are links back to the discussion. For example, https://github.com/ocsigen/lwt/commit/e4f39b850b1dc37e15f1ee03bf8250584ec7758c links to this PR, which links to #223. I grant that this isn't so useful for non-GH interfaces, so I suppose that I should prefer to merge using the GH web UI in the future.

For merging #239, I decided that I would rebase the relevant commits again myself. You are welcome to do this yourself or offer another course of action. Either way, I don't see what is so complicated – it is two `git cherry-pick`s and a `git push -f` away from being all rebased.

PR #241, opened 2016-05-09
Add Lwt_condition.broadcast_exn

nojb:


aantron:
Thanks! Looks reasonable, but, being new to maintaining Lwt, I will ponder this for a little while.

aantron:
By the way, can you sketch out, at your leisure, your use case for this function?

nojb:
When you have a set of listeners waiting on a condition the only way there is to signal failure to all of them is to introduce an auxiliary thread `stopper` and instead of using `Lwt_condition.wait` use something like:
```ocaml

let stopper, stop = Lwt.wait ()
let wait_with_cancellation cond =
  Lwt.pick [Lwt_condition.wait cond; stopper]
```
Then you can fail all waiting threads with exception `exn` by doing `Lwt.wakeup_exn stop exn`.  If then you want to do this again, you need to introduce a new `stopper`, etc. It just seemed cleaner to offer this idiom from inside `Lwt_condition`.

aantron:
Committed as a44d7a08339e9d334b017e37e55c6c20339db385. Thanks!

PR #246, opened 2016-05-17
Add builds on Cygwin in AppVeyor.

aantron:
cc @samoht

I don't believe `ocaml-ci-scripts` supports builds in Cygwin (i.e. on a compiler compiled to use Cygwin, e.g. a Cygwin package). Please let me know if anything looks wrong. This can also be contributed to `ocaml-ci-scripts`, with suitable adaptations.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,20 @@
+version: "{build}"
+
+shallow_clone: true
+
+environment:
+  global:
+    CYGWIN: C:\Cygwin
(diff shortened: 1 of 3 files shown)


aantron:
Not building on 4.03 for now due to https://github.com/ocaml/ocamlbuild/issues/84 (though it is not an Ocamlbuild issue). See [log](https://ci.appveyor.com/project/aantron/lwt/build/23/job/c65l0bwh6tvdautx#L2468).

Issue #242, opened 2016-05-12
Supporting the brand new type result. 


mmaker:
How about creating `return_ok` and `return_error` in Lwt, to behave pretty much the same as `return_nil`, `return_some`, `return_true` & co.?

<https://github.com/mmaker/ocaml-acme/commit/2478f9a0afddb2d1c5f8e0b746bcac0098644dc3>

Drup:
Related: https://github.com/ocsigen/lwt/issues/226

aantron:
[Done](https://github.com/ocsigen/lwt/blob/b0290579016d22520f23764b15ae74efec47b640/src/core/lwt.mli#L124) in #247.

mmaker:
♥

Issue #226, opened 2016-03-27
lwt_result module?


c-cube:
Since `result` is now a standard type, it would be nice to have a helper module to deal with `('a, 'b) result Lwt.t` (same as what `Lwt_list` does for lists). I can contribute an initial implementation is needed.

hcarty:
I have one that I've been using internally.  I'd be happy to share that as well.

c-cube:
I suspect many people do :-)

Drup:
Note that `Lwt.t` already contains a result type. There is also `Lwt.result`, which is isomorph to `('a, exn) result` (yes, without Lwt.t).

I think we could make `Lwt.result` really a result and add your module. @diml ?

diml:
Having `'a Lwt.result` be `('a, exn) result` seems right. Right now the `'a result -> 'a thread_state` operation would be less efficient with this equality, but eventually maybe the compiler could see it's the same thing. Anyway this optimization probably doesn't matter; in Lwt this type is used only in `Lwt_unix` and `Lwt_preemptive` and in both cases it wouldn't be a problem to not have the optimization

aantron:
Proposed implementations are welcome, especially ones informed by recent discussions on usage of result types.

aantron:
Added in #247.

PR #247, opened 2016-05-17
add `Lwt_result`

c-cube:
See #226 and #242 
This is very basic, I can expand as required by nitpickers ;)

### DIFF
--- a/src/core/lwt_result.mli
+++ b/src/core/lwt_result.mli
@@ -40,9 +40,14 @@ val catch : 'a Lwt.t -> ('a, exn) t
 (** [catch x] behaves like [return y] if [x] evaluates to [y],
     and like [fail e] if [x] raises [e] *)
 
+val get_exn : ('a, exn) t -> 'a Lwt.t
+(** [get_exn] is the opposite of {!catch}: it unwraps the result type,
+    returning the value in case of success, calls {!Lwt.fail} in
+    case of error. *)
+
 val map : ('a -> 'b) -> ('a,'e) t -> ('b,'e) t
 
-val map_err : ('e1 -> 'e2) -> ('a,'e1) t -> ('b,'e2) t
+val map_err : ('e1 -> 'e2) -> ('a,'e1) t -> ('a,'e2) t
 
 val bind : ('a,'e) t -> ('a -> ('b,'e) t) -> ('b,'e) t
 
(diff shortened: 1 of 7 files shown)


hcarty:
We should have `Lwt.return_ok` and `Lwt.return_error`, similar to `Lwt.return_some`.

src/core/lwt_result.mli
```diff
@@ -0,0 +1,54 @@
...
+
+(** Module [Lwt_result]: explicit error handling *)
+
+(** This module provides helpers for values of type [('a, 'b) result Lwt.t] *)
+
+type (+'a, +'b) t = ('a, 'b) Result.result Lwt.t
+
+val return : 'a -> ('a, _) t
+
+val fail : 'b -> (_, 'b) t
```
hcarty: I'm worried this name may be a little confusing when seen with `Lwt.fail`.  Would `error` or maybe `return_error` be clearer?

src/core/lwt_result.mli
```diff
@@ -0,0 +1,54 @@
...
+
+val lift : ('a, 'b) Result.result -> ('a, 'b) t
+
+val ok : 'a Lwt.t -> ('a, _) t
+
+val map : ('a -> 'b) -> ('a,'e) t -> ('b,'e) t
+
+val map_err : ('e1 -> 'e2) -> ('a,'e1) t -> ('b,'e2) t
+
+val bind : ('a,'e) t -> ('a -> ('b,'e) t) -> ('b,'e) t
```
hcarty: It would be nice to have this and `bind_lwt` in both `Ok` and `Error` forms for conditionally rewriting each case specifically.

c-cube:
I'm considering adding something like

```ocaml
val get_exn : ('a, exn) t -> 'a Lwt.t
```

to go with `catch : 'a Lwt.t -> ('a, exn) t`.
Any opinion?

hcarty:
`get_exn` :+1:

c-cube:
@drup proposes to deprecate `Lwt.make_value` and `Lwt.make_error`. I have no opinion on that, what do you decide, @aantron?

aantron:
I think we should deprecate them, in agreement with @Drup.

aantron:
Thanks @c-cube, @hcarty.

As agreed, we can make modifications in follow-on PRs.

Issue #251, opened 2016-05-24
Lwt_preemptive should provide an "at_exit" for use when a thread ends


hcarty:
I'm working on bindings for [libyara](https://github.com/plusvic/yara) which requires a cleanup call when a thread which used the library exits.  It would be nice to have a way to provide init/cleanup functions which would be applied to each new/exiting thread.

aantron:
Agreed in principle. I have zero experience with `Lwt_preemptive`, but, just glancing at it, couldn't you achieve this with some variation on

```ocaml
Lwt_preemptive.detach f (init (); v) >|= fun r -> cleanup (); r
```

Let me know what I am missing. If this turns out to be a documentation issue, we should fix that too.

hcarty:
The cleanup, in this case, needs to be called once the preemptive thread exits, not the Lwt thread.  So the `at_exit` function would need to be called when the preemptive thread ends which may be after calls to `Lwt_preemptive.detach`.

aantron:
I just looked at `Lwt_preemptive` more closely, and I see what you mean. PRs welcome; else I will reach this "eventually."

hcarty:
I'm working on something - I'll hopefully be able to get an initial PR out for review soon.

hcarty:
See #252

aantron:
Closing as discussion was continued in #252.

PR #252, opened 2016-05-25
Add ?init and ?at_exit to Lwt_preemptive

hcarty:
Addressing #251

This PR is a proposal in its current state - I'd want to add some tests before merging to ensure these work as they should.

These functions are intended to be called once per preemptive thread.
This is meant to allow setup and cleanup for libraries which require
per-thread management.

### DIFF
--- a/src/preemptive/lwt_preemptive.ml
+++ b/src/preemptive/lwt_preemptive.ml
@@ -60,6 +60,12 @@ type thread = {
   (* Channel used to communicate notification id and tasks to the
      worker thread. *)
 
+  mutable init : (unit -> unit) option;
+  (* function to call when a new thread is created *)
+
+  mutable at_exit : (unit -> unit) option;
+  (* Function to call when the thread exits *)
+
   mutable thread : Thread.t;
   (* The worker thread. *)
 
@@ -77,19 +83,33 @@ let waiters : thread Lwt.u Lwt_sequence.t = Lwt_sequence.create ()
 (* Code executed by a worker: *)
 let rec worker_loop worker =
   let id, task = Event.sync (Event.receive worker.task_channel) in
+  begin
+    match worker.init with
+    | None -> ()
+    | Some f -> f (); worker.init <- None
+  end;
   task ();
   (* If there is too much threads, exit. This can happen if the user
      decreased the maximum: *)
   if !threads_count > !max_threads then worker.reuse <- false;
   (* Tell the main thread that work is done: *)
   Lwt_unix.send_notification id;
-  if worker.reuse then worker_loop worker
+  if worker.reuse then (
+    worker_loop worker
+  )
+  else (
+    match worker.at_exit with
+    | None -> ()
+    | Some f -> f (); worker.at_exit <- None
+  )
 
 (* create a new worker: *)
-let make_worker () =
+let make_worker init at_exit =
   incr threads_count;
   let worker = {
     task_channel = Event.new_channel ();
+    init = init;
+    at_exit = at_exit;
     thread = Thread.self ();
     reuse = true;
   } in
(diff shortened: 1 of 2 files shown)


aantron:
Concerning the interface, does it make sense to avoid the `init` and `at_exit` arguments in all functions? Since `Lwt_preemptive` is (apparently) a per-process pool of threads, perhaps the best we can do anyway is some global state?

Does it make sense to make this similar to Lwt's thread-local storage, i.e. you get some kind of "key" when you add a handler, so you can remove the handler later? That way, different modules linked into the program should be able to add their own handlers.

hcarty:
It we have a concept of `Lwt_preemptive`-thread local storage I think that could work.

Another potential solution would be to allow users to create multiple thread pools.  `Lwt_preemptive` currently has one process-global thread pool.  If multiple pools could be created then the `init` and `at_exit` hooks could be set per-pool.

aantron:
It seems you would like the ability to selectively associate handlers to some threads but not others, and maybe also to guarantee that a pool thread that runs particular Lwt "threads" will have handlers associated – as opposed to associating the handlers to all threads/all threads created after a certain point. Is that correct?

Could you clarify the scenario – which threads need handlers, when associating handlers is decided during execution, and whether it is wrong to "over-estimate," and associate handlers with "too many" threads? How should this feature work, if multiple external libraries required it?

hcarty:
> It seems you would like the ability to selectively associate handlers to some threads but not others, and maybe also to guarantee that a pool thread that runs particular Lwt "threads" will have handlers associated – as opposed to associating the handlers to all threads/all threads created after a certain point. Is that correct?

That is correct.

> Could you clarify the scenario – which threads need handlers, when associating handlers is decided during execution, and whether it is wrong to "over-estimate," and associate handlers with "too many" threads? How should this feature work, if multiple external libraries required it?

If `Lwt_preemptive` were updated to support multiple thread pools then I think it would make sense to allow users to pick which pool they want to `detach` an operation into.  That would allow a user to decide if it is safe to use multiple external libraries within a single pool.

hcarty:
We've discussed this a bit on IRC, but for proper record a link to a WIP experiment around this is here: https://github.com/hcarty/mwt

mfp:
In addition to being much safer w.r.t. to the original motivation (per-thread setup/cleanup), multiple thread pools are an useful addition on their own. I've found myself reusing/adapting `Lwt_preemptive`'s worker pool at least twice (to [manage Curl workers](https://github.com/mfp/curl-lwt/blob/master/curl_lwt_ez.mli) and to [work with Sqlite3 DB handles](https://github.com/mfp/ocaml-sqlexpr/blob/master/src/sqlexpr_sqlite_lwt.ml#L181), where the latter moreover requires that the handles be used in the same thread where they were created, i.e., TLS). I've spotted a couple more instances in the wild by looking for `Lwt_unix.make_notification`.

So what's the plan for this? Integration in `Lwt_preemptive`, or release as a separate library blessed in `Lwt_preemptive`'s docs, with eventual deprecation of the latter?

aantron:
I'm not sure where @hcarty wants to take this – @hcarty?

More broadly, `Lwt_preemptive` needs study, including concerning any future interactions with multicore and how to get that right.

hcarty:
If there's support behind it I'm ok with moving `Mwt` or something like it into a proper replacement for `Lwt_preemptive`.  Including any changes required to address the TLS-related requirements @mfp mentioned.  It could be maintained outside of Lwt-proper or as part of Lwt.

aantron:
We discussed with @hcarty and decided that this will continue in [Mwt](https://github.com/hcarty/mwt). If it looks like it's a good idea to merge Mwt into Lwt later, we can merge the histories at that point :)

PR #256, opened 2016-06-20
Add Lwt_switch.with_switch

talex5:
I find myself writing this helper every time I use Lwt. Seems like it should be in the library.

I also updated the example to use it. Technically, it's not quite the same since the example now turns off the switch on success too, but that seems like a reasonable thing to do anyway.


aantron:
Looks good, thanks!

Issue #250, opened 2016-05-23
Flaws of Lwt_stream


aantron:
Please tell why `Lwt_stream` doesn't work for you, ideally at the level of semantics (i.e. prefer "lazy streams aren't right for our use case because blahblah" to "`Lwt_stream` is missing a specific function that can easily be added").

Hopefully we can
- change it in a minor or major release,
- settle on some helper library or libraries aside from Lwt, and gradually starting moving `Lwt_stream` towards deprecation,
- keep it, if you think it's great,
- or your alternative proposals.

cc @c-cube @hcarty @seliopou @Drup @diml 

Please pull in anyone else who has interest/expertise.

Some existing anguish: #239, #151.

EDIT: Despite the issue title, [thank you](https://github.com/ocsigen/lwt/issues/250#issuecomment-222287378) to @diml.

EDIT: Note also #155.

c-cube:
I don't have the solution(s), but issues with `Lwt_stream` that come to my mind are:

- the semantics is not trivlal (several ways of creating a stream, with several distinct semantics)
- biggest issue is resource management. I am not sure how to handle this at all (doing it really cleanly might require linear types?); the thing is, you never know which part of the code might be using a stream (or the `Lwt_stream.map f  (Lwt_stream.filter g the_stream)`), so when the consumer is done it doesn't really have a way of signaling the whole pipeline and having the underlying source closed (if `the_stream` is a file descriptor, say).
- I have an issue somewhere with multiple producer/multiple consumers, I think, although it's been a while and I don't really remember precisely.

I think `Lwt_stream` should mostly stay as is; opam makes it easy to depend on 3rd party libraries that have more freedom to explore, at least for now. It's like `Stream` in the stdlib, it's almost impossible to improve in a retrocompatible way. I don't think it's time for deprecation yet, not until we have robust, well tested alternatives in 3rd party libs.

aantron:
Thanks. Agreed.

> 3rd party libraries that have more freedom to explore

I would like to move this along. Of course, `Lwt_stream` will remain mostly as is for a while – we have to be conservative. But I am highly open to, for example, "blessing" an alternative library in the docs of `Lwt_stream`, if we have/develop one that addresses many of the issues people have. It's kind of sad to offer a module that:

- appears to provide nice streams,
- but, after trying it, users discover that it doesn't do what they want,
- which then forces people to do research to find an alternative.

[...]

PR #257, opened 2016-06-22
Add IVars to LWT

rneswold:
This commit creates a new module, `Lwt_ivar`, that defines an IVar[1] type. IVars can only be set once. Threads trying to get an IVar's value before it is set will block.

----

[1] Described in John Reppy's "Concurrent Programming in ML", found in Async's `Std.IVar` module and also Daniel Bünzli's `Fut` library (as `Fut.promise`).

### DIFF
--- a/src/core/lwt_ivar.ml
+++ b/src/core/lwt_ivar.ml
@@ -1,0 +1,56 @@
+(*****************************************************************************)
+(* Lightweight thread library for OCaml
+ * http://www.ocsigen.org/lwt
+ * Interface Lwt_ivar
+ ******************************************************************************
+ * Copyright (c) 2016, Richard M Neswold, Jr
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above
+ *       copyright notice, this list of conditions and the following
+ *       disclaimer in the documentation and/or other materials provided
+ *       with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY METAWEB TECHNOLOGIES ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL METAWEB TECHNOLOGIES BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR
+ * BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ * WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
+ * OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN
+ * IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *****************************************************************************)
+
+exeception Already_set
+
+type 'a t = { mutable contents : 'a option;
+              readers : 'a Lwt.u Lwt_sequence.t }
+
+let create () = { contents = None;
+                  readers = Lwt_sequence.create () }
+
+let put ivar v =
+  match ivar.contents with
+  | None ->
+     begin
+       ivar.contents <- Some v;
+       Lwt_sequence.iter_l (fun w -> Lwt.wakeup_later w v) ivar.readers;
+       Lwt.return_unit
+     end
+  | Some _ ->
+     Lwt.fail Already_determined
+
+let get ivar =
+  match ivar.contents with
+  | Some v ->
+     Lwt.return v
+  | None ->
+     Lwt.add_task_r ivar.readers
--- a/src/core/lwt_ivar.mli
+++ b/src/core/lwt_ivar.mli
@@ -1,0 +1,55 @@
+(*****************************************************************************)
+(* Lightweight thread library for OCaml
+ * http://www.ocsigen.org/lwt
+ * Interface Lwt_ivar
+ ******************************************************************************
+ * Copyright (c) 2016, Richard M Neswold, Jr.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ * * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided
+ * with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY METAWEB TECHNOLOGIES ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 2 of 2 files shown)


aantron:
@rneswold Can you show a use-case, where `Lwt_ivar.t` is preferable to using `Lwt.t`? It seems like you could implement this easily using [`Lwt.wait`](http://ocsigen.org/lwt/2.5.1/api/Lwt#VALwait). Here's a sketch of typical usage:

```ocaml
let wait_for_something, put_something = Lwt.wait () in

wait_for_something >>= fun i -> ... (* like "get", blocks *)
wait_for_something >>= fun i -> ... (* another "get" *)

Lwt.wakeup put_something 42; (* those two waiters will get 42 *)
(* additional calls to Lwt.wakeup after this will raise Invalid_argument *)

wait_for_something >>= fun i -> ... (* gets the same 42, immediately *)
```

You could, of course, wrap it all in the `Lwt_ivar` interface, but I'm not sure there is a gain. Again, a sketch:

```ocaml
type 'a t = 'a Lwt.t * 'a Lwt.u
let create = Lwt.wait
let put ivar v = Lwt.wakeup (snd ivar) v; Lwt.return_unit
let get = fst
```

The one non-trivial difference I see is the inability to communicate exceptions through an ivar. This may be good or bad. But `Lwt.wait` with only `Lwt.wakeup` (vs. `Lwt.wakeup_exn`) also cannot communicate exceptions. And, regardless, the `Lwt_ivar` interface loses the information about being exception-free right away in `get`, by evaluating to an `Lwt.t`.

rneswold:
One use-case is an efficient implementation of streams. The request-for-comments in issue #250 got me thinking about this.

Take this simplified stream signature:

```OCaml
module Lwt_stream :
sig
  exception End_of_stream

  type 'a t

  val create : unit -> 'a t * ('a -> unit)
  val clone : 'a t -> 'a t
  val next : 'a t -> 'a Lwt.t
  val iter : ('a -> unit Lwt.t) -> 'a t -> unit Lwt.t
  val map : ('a -> 'b) -> 'a t -> 'b t
  val filter : ('a -> bool) -> 'a t -> 'a t
 end
```

A stream can be implemented using a chain of IVars (data is added to the end of the chain):

```OCaml
type 'a node = Node of ('a * 'a node Lwt_ivar.t)
             | End of exn

type 'a t = { mutable last : 'a node Lwt_ivar.t;
              next_value : ('a t -> 'a Lwt.t }
```

Each `'a t` (i.e. stream) keeps track of where it is in the IVar chain. As threads consume the chain, nodes become unreferenced and get garbage collected.

Creating the IVar is mostly just setting up the returned record and the push function. We also register a finalizer.

```OCaml
(* One of [create]'s return values is a function to push a value
   at the end of the IVar chain. It is returned curried in that
   the caller only gets to provide the value, not the node to
   which they're adding a value. *)

let push node v =
  let next = Lwt_ivar.create () in
    begin
      Lwt_ivar.put !node @@ Node (v, next);
      node := next
    end

(* A finalizer is added to the push function. When the function is
   no longer used, the stream gets an End_of_stream indicator. *)

let final rn p =
  Lwt_gc.finalize (fun _ -> Lwt_ivar.put !rn (End End_of_stream)) p

(* Returns the stream and the function which adds values to the
   end of it. *)

let create () =
  let node = Lwt_ivar.create () in
    let rn = ref node
    and next t =
      match%lwt Lwt_ivar.get t with
      | Node (v, next_node) -> ( t.last := next_node; Lwt.return v )
      | End e -> Lwt.fail e
    in
    let p = push rn in
    ( final rn p; ({ last = node; next_value = next }, p) )
```

Cloning is simple:

```OCaml
let clone t = { last = t.last; next_value = t.next_value }
```

Pulling the next value uses the embedded function. I chose to embed the "next element" function in the record to make it easier to implement `map` and `filter`  (below.)

```OCaml
let next t = t.next_value t
```
Iterating through the stream is easy:

```OCaml
let iter f t =
  let rec loop () = next t >>= f >>= loop in
  loop ()
```

For `map` and `filter`, we clone the stream, but wrap the embedded "next element" function.

```OCaml
let map f t =
  let nf t' = t.next_value t' >>= Lwt.wrap1 f in
  { last = t.last; next_value = nf }

let filter f t =
  let rec nf t' =
    let%lwt v = t.next_value t' in
    if f v then
      Lwt.return v
    else
      nf t'
  in
  { last = t.last; next_value = nf }
```

Benefits:

* Streams have little state. Cloning is cheap.

* No complexity keeping track whether a stream is "closed". Just traverse the chain until the end.

* All resources are handled by reference counting and garbage collection.

Drawbacks:

* Two threads using the same stream is still undefined. However, a mutex could be added to the stream state which serializes access to the stream. A group of "worker" threads could then read from the same stream. New data would get assigned to a worker.

* Bounded streams are no longer possible without greatly complicating the code. I prefer, however, a bounded stream to be called something else: bounded queue or FIFO, for instance.

* Wrapping streams with `map` and `filter` simply wrap the function that pulls the next value. This, in itself, is fine. However, cloning these streams means each clone is doing the same work for each element.

I thought that maybe `Lwt.wait` could be used (and it's good to know that, once awakened, the thread will continue to return the same value.) My concern was, if a thread was blocked waiting for the next stream value and was canceled, wouldn't the `Lwt.wait` thread be permanently put in a Fail state? I want the stream to remain valid even if a waiting thread gets canceled (because cloned streams are still using the underlying IVars.)

aantron:
> My concern was, if a thread was blocked waiting for the next stream value and was canceled, wouldn't the Lwt.wait thread be permanently put in a Fail state?

It's not possible to cancel a thread created with `Lwt.wait` at all. However, if `create` used `Lwt.task` instead, then it would have the behavior you described.

>  I want the stream to remain valid even if a waiting thread gets canceled

You can achieve cancelable waiters, without canceling the underlying cells (not that that would work, due to `Lwt.wait`), by using this for `get` instead of just `fst`:

```ocaml
let get ivar = Lwt.protected (fst ivar)
```

IMO an ivar, as proposed here, is exactly an `Lwt.t`, and we should leave it up to the user to decide if a particular thread's wait should be cancelable or not:

```ocaml
let ivar, put = Lwt.wait () in

ivar >>= ...               (* non-cancelable get *)
Lwt.protected ivar >>= ... (* cancelable get *)
```

Either way, the "ivar" itself cannot be canceled, due to `Lwt.wait`.

Perhaps what's needed is to improve the documentation.

Whether you have specific ivars, or just use `Lwt.t`, I'm wondering whether this could be an implementation detail of a proposed stream module, at least for now? Is there a use case beyond implementing other things in Lwt?

Regarding streams, when you are ready, please add your comments to #250, so the other participants can easily see them :)

rneswold:
> IMO an ivar, as proposed here, is exactly an `Lwt.t`

Excellent! Then I withdraw my pull request. I'll rework my stream example using `Lwt.wait` and post it to #250 to spur on more discussion.

> Perhaps what's needed is to improve the documentation.

Yes, I agree. There are many functions that have no documentation (because we are to assume they function similarly to the `Unix` or `List` versions), and many others with one or two sentences of description. That leaves them open to a lot of interpretation.

Take `Lwt.protected`, for instance. Its documentation says:

> `protected thread` creates a new cancelable thread which behaves as `thread` except that canceling it does not cancel `thread`.

Given the following expression:

```OCaml
do_something () >>
  Lwt.protected (crit_thing_1 () >> crit_thing_2 ()) >>
  do_something_else ()
```

my original interpretation was that, if `crit_thing_1` is being resolved and the thread gets cancelled, `Lwt.protected` would let its thread complete (in this case, up to the resolution of `crit_thing_2`) before being canceled. In other words, `Lwt.protected` creates a critical section.

With your examples, however, now I believe that the thread is canceled at the `Lwt.protected` boundary and the sub-thread (`crit_thing_1 () >> crit_thing_2 ()`) will still run to completion, since it never sees the cancellation.

Am I getting closer to understanding `Lwt.protected`?

aantron:
> I believe that the thread is canceled at the Lwt.protected boundary
> ...
> Am I getting closer to understanding Lwt.protected?

I believe so. What gets canceled ultimately depends on what you try to cancel :) So, binding both threads to names in order to be clear:

```ocaml
let wrapped_thread, _ = Lwt.task () in
let wrapper_thread = Lwt.protected wrapped_thread in

Lwt.cancel wrapper_thread; (* cancels wrapper_thread, but not wrapped_thread *)
Lwt.cancel wrapped_thread; (* cancels both threads *)
```

So, it's just a matter of whether the cancelation propagates to the enclosed thread or not (it doesn't).

I got many documentation TODOs out of this PR, so yeah :)

rneswold:
Not related to IVars or doc TODOs, but I've rewritten my stream example using `Lwt.wait` and it seems to work. Plus the code is cleaner; since we're dealing with threads, I may be able to remove the record representing the stream state and simply use a thread with closures. Very nice! Thanks for your input, @aantron.

Issue #196, opened 2015-12-04
usefullness of Lwt_io.establish_server


zoggy:
Hello,

The type of `Lwt_io.establish_server` is the following:
````ocaml
val establish_server : 
  ?fd:Lwt_unix.file_descr ->
  ?buffer_size:int ->
  ?backlog:int ->
  Unix.sockaddr ->
  (input_channel * output_channel -> unit) ->
  server
````
The function `f` handling connections is supposed to return `unit` instead of `unit Lwt.t`. Is this intended ? With this type, `establish_server` is useless, since using the channels requires using Lwt_io functions. Looking at the implementation, the server does not accepted any other request during the execution of `f`. This is not very Lwt-ish.

diml:
I wrote this signature, the intent was for the callback to start some computation in the background and return immediately. You can use `Lwt.async` for instance

zoggy:
Hum, ok.

PR #258, opened 2016-06-23
Safer alternative to Lwt_io.establish_server

### LINKED ISSUE #208: Easily reproducible leaking of file descriptors, show stopper.
I have boiled down my usage to this code sample which will leak file descriptors. 

say you have:

```ocaml
#require "lwt.unix"

open Lwt.Infix

let echo ic oc = Lwt_io.(write_chars oc (read_chars ic))

let program =
  let server_address = Unix.(ADDR_INET (inet_addr_loopback, 2000)) in

  let other_addr = Unix.(ADDR_INET (inet_addr_loopback, 2001)) in

  let server = Lwt_io.establish_server server_address begin fun (tcp_ic, tcp_oc) ->
      Lwt_io.with_connection other_addr begin fun (nc_ic, nc_oc) ->

        Lwt_io.printl "Created connection" >>= fun () ->
        echo tcp_ic nc_oc <&> echo nc_ic tcp_oc >>= fun () ->
        Lwt_io.printl "finished"

      end
      |> Lwt.ignore_result

    end
  in
  fst (Lwt.wait ())

let () =
  Lwt_main.run program
```

and then you create a simple server with:

```shell
nc -l 2001
```

and then let's start up the OCaml code with 
```shell
utop example.ml
```

and then open up a client

```shell
nc localhost 2000
blah blah
^c
```
Then looking at the connections for port 2000 using lsof, we see 
```shell
ocamlrun 71109 Edgar    6u  IPv4 0x7ff3e309cb80aead      0t0  TCP 127.0.0.1:callbook (LISTEN)
ocamlrun 71109 Edgar    7u  IPv4 0x7ff3e309c9dc8ead      0t0  TCP 127.0.0.1:callbook->127.0.0.1:54872 (CLOSE_WAIT)
```
In fact for each usage of `nc localhost 2000`, we'll get a leftover `CLOSE_WAIT` record from the lsof usage. 

Eventually this will lead to the system running out of file descriptors, which will MOST annoyingly not crash the program, but will lead to Lwt to just hang.

I can't tell if I am doing something wrong or if this is a genuine bug, in any case this is a serious bug for me and I run out of file descriptors within 10 hours...

EDIT: It seems to me that the problem is that one side of the connection is closed but the other isn't, I would have thought that `with_connection` should cleanup/close up whenever either side closes.

EDIT II: I have tried every which way where I manually close the descriptors with `Lwt_io.close`, but I still have the `CLOSE_WAIT` message.

EDIT III: Even used Lwt_unix.close on a raw fd given to with_connection's optional fd argument with similar bad results.

EDIT IV: Most insidious is if I use `Lwt_daemon.daemonize`, then this problem seemingly goes away

### LINKED ISSUE #225: close sockets eagerly in `Lwt_io.establish_server`
Currently, at https://github.com/ocsigen/lwt/blob/ebd0a99438512dff283fa779e478ed87cc353d22/src/unix/lwt_io.ml#L1416 , it seems that `establish_server` doesn't close the FDs itself. A server receiving lots of short-lived connections quickly blocks by FD starvation, unless on closes the two channels manually. I think the channels should be closed by `establish_server` itself by default (or at least providing an optional flag to enable it), à la `with_file`, etc..

aantron:
[`Lwt_io.establish_server`][establish_server] is [designed][intent] to be analogous to [`Lwt_io.open_connection`][open_connection]: it hands you channels, but doesn't try to close them. You have to close them yourself.

People have expressed ([1][issue-208], [2][issue-225]) a desire for a version of `establish_server` that is more like [`Lwt_io.with_connection`][with_connection]: it runs your callback `f` to spawn handler threads, and automatically schedules closing the associated channels once each thread completes. This PR adds such a function:

```ocaml
val establish_server_async :
  ?fd : Lwt_unix.file_descr ->
  ?buffer_size : int ->
  ?backlog : int ->                                     (* !!!!! *)
  Unix.sockaddr -> (input_channel * output_channel -> unit Lwt.t) -> server
```

I'm calling it `establish_server_async` as [suggested][name-suggestion] by @c-cube for now, but other suggestions are welcome.

This seems like a difficult problem at first, since if the implicit `close` calls raise exceptions, the user code has no opportunity to catch them and do logging, etc – among other potential challenges. However, if the user needs control over that, it can explicitly close the channels in `f`, and surround that with whatever handling code it wants. The additional implicit `close`s will then have no effect.

So, this should give a safer API for simple usage, that nonetheless remains usable in more complex scenarios.

In fact, I've thought about this at length, and I can find only one reason, besides API breakage, not to replace `establish_server`: one may want to disable implicit channel closing, because the channels are, for example, being stored by `f` in some data structure. However, we could control that with an optional argument.

Other things to consider:

- graceful shutdown,
- concurrent connection limiting,
- aforementioned exception handling,

I don't think `establish_server_async` has any impact on these, relative to `establish_server`, but...

- there are probably other things I didn't think of.

I will probably rewrite the documentation after getting a fresh look at it in some days.

cc @c-cube @fxfactorial @artemkin

Resolves #208.
Resolves #225.

[establish_server]: http://ocsigen.org/lwt/2.5.1/api/Lwt_io#VALestablish_server
[intent]: https://github.com/ocsigen/lwt/issues/196#issuecomment-161936473
[open_connection]: http://ocsigen.org/lwt/2.5.1/api/Lwt_io#VALopen_connection
[with_connection]: http://ocsigen.org/lwt/2.5.1/api/Lwt_io#VALwith_connection
[issue-208]: https://github.com/ocsigen/lwt/issues/208
[issue-225]: https://github.com/ocsigen/lwt/issues/225
[name-suggestion]: https://github.com/ocsigen/lwt/issues/225#issuecomment-202036456

### DIFF
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -455,6 +455,11 @@ let close : type mode. mode channel -> unit Lwt.t = fun wrapper ->
             (fun _ ->
               abort wrapper)
 
+let is_closed wrapper =
+  match wrapper.state with
+  | Closed -> true
+  | _ -> false
+
 let flush_all () =
   let wrappers = Outputs.fold (fun x l -> x :: l) outputs [] in
   Lwt_list.iter_p
@@ -1423,6 +1428,41 @@ let establish_server ?fd ?(buffer_size = !default_buffer_size) ?(backlog=5) sock
   in
   ignore (loop ());
   { shutdown = lazy(Lwt.wakeup abort_wakener `Shutdown) }
+
+let establish_server_safe ?fd ?buffer_size ?backlog sockaddr f =
+  let best_effort_close channel =
+    (* First, check whether the channel is closed. f may have already tried to
+       close the channel, received an exception, and handled it somehow. If so,
+       trying to close the channel here will trigger the same exception, which
+       will go to !Lwt.async_exception_hook, despite the user's efforts. *)
+    (* The Invalid state is not possible on the channel, because it was not
+       created using Lwt_io.atomic. *)
+    if is_closed channel then
+      Lwt.return_unit
+    else
+      Lwt.catch
+        (fun () -> close channel)
+        (fun exn ->
+          !Lwt.async_exception_hook exn;
+          Lwt.return_unit)
+  in
+
+  let handler ((input_channel, output_channel) as channels) =
+    Lwt.async (fun () ->
+      (* Not using Lwt.finalize here, to make sure that exceptions from [f]
+         reach !Lwt.async_exception_hook before exceptions from closing the
+         channels. *)
+      Lwt.catch
+        (fun () -> f channels)
+        (fun exn ->
+          !Lwt.async_exception_hook exn;
+          Lwt.return_unit)
+
+      >>= fun () -> best_effort_close input_channel
+      >>= fun () -> best_effort_close output_channel)
+  in
+
+  establish_server ?fd ?buffer_size ?backlog sockaddr handler
 
 let ignore_close ch =
   ignore (close ch)
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1346,26 +1346,26 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
 
 let file_length filename = with_file ~mode:input filename length
 
+let close_socket fd =
+  Lwt.finalize
+    (fun () ->
+      Lwt.catch
+        (fun () ->
+          Lwt_unix.shutdown fd Unix.SHUTDOWN_ALL;
+          Lwt.return_unit)
+        (function
+        (* Occurs if the peer closes the connection first. *)
+        | Unix.Unix_error (Unix.ENOTCONN, _, _) -> Lwt.return_unit
+        | exn -> Lwt.fail exn))
+    (fun () ->
+      Lwt_unix.close fd)
+
 let open_connection ?fd ?in_buffer ?out_buffer sockaddr =
   let fd = match fd with
     | None -> Lwt_unix.socket (Unix.domain_of_sockaddr sockaddr) Unix.SOCK_STREAM 0
     | Some fd -> fd
   in
-  let close = lazy begin
-    Lwt.finalize
-      (fun () ->
-        Lwt.catch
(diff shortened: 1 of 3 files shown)


c-cube:
Didn't look at the code, but this seems very useful indeed. Safe usage of resources is crucial (I hate debugging FDs leaks).

About the "concurrent connection limiting": this could be done with an optional `lwt_pool` (in fact, I often find myself using a dumb version of `lwt_pool` with units, basically acting as a semaphore — would `lwt_semaphore` be a useful shortcut for this common use-case of limiting concurrency? — it could also be nice for, say, `Lwt_list.map_p` and the likes)

aantron:
> About the "concurrent connection limiting": this could be done with an optional lwt_pool

An `Lwt_pool` could be used "orthogonally" to limit how many connection threads can run the "body" of their handlers, but it won't stop `establish_server_async` from `accept`ing an arbitrary number of connections (most of which would then block on the pool). I'm wondering if this is a problem or not, and if we need to add an optional argument to `establish_server[_async]` to control that. That makes me think it's better to wait until someone requests it, then they can explain why they need it :)

> would lwt_semaphore be a useful shortcut for this common use-case of limiting concurrency?

Seems like it might be a nice thing to have. Want to be careful, however, because ideally we could reduce the number of features in Lwt, and/or reinterpret more things in terms of others.

c-cube:
> Seems like it might be a nice thing to have. Want to be careful, however, because ideally we could reduce the number of features in Lwt, and/or reinterpret more things in terms of others.

It can be a simple addition to `Lwt_pool`, actually, with a simple function:

```ocaml
type semaphore = private int t
val semaphore : int -> semaphore
val with_semaphore : semaphore -> (unit -> 'a Lwt.t) -> 'a Lwt.t
```

aantron:
> It can be a simple addition to Lwt_pool, actually, with a simple function:

Yes. But I guess a mutex could also be a semaphore with limit 1, but an `Lwt_pool` doesn't, currently, have the same release semantics as `Lwt_mutex`, and I don't know, at this point, if there is an unacceptable performance hit to making an `Lwt_mutex` be an `Lwt_pool` "under the hood."

Also, I wonder if/how `Lwt_preemptive` is related.

PR #261, opened 2016-06-28
tests: add reproducer for pull request 202: mutex deadlock on cancel

stijn-devriendt:
This test can reproduce the issue on 2.3.2, while the fix provided in the
pull request below fixes the issue.
However, due to changes in the lwt core, this issue is no longer
reproducible on 2.5.0.

By inspection, it looks like this issue was introduced by
d6e3bd423e95695e144509e3cb97439b792cb556
and fixed by
caf3ed199306391e20e93c859666066bfeb433de

This commit immediately sets the state of the task in wakeup_later, making
the subsequent cancel a noop.

See also:
https://github.com/ocsigen/lwt/pull/202

### DIFF
--- a/tests/core/test_lwt_mutex.ml
+++ b/tests/core/test_lwt_mutex.ml
@@ -115,4 +115,32 @@ let suite = suite "lwt_mutex" [
       (* Run thread 3. *)
       Lwt.wakeup wake_top_level_waiter ();
       while_waking);
+
+  (* See https://github.com/ocsigen/lwt/pull/202 - This is an actual reproducer on 2.3.2 *)
+  test "mutex issue"
+    (fun () ->
+       (* Create and lock the mutex *)
+       let mtx = Lwt_mutex.create () in
+       let _prelock = Lwt_mutex.lock mtx in
+
+       (* create a task (lock-try-noop-finally-unlock) pending on the lock *)
+       let t = Lwt_mutex.with_lock mtx (fun () -> Lwt.return ()) in
+
+       (* create a task to unlock and immediately cancel t *)
+       let waiter, wakener = Lwt.task () in
+       let _ = Lwt.bind waiter (fun () ->
+           let () = Lwt_mutex.unlock mtx in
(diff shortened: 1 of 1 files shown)


aantron:
Thanks for looking into this further!

I looked at the proposed test, and it is very similar to the new [`cancel while queued by unlock`](https://github.com/ocsigen/lwt/blob/master/tests/core/test_lwt_mutex.ml#L63). However, they have different behavior in 2.3.2.

- I backported `cancel while queued by unlock` to 2.3.2, and removed all checked conditions except `not (Lwt_mutex.is_locked mutex)`. To my surprise, the test *passed*, while the test in this PR fails.
- The difference is due to the test [using `Lwt.wakeup`](https://github.com/ocsigen/lwt/blob/master/tests/core/test_lwt_mutex.ml#L116) instead of `Lwt.wakeup_later`. `Lwt.wakeup` in `master` sets `wakening`, but in 2.3.2 only `Lwt.wakeup_later` sets `wakeuping`. Once I changed the test to use `Lwt.wakeup_later`, it also failed in 2.3.2.

So, I suggest that I (or you) make a different commit/PR, that changes that line, to make the existing `master` test fail on that condition, if it were backported to older Lwt. The test actually tests more things, some of which we cannot expect in 2.3.2 anyway, so it will also fail for other reasons, and is therefore not truly accurate for 2.3.2. However, it is accurate for `master`. And, while the proposed test is more accurate on 2.3.2, I think, with the above change to the existing test, it becomes redundant on `master`.

The proposed test, and the patch from #202, would be nice to have in a bugfix series for 2.3, but Lwt does not keep those. So, I think the next best thing is to keep this PR around, mention it from the new commit/PR, add a comment linking here above `cancel while queued by unlock`, and close this PR with tag "archive".

What do you think?

stijn-devriendt:
I'm fine with that. Have a look at https://github.com/ocsigen/lwt/pull/262

aantron:
Thanks. Really appreciate your thoroughness.

PR #274, opened 2016-09-27
Add _opam in gitignore for OPAM 2.0 local switch

dannywillems:
OPAM 2.0 allows to have a local configuration and compiler (see https://opam.ocaml.org/blog/opam-2-0-preview/). This configuration is put in `_opam` directory.
Useful for people who wants to contribute to `lwt` without having to install all dependencies in another switch.


aantron:
Thanks!

For reference: https://opam.ocaml.org/doc/2.0/Manual.html#Switches

PR #278, opened 2016-10-17
Lwt_unix: do not truncate notification ids in C

### LINKED ISSUE #277: type mismatch causes lwt_unix job hangs
To reproduce:
In Lwt_unix.ml, change the following::

      let current_notification_id = ref (0x7FFFFFFF - 1000)

Then run the tests. Test lwt_io_non_block:"many read file" will hang.

Root cause is a type mismatch in the C-code. notification_id is used as int (Int_val) in the job code and long (Long_val) in other cases. The result is that when the notification ID reaches 2^31, notifications from unix jobs are lost. This is because the 0x80000000L is cast to int, yielding INT_MIN. Later on, this value is cast back to long with sign extension (0xFFFFFFFF80000000) to generate the notification. This value is not found in the notifications hashtable, causing loss of notifications and thus a hang of the affected Lwt tasks.

All uses of notification id's should be long to match the ocaml value's size.

aantron:
This patch types all C notification id fields, arguments, and variables at type `intnat`, which is exposed by `caml/mlvalues.h`, and ultimately comes from `caml/config.h`. `intnat` is the C data type used by the OCaml runtime for storing OCaml `int`s. Its definition is selected according to the machine word size and the C compiler detected during configuration of OCaml.

The corresponding conversion macros between `intnat` and `value` are `Long_val` and `Val_long`.

Fixes #277.

cc @stijn-devriendt

I was able to reproduce the problem as described in the issue. This patch eliminates it.

Hope I got everything. I did not test the signal code; I mostly proceeded by static analysis.

### DIFF
--- a/src/unix/lwt_unix.h
+++ b/src/unix/lwt_unix.h
@@ -81,7 +81,7 @@ void lwt_unix_not_available(char const *feature) Noreturn;
    +-----------------------------------------------------------------+ */
 
 /* Sends a notification for the given id. */
-void lwt_unix_send_notification(int id);
+void lwt_unix_send_notification(intnat id);
 
 /* +-----------------------------------------------------------------+
    | Threading                                                       |
@@ -178,7 +178,7 @@ struct lwt_unix_job {
 
   /* Id used to notify the main thread in case the job do not
      terminate immediately. */
-  int notification_id;
+  intnat notification_id;
 
   /* The function to call to do the work.
 
--- a/src/unix/lwt_unix.ml
+++ b/src/unix/lwt_unix.ml
@@ -92,7 +92,9 @@ module Notifiers = Hashtbl.Make(struct
 
 let notifiers = Notifiers.create 1024
 
-let current_notification_id = ref 0
+(* See https://github.com/ocsigen/lwt/issues/277 and
+   https://github.com/ocsigen/lwt/pull/278. *)
+let current_notification_id = ref (0x7FFFFFFF - 1000)
 
 let rec find_free_id id =
   if Notifiers.mem notifiers id then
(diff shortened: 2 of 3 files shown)


stijn-devriendt:
Patch looks OK to me. It's a superset of the internally proposed fix.

Testing this without initializing the current_notification_id to a sufficiently high number is hard to do. We didn't see this issue until after a month of very intensive testing. Perhaps it makes sense to always initialize the value at a high number to catch these kinds of errors (and make them testable in the test suite). This is also what the Linux kernel does internally with the monotonic time, initializing it at -5min.

Thanks for the quick turnaround!

aantron:
Thanks for the look. I was also thinking to always initialize the id as in your issue. I'll amend to put that back in and merge.

aantron:
Actually, I remember why I didn't want to do that – I don't think it makes sense on 32-bit systems. I'm not even sure it would compile. Looking into it...

aantron:
Ok, `0x7FFFFFFF` luckily fits into 31 bits, it's just -1 in 31-bit two's complement. This [works](https://ci.appveyor.com/project/aantron/lwt/build/153/job/r3p0l248ayb33xue) on 32-bit systems. It doesn't really test the problem in the issue, but then again that problem would only be expected on 64-bit systems.

stijn-devriendt:
The unfortunate issue is that the initialization value can test 2 things:
- this issue, when initialized slightly below 0x7FFFFFFF
- type overflow issues, when initialized according to the actual type (e.g. -1000)

Having this bug fixed, would leave only the second group open, but wouldn't find any regressions in the first. So, unless you can force/inject this value during testing, all you can do is pick one category and hope the other doesn't regress.

aantron:
Yes, I am guessing that it is much more likely for someone (perhaps me in the future) to forget to use `intnat` somewhere, and thus potentially reintroduce truncation, than it is for someone to redesign the C and/or OCaml code so that notification ids are no longer opaque and order and overflow start to matter. So, I think that setting the initial value to catch truncation in 64-bit systems is good enough for now.

PR #287, opened 2016-11-01
Merge some patterns which execute the same code.

rneswold:
Some minor changes. Not sure whether the compiler would merge these cases, but now the code clearly says the same code handles the two states of the thread.


aantron:
Thank you.

PR #288, opened 2016-11-01
Convert 'Let.return ()' to 'Lwt.return_unit'.

rneswold:
`for%lwt` and `while%lwt` should use `Lwt.return_unit` to avoid extra allocations.

### DIFF
--- a/src/ppx/ppx_lwt_ex.ml
+++ b/src/ppx/ppx_lwt_ex.ml
@@ -158,7 +158,7 @@ let lwt_expression mapper exp attributes =
   (** [while%lwt $cond$ do $body$ done] ≡
       [let rec __ppx_lwt_loop () =
          if $cond$ then Lwt.bind $body$ __ppx_lwt_loop
-         else Lwt.return ()
+         else Lwt.return_unit
        in __ppx_lwt_loop]
    *)
   | Pexp_while (cond, body) ->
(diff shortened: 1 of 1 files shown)


aantron:
Thanks again!

These [may be deprecated](https://github.com/ocsigen/lwt/wiki/Plan#flambda-and-lwtreturn_xyz-easy) in the future.

FYI GH isn't associating your recent commits with your username, but you've committed with a different email before. Not sure if this is intentional/it matters to you, just a heads up.

agarwal:
Might be a good idea to add a `.mailmap file to this project.

aantron:
That's cool, didn't know about `.mailmap` before. Does GitHub read and use it?

rgrinberg:
While I agree that flambda will make the various `return_*` functions useless in the future. I don't quite see the benefit of getting rid of them as aliases. All it would is cause pointless breakage for packages down the road without any real gain. As after all, there's nothing harmful about these specializations and they don't impact the evolution of Lwt in any way.

rneswold:
@aantron :

> These may be deprecated in the future.

Cool! Good to know.

> FYI GH isn't associating your recent commits with your username, but you've committed with a different email before. Not sure if this is intentional/it matters to you, just a heads up.

Thanks. Although I created this as a personal account, I've been using my repos at work so, occasionally, changes are being associated with my work email.

I added it as an additional address, so maybe it'll be less confusing (not sure what GH does with multiple email addresses.)

aantron:
@rgrinberg: Agreed. I will not actually remove them while I suspect there is code using them, that is compatible with the contemporary Lwt. I would guess the time scale for removing them is years to forever. I will fix the wiki :p

@rneswold: Seems adding the email fixed it :)

agarwal:
> That's cool, didn't know about .mailmap before. Does GitHub read and use it?

Not sure if GitHub supports it, but it is supported by `git`.

Issue #293, opened 2016-11-21
Semantic versioning; safely breaking Lwt (proposal)


aantron:
I suggest:

- After `2.7.0`, Lwt will follow semantic versioning. Any breaking changes will bump the major version number. I expect this will happen twice per year at most. The actual changes will typically be minor (e.g. #230, #269, #260). They will be batched.

How breaking changes will be done:

- An API that is being changed will be versioned. So, if `Lwt_unix.foo` is to be changed, there will be `Lwt_unix.Versioned.foo_1` and `Lwt_unix.Versioned.foo_2`. A detailed deprecation annotation on `foo` will say:

  - If you don't wish to risk refactoring your code, use `foo_1` – an alias for the current `foo`.
  - To get the new functionality immediately, use `foo_2`.

- This can be released right away.
- I (or other maintainer(s)) will notify the maintainers of any packages in OPAM that are using `foo`.
- After being released deprecated for at least, say, 3 months, `foo` will become an alias for `foo_2` instead of `foo_1`, during a major release. This will be clearly noted in the [change log](https://github.com/ocsigen/lwt/blob/master/CHANGES).
- `foo_1` and `foo_2` will be marked deprecated, and will be guaranteed to persist in Lwt for at least, say, one more major release. After that, each will remain indefinitely, unless/until further maintenance becomes an unreasonable burden. This also applies to APIs that are deprecated but not broken (i.e. whose names are not being reused).

aantron:
I guess this can be closed, now that 3.0.0 is out, and this process is put in practice (though it is not yet complete, as Lwt may have to still support users that missed the breakage window).

PR #294, opened 2016-11-24
Add support for selecting the libev backend (more gently)

aantron:
This is a version of #269 that warns current users of `Lwt_engine.libev` that the signature will change, as per #293.

- The new version is available as `Lwt_engine.Versioned.libev_2`. It will be released in the next Lwt release, but will replace the current `libev` in a major release.
- The current `libev` has a warning attached to it in the meantime:

  ```
  Warning 3: deprecated: Lwt_engine.libev
  This class will soon have parameters for selecting a libev backend. This will
  be a breaking change. See
    https://github.com/ocsigen/lwt/pull/269
  To preserve the current signature, use Lwt_engine.Versioned.libev_1
  To use the replacement immediately, use Lwt_engine.Versioned.libev_2 ()
  ```

Resolves #269

cc @yallop I modified the first commit slightly:
  - renamed `libev` to `libev'` in the `.ml` file,
  - defined `libev` in terms of that, and
  - undid the change to the signature of `libev`

Hope that is ok.

### DIFF
--- a/src/unix/lwt_engine.ml
+++ b/src/unix/lwt_engine.ml
@@ -128,7 +128,27 @@ type ev_loop
 type ev_io
 type ev_timer
 
-external ev_init : unit -> ev_loop = "lwt_libev_init"
+module Ev_backend =
+struct
+  type t =
+    | EV_DEFAULT
+    | EV_SELECT
+    | EV_POLL
+    | EV_EPOLL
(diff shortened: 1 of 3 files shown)


yallop:
Yes, that's fine.  Thanks for taking care with git history & authorship!

PR #295, opened 2016-11-26
Deprecate Lwt_stream.result

aantron:
And prepare to break `Lwt_stream.map_exn`, which uses this type.

Users of the current `map_exn` will see:

```
Warning 3: deprecated: Lwt_stream.map_exn
This function will soon use Lwt.result instead of Lwt_stream.result.
To continue using the present signature, use Lwt_stream.Versioned.map_exn_1
To use the new variant immediately, use Lwt_stream.Versioned.map_exn_2
```

cc @c-cube more `result` meltdown :P

### DIFF
--- a/src/core/lwt_stream.ml
+++ b/src/core/lwt_stream.ml
@@ -469,6 +469,22 @@ let rec get_exn_rec s node =
 
 let map_exn s = from (fun () -> get_exn_rec s s.node)
 
+let rec get_exn_rec' s node =
+  if node == !(s.last) then
+    Lwt.try_bind
+      (fun () -> feed s)
+      (fun () -> get_exn_rec' s node)
+      (fun exn -> Lwt.return (Some (Result.Error exn)))
+  else
+    match node.data with
+      | Some value ->
+          consume s node;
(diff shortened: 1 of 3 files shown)


Drup:
`Lwt_stream.Versioned` ? *shudder a little bit*.

That's not a fabulous way of making breaking change. I'm pretty sure you could just come up with a better name (`wrap_all` ?) and make the transition like that.

vasilisp:
@Drup, see #293 if you haven't already.

aantron:
@vasilisp, @Drup Thanks. #293 is how I intend to make breaking changes if we want to keep the name. 

But, if `map_exn` is the wrong name for this function to begin with, fair enough. Then, if we agree on a good name for the new version (will take a look at `wrap`/`wrap_exn` later, preoccupied with something), I will only deprecate `map_exn` and the type and move them to the bottom of the API docs, and keep them there for a very long time. No need for `.Versioned` in that case.

aantron:
Renamed it to `wrap_exn`. There is no `.Versioned` module anymore.

PR #296, opened 2016-11-28
Cooperative Lwt_unix.bind (sockets)

### LINKED ISSUE #230: Lwt_unix.bind should be in the monad
``` ocaml
Lwt_unix.(bind sock (ADDR_UNIX sock_path)) : unit
```

could block indefinitely due to FS `mknod` on `sock_path`.

aantron:
Resolves #230.

Making `Lwt_unix.bind` cooperative is a breaking change, so:

1. The existing `bind` is deprecated, starting in the [next release](https://github.com/ocsigen/lwt/wiki/Plan#270-around-1-january-2017), with this message:

```
Warning 3: deprecated: Lwt_unix.bind
This function will soon return threads (-> unit Lwt.t), because the bind system
call can block on Unix filesystem sockets. See
  https://github.com/ocsigen/lwt/issues/230
To keep using the current signature, use Lwt_unix.Versioned.bind_1
To use the new non-blocking version immediately, use Lwt_unix.Versioned.bind_2
```

2. `Lwt_unix.Versioned.bind_2`, added by this PR, will replace `Lwt_unix.bind` in Lwt [3.0.0](https://github.com/ocsigen/lwt/wiki/Plan#300-around-1-april-2017).

cc @dsheets


aantron:
I think this change has to affect the new `establish_server` as well. It should have result type `server Lwt.t`.

Issue #41, opened 2014-02-14
Distributing lwt as several opam packages


trefis:
Hi!

Nowadays I (as well as several other people I would guess) am using opam for installing basically every ocaml library I use and among them is lwt.
However, when installing lwt opam will by default only install core (and unix most probably), but not the sublibraries which depend on some optional dependencies (unless these are present, ofc).

This creates (imho) two problems : 
1. Lwt (and the other packages depending on it) gets recompiled "for no reason". For example on a fresh opam switch I'll always do "opam install utop", which will pull lwt and lambda-term (among others). At somepoint in the future I'll install a package which depends on ocaml-ssl and that will trigger the recompilation of lwt (with ssl enabled), lambda-term and utop.
   That kind of problem was already raised about Coq ( https://github.com/ocaml/opam-repository/issues/805 ), and although it's not as costly here it still is a pain.
2. Some package will indicate lwt as their dependency while they actually want lwt-text. That can obvious be fixed by the packager and probably doesn't happen much on the official opam repository as people are keeping a close watch, but still it's annoying / gives "more work" for the packager.

What I would like to see is lwt distributed as several packages : lwt, lwt-ssl, lwt-text, lwt-react, ...

What do you think?

aantron:
I think I will begin doing this for `Lwt_ssl` and `Lwt_glib`, as those are really separate libraries that just happen to be in the same repo.

The process will be something like:

1. In the next Lwt release, release `lwt`, `lwt_ssl`, `lwt_glib`. The `lwt` package won't change. `lwt_ssl` will install `ssl` and `lwt`, and provide Findlib package `lwt_ssl`, which will just be a proxy for `lwt.ssl`. Users will be warned that `lwt.ssl` is deprecated, and to use both Findlib and OPAM packages `lwt_ssl`. Perhaps I will PR Findlib and try to get a deprecation warning feature in, so that it can be displayed automatically.
2. After being released in this state for a while, in a subsequent major release, package `lwt_ssl` will contain the real `Lwt_ssl`, while `lwt.ssl` from `lwt` will contain a dummy `Lwt_ssl` that contains an `[@@@ocaml.deprecated "..."]` annotation with the info on how to get the real `Lwt_ssl`.

Likewise for `Lwt_glib`.

cc @gerdstolpmann

aantron:
[...]

PR #301, opened 2016-12-06
Begin factoring out Lwt_ssl, Lwt_glib, Lwt_react

aantron:
First step, as described in [#41](https://github.com/ocsigen/lwt/issues/41#issuecomment-263635674).

Not sure if we want to also do this to `Lwt_react`. After `Lwt_ssl` and `Lwt_glib` are factored out, `react` will be the last "true" optional dependency of Lwt.

cc @trefis


rgrinberg:
My vote is +1 on react. It's more friendly towards downstream packages as it allows them to describe dependencies more precisely.

aantron:
> +1 on react

Alright, I tend to agree. Might as well get it over with. Will add that to the PR soon.

Issue #33, opened 2014-01-11
configure fails with misleading error message if curses is missing


timbertson:
I installed both pthread and libev via system packages (Fedora 20), so they ought to be in reasonable locations. But I still get "The following recquired C libraries are missing: pthread, libev." when I run `ocaml setup.ml -configure --enable-unix`

Digging into discover.ml, I modified it to print the location of log_file and not to delete it. Aside: it seems rather awkward to debug this sort of build failure, as everything gets cleaned up immediately even in the case of failure. Unless I've missed a -debug argument or something.

Anyway, the log says:

```
/tmp/lwt_stub35ffd5.c: In function ‘lwt_test’:
/tmp/lwt_stub35ffd5.c:7:3: warning: null argument where non-null required (argument 1) [-Wnonnull]
   pthread_create(0, 0, 0, 0);
   ^
/tmp/lwt_stub35ffd5.c:7:3: warning: null argument where non-null required (argument 3) [-Wnonnull]
/usr/bin/ld: cannot find -lcurses
collect2: error: ld returned 1 exit status
File "/tmp/lwt_caml396824.ml", line 1:
Error: Error while building custom runtime system
```

So I've installed ncurses-devel, and that fixed it. To be honest I'm not sure _what_ is requiring -lcurses since the word "curses" doesn't appear in the lwt repo, but I thought you'd like to know.

PR #313, opened 2017-02-08
discover: Sanity check compilation before checking for missing dependencies

timbertson:
I have a bit of a novel environment in [opam2nix][] where only the (transitive) dependencies are made available to each package while building. Because of this, `lwt` does not have `ncurses` available unless it depends transitively on on an `ncurses` depext (or `conf-ncurses`). There's been a lot of back and forth on where the dependency on `ncurses` belongs, but it hopefully won't be contentious to fail earlier if a basic toplevel can't be compiled.

This PR gives a better error message - currently `lwt` tells me that "pthread" isn't available, because that's just the first thing it attempts to compile. This is misleading, since it would tell me everything is unavailable except for the real problem (in my case, ncurses).

[opam2nix]: https://github.com/timbertson/opam2nix-packages

### DIFF
--- a/src/util/discover.ml
+++ b/src/util/discover.ml
@@ -94,6 +94,15 @@ external test : unit -> unit = \"lwt_test\"
 let () = test ()
 "
 
+let trivial_code = "
+#include <caml/mlvalues.h>
+
+CAMLprim value lwt_test(value Unit)
+{
+  return Val_unit;
+}
+"
+
 let pthread_code = "
 #include <caml/mlvalues.h>
 #include <pthread.h>
@@ -549,6 +558,10 @@ let () =
   let have_pkg_config = !not_available = [] in
   not_available := [];
 
+  let test_basic_compilation () =
+    test_code ([], []) trivial_code
+  in
+
   let test_libev () =
     let opt, lib =
       lib_flags "LIBEV"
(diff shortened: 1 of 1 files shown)


timbertson:
(heh, somehow I manage to keep hitting this - from three years ago: https://github.com/ocsigen/lwt/issues/33)

As a related matter, I just checked the opam-repository and it's unfortunately very divided on whether `lwt` needs curses or not:

```
$ find . -name opam | xargs grep --files-with-match curses | sort
./lwt.2.4.5/opam
./lwt.2.4.6/opam
./lwt.2.4.7/opam
./lwt.2.4.8/opam
./lwt.2.5.0/opam
./lwt.2.5.1/opam
./lwt.2.5.2/opam

$ find . -name opam | xargs grep --files-without-match curses | sort
./lwt.2.3.2/opam
./lwt.2.4.0/opam
./lwt.2.4.1/opam
./lwt.2.4.2/opam
./lwt.2.4.3/opam
./lwt.2.4.4/opam
./lwt.2.6.0/opam
./lwt.2.7.0/opam
```

I know `lwt` only depends on curses because it builds a custom toplevel, but it would be good to be consistent (and obviously I'd vote for including the dependency, since that makes my life easier ;))

rgrinberg:
I know the toplevel is deprecated and probably only on life support at this point. But it would be nice if it was its own opam sub package as well as I'm not sure how long we'd wait for 3.0.

aantron:
Thanks for this, and the other PR.

I'm pretty sure `simple-top` isn't used anymore, and we will just remove it completely, in the next major version of Lwt (4.0.0, I guess). Of course, that will be preceded by proper warnings and announcements.

3.0.0 is coming out this weekend (hopefully).

PR #314, opened 2017-02-08
discover: Print more information on failure

timbertson:
Just putting this here in case it's useful. Without peeking into discover.ml I'd have no idea how to get at the logs when configure fails. This tells you to set `DEBUG=1` if the logs are going to be deleted, and once you _have_ done that it'll print the location of the log on failure.


aantron:
Thanks. The shadowing of `exit` is probably not that great, but we will have to take a look at improving `discover.ml` anyway in the future, and we can pick a better name then.

PR #317, opened 2017-02-18
Fix typo

ryanslade:


Drup:
Not sure what's up with CI, but I'll merge as this is obviously right. Thanks!

mfp:
AppVeyor fails in one test out of 3. It [looks like a spurious error](https://ci.appveyor.com/project/aantron/lwt/build/340/job/3ahjtu5cvr5n34w8#L931) and I can't see why it would fail to find `topfind` when `ocamlfind` has been clearly installed, and the environment initialized with `eval $(opam config env)`; moreover, a nearly identical CI build that only differs in `libev-devel` not being installed succeeded (and that above failure is not at Lwt build time, it happens later when building the tests).

The [Travis error](https://aantron.github.io/binaries/) is more interesting:

```
Running 47 tests for library "unix".
(33/47) Running test "establish_server_2: order of exceptions" from suite "lwt_io"Fatal error: exception Unix.Unix_error(Unix.EBADF, "close", "")
W: Test 'unix' fails: Command '/home/travis/.opam/system/build/lwt.dev/_build/tests/unix/main.native' terminated with error code 2
```

That `EBADF` error looks like a double-close.

aantron:
I temporarily disabled that AppVeyor row, as I don't have time immediately to look into it, and that build row is not that valuable over the others: https://github.com/ocsigen/lwt/commit/2413635038457d50b8d1fd69f747d7b32fa80fb6.

The Travis error occurs sporadically. I used to get it "almost never," but now it seems to have become a fairly regular occurrence. I will take a hard look at that hacky test, and either find some way to rewrite it, or delete it completely. If it continues causing problems in the meantime, you're welcome to remove the test, remove the build row, or put it in an `allow_failures` clause in `.travis.yml`, as temporary measures.

aantron:
Oh, and thanks for the fix :)

Issue #321, opened 2017-02-28
Compilation against 4.05


vasilisp:
`Lwt_unix` doesn't compile against OCaml 4.05 (currently in beta). The reason is an additional constructor in `Unix.open_flag` (`O_KEEPEXEC`).

This brings up quite an ugly situation. The constructor aliases could be removed, but that would break a lot of code. Alternatively, some conditional compilation mechanism could be used (and that's not exactly elegant). And of course some of the current packages in OPAM need to be fixed (with separate `+4.05` versions?).

mfp:
Unpleasant as it is, conditional compilation (with cppo, I'd add) is the de-facto standard way to cope with incompatibilities across OCaml versions (most often in ppx extensions). AFAICS no separate OPAM packages are required if we go that route, and no existing code (that wouldn't break on its own anyway with OCaml 4.05) should break.

aantron:
Should be fixed now in `master` by #322, and will be released in the coming days to OPAM.

Issue #170, opened 2015-07-06
Ocaml 4.02.2 and Lwt_unix.O_CLOEXEC


domsj:
The following fails in utop as of ocaml 4.02.2 (works just fine with 4.02.1)
```
Lwt_unix.openfile
  "/tmp/test"
  [Lwt_unix.O_RDWR; Lwt_unix.O_CLOEXEC; Lwt_unix.O_CREAT; Lwt_unix.O_EXCL] 0o600
 >>= fun fd ->
Lwt_unix.write fd "a" 0 1;;
Exception: Unix.Unix_error (Unix.EBADF, "write", ""). 
```

It not longer fails when I remove the O_CLOEXEC flag.

Doing the same things with (blocking) functions from the standard Unix module does not have this problem...

vasilisp:
I can reproduce this, with or without `utop`. I have no idea what is going on.

vbmithr:
Should be fixed with 2.5.1, please try!

rgrinberg:
:100: @vbmithr thanks for the release

PR #322, opened 2017-02-28
Use conditional compilation for Lwt_unix + OCaml 4.05 compat (O_KEEPEXEC)

mfp:
Compatibility with OCaml 4.05, broken with the introduction of the `O_KEEPEXEC` constructor in `Unix.open_flag`, see #321.

Tested on 4.02.3 and 4.05+beta2.
I'm getting these apparently benign warnings, which I have not been able to prevent (I didn't find the proper way to use InterfacePatterns/ImplementationPatterns apparently):

```
W: Cannot find source file matching module 'Lwt_unix' in library lwt-unix.
W: Use InterfacePatterns or ImplementationPatterns to define this file with feature "source_patterns".
W: No exported module defined for library lwt-simple-top
W: No exported module defined for library lwt-syntax-options
W: Cannot find source file matching module 'Lwt_unix' in library lwt-unix.
W: Use InterfacePatterns or ImplementationPatterns to define this file with feature "source_patterns".
W: No exported module defined for library lwt-simple-top
W: No exported module defined for library lwt-syntax-options
W: Cannot find source file matching module 'Lwt_unix' in library lwt-unix.
W: Use InterfacePatterns or ImplementationPatterns to define this file with feature "source_patterns".
W: Cannot find source file matching module 'Lwt_unix' in library lwt-unix.
W: Use InterfacePatterns or ImplementationPatterns to define this file with feature "source_patterns".
```

### DIFF
--- a/lwt.opam
+++ b/lwt.opam
@@ -35,6 +35,7 @@ depends: [
   "ocamlfind" {build & >= "1.5.0"}
   "ocamlbuild" {build}
   "result"
+  "cppo"
   # See https://github.com/ocsigen/lwt/issues/266
   ( "base-no-ppx" | "ppx_tools" {build} )
   ## OASIS is not required in released version
--- a/lwt.opam
+++ b/lwt.opam
@@ -35,7 +35,7 @@ depends: [
   "ocamlfind" {build & >= "1.5.0"}
   "ocamlbuild" {build}
   "result"
-  "cppo"
+  "cppo" {build}
   # See https://github.com/ocsigen/lwt/issues/266
   ( "base-no-ppx" | "ppx_tools" {build} )
   ## OASIS is not required in released version
--- a/lwt.opam
+++ b/lwt.opam
@@ -39,7 +39,7 @@ depends: [
   # See https://github.com/ocsigen/lwt/issues/266
   ( "base-no-ppx" | "ppx_tools" {build} )
   ## OASIS is not required in released version
-  "oasis" {build & >= "0.4.7"}
+  "oasis" {build & >= "0.4.8"}
 ]
 depopts: [
   "base-threads"
--- a/src/unix/lwt_unix_unix.c
+++ b/src/unix/lwt_unix_unix.c
@@ -26,6 +26,8 @@
 
 #define ARGS(args...) args
 
+#include <caml/version.h>
+#include <caml/unixsupport.h>
 #include <sys/uio.h>
 #include <sys/un.h>
 #include <sys/time.h>
@@ -1124,29 +1126,24 @@ static int open_flag_table[] = {
   O_SYNC,
   O_RSYNC,
   0,
-#ifdef O_CLOEXEC
-  O_CLOEXEC,
-#else
-#define NEED_CLOEXEC_EMULATION
-  0,
-#endif
-  0
+  0, /* O_CLOEXEC, treated specially */
+  0  /* O_KEEPEXEC, treated specially */
 };
 
-#ifdef NEED_CLOEXEC_EMULATION
+enum { CLOEXEC = 1, KEEPEXEC = 2 };
+
 static int open_cloexec_table[15] = {
   0, 0, 0, 0, 0, 0, 0, 0,
   0, 0, 0, 0,
   0,
-  1, 0
+  CLOEXEC, KEEPEXEC
 };
-#endif
 
 struct job_open {
   struct lwt_unix_job job;
   int flags;
   int perms;
-  int fd;
+  int fd; /* will have value CLOEXEC or KEEPEXEC on entry to worker_open */
   int blocking;
   int error_code;
   char *name;
@@ -1156,10 +1153,28 @@ struct job_open {
 static void worker_open(struct job_open *job)
 {
   int fd;
+  int cloexec;
+
+  if (job->fd & CLOEXEC)
+    cloexec = 1;
+  else if (job->fd & KEEPEXEC)
+    cloexec = 0;
+  else
+#if OCAML_VERSION_MAJOR >= 4 && OCAML_VERSION_MINOR >= 5
+    cloexec = unix_cloexec_default;
+#else
+    cloexec = 0;
+#endif
+
+#if !defined(NEED_CLOEXEC_EMULATION)
+  if (cloexec) job->flags |= O_CLOEXEC;
+#endif
+
   fd = open(job->name, job->flags, job->perms);
 #if defined(NEED_CLOEXEC_EMULATION) && defined(FD_CLOEXEC)
-  if (fd >= 0 && job->fd) {
+  if (fd >= 0 && cloexec) {
     int flags = fcntl(fd, F_GETFD, 0);
+
     if (flags == -1 ||
         fcntl(fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
       int serrno = errno;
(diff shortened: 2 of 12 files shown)


mfp:
hmm no CI? Are all the bots sleeping, or is CI skipped if the branch is created directly on this repos?

lwt.opam
```diff
@@ -35,6 +35,7 @@ depends: [
   "ocamlfind" {build & >= "1.5.0"}
   "ocamlbuild" {build}
   "result"
+  "cppo"
```
vasilisp: You can add the `{build}` predicate here.

aantron:
`XOCamlbuildPluginTags` was broken in OASIS 0.4.7 and fixed in 0.4.8 (https://github.com/ocaml/oasis/pull/84, [changelog](https://github.com/ocaml/oasis/blob/cb3cc3442da92d4ba7221e2da571a27a2c72cfdf/CHANGES.txt#L9)), so should also change the `oasis` constraint in `opam` to `>= 0.4.8`. Looks good modulo this and @vasilisp's comment.

Not sure what just happened with the AppVeyor CI. I don't have time to debug it at the moment. I restarted the builds. If you keep having problems, you're welcome to comment out build rows/disable AppVeyor, and I will look later.

aantron:
And

> hmm no CI? Are all the bots sleeping, or is CI skipped if the branch is created directly on this repos?

Maybe it was just taking a while to be submitted by GitHub. CI is enabled for branches created directly on the repo. In fact, I'd like to disable double building of PRs from branches on the repo, but neither AppVeyor nor Travis supported this as of two months ago.

lwt.opam
mfp: Thank you, pushed.

mfp:
Thank you both for the extra eyeballs :)

Indeed CI eventually launched after a while (around an hour an a half or so) -- I was used to Github doing its thing faster.

Will sit on this for a few days (4.05 is not released yet) before merging just in case we/somebody find a reason not to.

aantron:
@mfp I pushed into this branch a commit to enable testing on `4.05+beta2` in Travis. Hopefully, that will go well.

Also, I think it will be necessary to add `< 4.05` compiler constraints to all the existing `lwt` packages in OPAM. If you don't want to do this tedious work, I will do it some time before the 3.0.0 release :)

mfp:
4.05 CI: was doing the same in a (now deleted) branch :)

Most lwt versions had more conservative constraints (4.02 or 4.03), so there weren't that many in need of a change. https://github.com/ocaml/opam-repository/pull/8578

aantron:
Looks like we need to wait on Camlp4 for 4.05 to be in OPAM for the CI to work. Fortunately, there is a PR: https://github.com/ocaml/opam-repository/pull/8583. I'll restart the build when that's in.

Thanks for the constraints.

> Surprisingly, 2.4.5 builds on 4.04

2.4.5 builds on 4.04 because that was the last old release before the Lwt PPX:

```
$ git tag --contains `git log --reverse --format='%H' -- ppx | head -n 1`                                                                      141
2.4.6
2.4.7
2.4.8
2.5.0
2.5.1
2.5.2
2.6.0
2.7.0
```

There were some incompatibilities in the PPX when 4.03 came out, fixed in Lwt 2.5.2. I guess the PR was https://github.com/ocsigen/lwt/pull/227.

aantron:
Should be ready for merge now.

fdopen:
Don't forget to update the c stub, it already happend in the past at this point: https://github.com/ocsigen/lwt/issues/170

The sizes of https://github.com/ocaml/ocaml/blob/dd0d93f0f1b2c9e9c91713ae3aac0accbbf82310/otherlibs/unix/open.c#L41 and https://github.com/ocsigen/lwt/blob/0f37f8e8c970cec9b22588ee9b2034e141bbbb17/src/unix/lwt_unix_unix.c#L1113

must be kept in sync. Otherwise `caml_convert_flag_list` could return garbage, if you pass `O_KEEPEXEC` to it.

mfp:
On Mon, Mar 06, 2017 at 04:43:44AM -0800, Andreas Hauptmann wrote:
> Don't forget to update the c stub, it already happend in the past at this point: https://github.com/ocsigen/lwt/issues/170
> 
> The sizes of https://github.com/ocaml/ocaml/blob/dd0d93f0f1b2c9e9c91713ae3aac0accbbf82310/otherlibs/unix/open.c#L41 and https://github.com/ocsigen/lwt/blob/0f37f8e8c970cec9b22588ee9b2034e141bbbb17/src/unix/lwt_unix_unix.c#L1113
> 
> must be kept in sync. Otherwise `caml_convert_flag_list` could return garbage, if you pass `O_KEEPEXEC` to it.

Great catch, thanks a lot!
I'm updating the PR.

-- 
Mauricio Fernández

avsm:
We're now using ocaml-migrate-parsetree in cstruct, so the bulk build train for 4.05 is now blocked on Lwt again :-) https://github.com/mirage/ocaml-cstruct/pull/127

Would be great to get this built into a release so that the 4.05 betas work.

aantron:
This is going into a `2.x.x` release within the next two weeks or so, to be released concurrently with `3.0.0` (where `3.0.0` will contain only the planned breaking changes).

mfp:
On Wed, Mar 22, 2017 at 10:15:54AM -0700, Anil Madhavapeddy wrote:
> We're now using ocaml-migrate-parsetree in cstruct, so the bulk build train for 4.05 is now blocked on Lwt again :-) https://github.com/mirage/ocaml-cstruct/pull/127
> 
> Would be great to get this built into a release so that the 4.05 betas work. 

Sorry about the delay. I'm planning to revisit the patch this weekend.

-- 
Mauricio Fernández

mfp:
I tried to write some tests for O_CLOEXEC/O_KEEPEXEC but frankly it was a PITA: pre-processor needed to test 4.05's O_KEEPEXEC, platform-dependent code to test `Lwt_unix.fork + exec` on Unix-like and CreateProcess on win32 separately, etc.

aantron:
I think writing a private one-off program for testing, and exercising the various combinations once locally, is fine for now.

src/unix/lwt_unix_unix.c
```diff
@@ -1156,10 +1153,28 @@ struct job_open {
...
+  else if (job->fd & KEEPEXEC)
+    cloexec = 0;
+  else
+#if OCAML_VERSION_MAJOR >= 4 && OCAML_VERSION_MINOR >= 5
+    cloexec = unix_cloexec_default;
+#else
+    cloexec = 0;
+#endif
+
+#if !defined(NEED_CLOEXEC_EMULATION)
```
aantron: As I understand it, with this diff, `NEED_CLOEXEC_EMULATION` will no longer ever be defined.

mfp: Indeed. I lost the line that defined `NEED_CLOEXEC_EMULATION` if `O_CLOEXEC` wasn't defined in the last-second patch cleanup, and missed it Linux + glibc does have `O_CLOEXEC`.
Should be fixed now.

mfp:
I bit the bullet and tested both `O_CLOEXEC` and `O_KEEPEXEC` (when available) on non-Windows.
On win32, `Lwt_unix.openfile` uses `Unix.openfile`, so there's not that much to gain in going out of our way to test it, anyway.

tests/unix/test_lwt_unix.cppo.ml
```diff
@@ -22,6 +22,46 @@
...
+          Unix.putenv assertion (string_of_int @@ Obj.magic fd_);
+          (* There's no portable way to obtain the executable name (which
+           * may even no longer exist at this point), but argv[0] fortunately
+           * has the right value when the tests are run with "make test". *)
+          Unix.execv Sys.argv.(0) [||]
+      | n ->
+          Lwt_unix.close fd >>= fun () ->
+          Lwt_unix.waitpid [] n >>= function
+            | _, Unix.WEXITED 0 -> Lwt.return_true
+            | _ -> Lwt.return_false
```
aantron: It looks like this is failing the automated builds with warning 4, because it does not list the cases for the `Unix.process_status` explicitly, and we have lots of warnings enabled, and build with warnings as errors in CI. I think in this case, it's best to just list the two remaining possible cases explicitly, so we fail out in the extremely unlikely event `Unix.process_status` changes one day.

aantron:
Much appreciated!

Issue #308, opened 2017-01-03
Lwt 3.0.0 breaking changes notice


aantron:
Lwt 3.0.0 will be released near 1 April with minor breaking changes to several APIs. This issue is for informing maintainers of OPAM packages that use those APIs.

Each API that is breaking has been given a deprecation warning in the recent 2.7.0 release. Versioned APIs are provided so you can optionally make a smooth transition. If you switch to one of the versioned APIs, you should add constraint `lwt {>= "2.7.0"}`. Otherwise, you should add `lwt {< "3.0.0"}` to your current code.

If you need to temporarily suppress a deprecation warning on an expression `e`, use

```ocaml
(e [@ocaml.warning "-3"])
```

*I will cc the package maintainers once Lwt 2.7.0 and the factored-out packages are installable from OPAM.*

---

- [`Lwt_unix.bind`](http://ocsigen.org/lwt/dev/api/Lwt_unix#VALbind) (#230, #296):

    ```
    Warning 3: deprecated: Lwt_unix.bind
     This function will soon evaluate to a promise (-> unit Lwt.t), because the
     bind system call can block for Unix domain sockets. See
       https://github.com/ocsigen/lwt/issues/230
     This will be a breaking change in Lwt 3.0.0.
     If you don't use Unix domain sockets and use Lwt_unix.bind ... ; rather than
     let () = Lwt_unix.bind ... in, you can ignore this warning.
     To retain the current signature, use Lwt_unix.Versioned.bind_1
     To use the new non-blocking version immediately, use Lwt_unix.Versioned.bind_2
     Both alternatives require Lwt >= 2.7.0.
    ```

    Since `Lwt_unix.bind` currently evaluates to unit, and the new `bind` internally has the same synchronous semantics *except* for Unix domain sockets, it is possible to do nothing if you are not using Unix domain sockets and use `;` for sequencing `bind`. However, you will start getting warnings like this after Lwt 3.0.0:

    ```
    Warning 10: this expression should have type unit
    ```

    The semantics are not guaranteed for the future, so this is not recommended.

    `Lwt_unix.bind` is used in packages:

[...]

PR #334, opened 2017-04-08
3.0.0: the breaking changes (OCaml APIs)

aantron:
This PR would commit the breaking changes listed in #308, excluding the one about factoring out packages (the last one). I will make a separate PR for that, as it's a pretty different type of change.

The commits mostly consist of renaming values and editing deprecation messages.


mfp:
LGTM

The build failures on Mac seem spurious (timeout while updating MacPorts' port tree). Since the PR mostly renames stuff, the tests pass in other systems, and 2.7.0, which already included the versioned code was OK, we can assume no breakage happened here.

aantron:
Thanks. Agreed on MacPorts. That has happened before in Travis, even though the local MacPorts was doing fine. I restarted those jobs, for a final attempt.

PR #335, opened 2017-04-09
3.0.0: packaging: factor out packages to lessen depopts

aantron:
This PR would commit the change to packaging described in #308.

I gave each of `lwt_ssl`, `lwt_glib`, and `lwt_react` their own little build systems, with `_oasis` files. OASIS seemed like the easiest way to get `lwt_glib` compiling with all the right flags for the external dependency, and the easiest way to generate all the various output files. These new build systems are only used during *release* packaging. It is still possible (and easier during *development*) to build these packages from the main build system, by supplying the right `configure` flags.

The main `opam` file is no longer aware of these flags, however, so it is not possible to accidentally build the extra packages when installing package `lwt`. The `opam` depopts for these extra packages have been removed. I kept the `opam` messages, however. The idea is that if someone does, e.g., `opam install lwt ssl`, we should still try to warn them that that's not how you get `Lwt_ssl` anymore. We can remove these messages after maybe a few more releases of Lwt.

I kept the possibility to build the packages in the main build system mainly because I don't feel like reworking it. In particular, I don't want to mess up the modules shown in the ocsigen.org manual.

In the course of working on `lwt_glib`, I noticed that it doesn't actually need `lablgtk`, so I removed that dependency. This may break a user's build system that expects `lablgtk` to appear transitively.

The extra packages have been bumped to version 1.0.1. They all require Lwt 3.0.0, mainly because installing them with Lwt < 3.0.0 will result in duplicate files in Findlib packages `lwt` and whatever extra package is installed. This will result in lots of Findlib warnings, and errors about inconsistent assumptions.


aantron:
I amended the commit. Travis and AppVeyor now generate the extra packages' release build systems and install the packages.

PR #339, opened 2017-04-13
Tests: new test suite for lwt.ml

aantron:
This PR gives `lwt.ml` a new test suite.

Pretty much every API is tested. The coverage of `lwt.ml` is something like 88%, and the code not covered falls into three categories:

- Unreachable code (`assert false` cases and the like), most of which is eliminated by the use of GADTs in the pending `lwt.ml` refactoring. This is most of the code not covered.
- Code that calls `exit` or is part of `atexit`, because I didn't want to kill the testing process.
- Some of the implementation details of "waiter" handling, which I might exercise with a couple extra tests later.

The thoroughness of the test suite extends well beyond coverage. Besides testing various promise state transition sequences, the test suite checks the interaction of cancelation and of storage with all relevant APIs. Of course, there are some oversights, as it's hard to be fully thorough here, but we can fill in the gaps as we find them.

The number of test cases is increased from 62 to 342. The new tests tend to be more fine-grained and focused, however, so this comparison is not completely fair. One complex existing test, `on_cancel race condition`, has no direct counterpart in the new test suite. Otherwise, as far as I can tell, everything tested by the old test suite is tested by the new one. If we decide that we need exactly `on_cancel race condition` later, we can restore it from history.

All tests now have names (instead of numbers). I tried to make the meaning of each test self-evident, based on only the test name and the code, but some tests have natural-language comments attached. Perhaps, more comments will have to be added.

The test suite is broken up into logical sections, and it's probably easiest to read it with code folding. Otherwise, you just see a massive amount of repetitive OCaml text.

This test suite automatically checks a considerable proportion of what I've learned about the semantics of `lwt.ml`. That should make the Lwt core easier to edit, especially (and hopefully) after the upcoming refactoring PR.

Ultimately, this was triggered by trying to write the new manual, for the many users, especially newcomers, confused by the semantics of Lwt. I found that I don't know the detailed semantics myself, and maybe only a few people do. So, I started reading `lwt.ml` *very* thoroughly. That led to some refactoring, which led to this test suite.

### DIFF
--- a/tests/core/test_lwt.ml
+++ b/tests/core/test_lwt.ml
@@ -2384,6 +2384,8 @@ let cancel_bind_tests = [
     Lwt.on_cancel p1 (fun () -> f_ran := true);
     Lwt.on_cancel p2 (fun () -> g_ran := true);
     Lwt.cancel p2;
+    (* Canceling [p2] doesn't cancel [p1], because the function passed to
+       [Lwt.bind] never ran. *)
     Lwt.return (not !f_ran && !g_ran)
   end;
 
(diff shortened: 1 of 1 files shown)


c-cube:
That is very impressive. Congratulations.

PR #348, opened 2017-05-08
tail recursive split function for Lwt_log

domsj:
This one actually bit us in https://github.com/openvstorage/alba/ (resulted in segfaults & stackoverflows)


aantron:
Much appreciated!

Issue #349, opened 2017-05-10
Assert failure in Lwt.fast_connect


domsj:
Got the following error while running alba stress2 test suite:
```
alba: internal error, uncaught exception:
      "Assert_failure src/core/lwt.ml:511:8"
      Raised at file "src/core/lwt.ml", line 511, characters 8-20
      Called from file "camlinternalLazy.ml", line 27, characters 17-27
      Re-raised at file "camlinternalLazy.ml", line 34, characters 10-11
      Called from file "src/unix/lwt_unix.cppo.ml", line 133, characters 54-78
      Called from file "src/unix/lwt_engine.ml", line 188, characters 6-24
      Called from file "src/unix/lwt_engine.ml", line 188, characters 6-24
      Re-raised at file "src/unix/lwt_engine.ml", line 191, characters 12-15
      Called from file "src/unix/lwt_main.ml", line 41, characters 8-82
      Called from file "src/cmdliner_term.ml", line 27, characters 19-24
      Called from file "src/cmdliner.ml", line 106, characters 32-39
```
It looks like some internal lwt error. We use lwt 3.0.0 with the libev engine.

aantron:
Hmm, looks like my attempt to reply by email didn't work.

> Ouch. This very likely means a mistake in some internal assumptions made by Lwt, and a serious bug – though it could also be some kind of race condition, if you might be accidentally touching the same Lwt promise from multiple threads. I'll take a look later today or tomorrow. 
>
> Thanks for running that stress test and for reporting this.

The stack trace unfortunately isn't very helpful, I guess several functions that must have called `fast_connect` are missing due to inlining (?).

aantron:
...and I don't see the potential mistake in Lwt. Have you been able to narrow it down to a small-ish example?

domsj:
Hi @aantron,
Thanks for looking into this quickly.

I've tried reproducing it, but it seems like I only had this specific error once.
The stress suite did fail with other random segfaults once in a while ... so something else is most likely wrong in alba.

After my last change (https://github.com/openvstorage/alba/pull/717 - use tail recursive List.map & friends everywhere) it seems so far that the stress suite now runs without problems, although it's too early to tell for sure. (If no errors come up in the next few hours I think the issue has been fixed...)

Anyway I'll close this ticket as it was most likely a false alarm.

aantron:
[...]

Issue #340, opened 2017-04-13
Lwt.npick is broken


aantron:
It seems that trying to resolve `Lwt.npick`, when the argument promises are initially pending, results in a promise that fails with `Lwt.Canceled`:

```ocaml
open Lwt.Infix

let () =
  let p1, r = Lwt.task () in
  let p2, _ = Lwt.task () in

  let p3 = Lwt.npick [p1; p2] in
  assert (Lwt.state p3 = Lwt.Sleep);

  Lwt.wakeup r ();

  (* Expected: *)
  (* assert (Lwt.state p3 = Lwt.Return [()]) *)

  (* Actual: *)
  assert (Lwt.state p3 = Lwt.Fail Lwt.Canceled);

(* ocamlfind opt -linkpkg -package lwt npick.ml && ./a.out *)
```

By comparison, if you replace `npick` with `nchoose`, you get the Expected behavior.

Since nobody has complained about this, I guess nobody is actually using `npick` with pending promises :)

aantron:
Probably best fixed after #354 is in.

Issue #315, opened 2017-02-08
Segmentation Fault within `Lwt.run_waiters_rec`?


smondet:
Has anybody seen this one before?

```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/home/opam/.opam/4.03.0/bin/coclobas-ketrew start-server -C /tmp/cocloketrew/co'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00000000007b5321 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:196
196     src/core/lwt.ml: No such file or directory.
[Current thread is 1 (Thread 0x7f1d11448740 (LWP 3780))]
(gdb) bt
#0  0x00000000007b5321 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:196
#1  0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#2  0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#3  0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#4  0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#5  0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#6  0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#7  0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#8  0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#9  0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#10 0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#11 0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#12 0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
#13 0x00000000007b5357 in camlLwt__run_waiters_rec_1372 () at src/core/lwt.ml:201
[...]

Issue #304, opened 2016-12-10
Write new manual


aantron:
It's time.

aantron:
I'm closing this in favor of the [New docs](https://github.com/ocsigen/lwt/projects/2) project, because it will take more than one issue – looks like it might take about 16. GitHub is a bit unfriendly by not showing the project description on the project page, but you can find it [here](https://github.com/ocsigen/lwt/projects).

PR #354, opened 2017-05-14
lwt.ml: human-friendly edition – major refactoring and commenting of the Lwt core

aantron:
This PR *heavily* updates `lwt.ml` to make it much more legible and friendly.

What the [new `lwt.ml`][lwt-full] looks like...

- [in the ordinary view][lwt-full], where you can see it in all its glory, read the friendly Reading Guide and Overview, and browse all the code,
- [with internal modules folded (94 lines)][lwt-modules], where you can neatly see the high-level structure of the Lwt core, and
- [with internal signatures expanded (282 lines)][lwt-signatures], where you can see what belongs where.

For comparison, what the previous `lwt.ml` looks like: [[link]][lwt-original].

[lwt-modules]: https://gist.github.com/aantron/9fab0bdead98a60fccf06e0189186863#file-lwt-modules-ml-L20
[lwt-signatures]: https://gist.github.com/aantron/97b58520d5bb4858ccac6f54700a24d7#file-lwt-sigs-ml-L32
[lwt-full]: https://github.com/ocsigen/lwt/blob/fdff5c09d47d2b020d4998ebed922acf383a2e9d/src/core/lwt.ml#L27
[lwt-original]: https://github.com/ocsigen/lwt/blob/8a397d68691a665320fefbdc72061967917fc2b5/src/core/lwt.ml#L29

<br/>

### Why?

This PR is a series of gradual refactorings of `lwt.ml` to make it more legible, such that the final version shares almost no lines in common with the original, though the ASTs are still largely similar. There are two main purposes:

- Make it much easier to contribute to Lwt. I've seen several comments along the lines of "I wanted to contribute, but I couldn't read `lwt.ml`," and *I* couldn't read it well either. Even if someone can't read the new `lwt.ml`, they should feel welcomed and encouraged to ask about it.
- Have enough people actually understand the semantics of `lwt.ml` deeply to be able to do some combination of

    - write the manual (#304),
    - adapt to effects,
    - adapt better to JS, and/or
    - change the semantics in various ways (e.g. #329).

    Ideally, enough people can read `lwt.ml` for the semantics to enter the community's organizational memory. This will hopefully free Lwt from some paralysis and inertia.

    I found that, while reading `lwt.ml`, I needed to refactor it in some places, to be able to continue reading it more and more deeply, instead of getting stuck on superficial difficulties. That eventually led to this whole PR.

<br/>

### Highlights of the new code

The new `lwt.ml` has:

- **Thorough documentation**, including the mentioned Reading Guide and Overview, as well as many implementation detail comments. The Overview discusses all concepts necessary to understand everything in `lwt.ml`.
- **Sectioning** with internal modules. Signatures limit the scope of helpers, and provide type-checked tables of contents. Internal modules can be folded by text editors that support code folding. Functions are reordered by category, compared to the original `lwt.ml`.
- Lots of renaming for clearer, **self-descriptive terminology**, and a switch to the "promise" language.
- General refactoring and cleanup to organize the code.
- GADTs to eliminate a large number of `assert false` cases. The original code had 35 `assert false` expressions, which correspond to a considerably larger number of cases, because many appear under or-patterns. The new code has four `assert false` expressions.
- Explicit annotation and documentation of promise state changes. Encoding (mutable) promise state in a GADT requires casts when state changes, and these casts explicitly document the state changes Lwt believes are possible between when Lwt gives a promise to a user, and when Lwt uses that promises again later.
- No first-class modules. These were used for encoding existential types, and have been replaced by GADTs.

It's undoubtedly possible to improve the code even further, but I'd like to commit this as a starting point for that process.

<br/>

### How to review

This PR is broken up into 57 commits, of which:

- The first 40 are trivial renamings, reorderings, and whitespace changes.
- The next 11, from "be explicit about public/internal casts" to "refactor nchoose, npick, nchoose_split" are non-trivial changes that require true thinking. Of these, "make promise state into a GADT" is the most involved. Many of these have detailed commit messages.
- The remaining 6 are again trivial renamings, reorderings, and whitespace changes.

It's still a lot to deal with, but hopefully this way of breaking up the history makes it somewhat possible to review the PR. It should also give `lwt.ml` fairly sane blame.

Each commit passes the Lwt [test suite](https://github.com/ocsigen/lwt/pull/339). Code coverage is 98%, though the test suite is more thorough than just achieving high coverage.

<br/>

### Pings and notes

I have to thank @dkim for some suggestions on improving the Overview. Thanks!

@talex5 Obviously, this affects tracing. Ultimately, I hope it makes tracing easier to maintain. We should probably also work on getting it maintained in the main Lwt repo at some point (#180).

This code should be bug-for-bug compatible with the previous code. That includes the assertion failures and segfaults we recently saw in #315 and #349. In fact, this code is likely to have worse behavior in such situations, due to having fewer assertion expressions. Preserved bugs also include #340.

Some of the stuff in the overview belongs in `lwt.mli`, and may get moved there when working on the manual.

I would appreciate it if anyone with access to a stress test, or any other non-trivial Lwt code, would give this rewrite a try (@mfp, @copy, @domsj?). The command to pin it should be:

```
opam pin add lwt https://github.com/ocsigen/lwt.git\#wat
```

<br/>

Thanks!

### DIFF
--- a/_tags
+++ b/_tags
@@ -9,6 +9,8 @@ not <src/ssl/*>: safe_string
 # to depend on how Ocamlbuild internally handles lists of warn() tags.
 <src/camlp4/*.ml> or <src/ppx/*.ml>: warn(-4)
 <src/camlp4/*.ml>: warn(-3)
+# Warning 4 is disabled due to https://caml.inria.fr/mantis/view.php?id=7451.
+<src/core/lwt.ml>: warn(-4)
 <**/*>: warn(+A-29-58)
 
 # Syntax extension
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -134,8 +134,11 @@
 
      ...are those that may become completed in the future. Each pending promise
      carries a list of callbacks. These callbacks are added by functions like
-     [Lwt.bind], and called by Lwt if/when the promise completes. Pending
-     promises are produced in three ways, according to how they can be
+     [Lwt.bind], and called by Lwt if/when the promise completes. These
+     callbacks typically end up completing additional promises; see section
+     "Completion loop" below.
+
+     Pending promises are produced in three ways, according to how they can be
      completed:
 
      - Initial promises
@@ -146,15 +149,16 @@
 
      - Sequential composition
 
-       For example, [Lwt.bind]. These promises only complete when some sequence
-       of "preceding" promises completes. The user cannot complete these
-       promises directly.
+       For example, [Lwt.bind]. These promises only complete when the preceding
+       sequence of promises completes. The user cannot complete these promises
+       directly (but see the section on cancelation below).
 
      - Concurrent composition
 
        For example, [Lwt.join] or [Lwt.choose]. These promises only complete
        when all or one of a set of "preceding" promises complete. The user
-       cannot complete these promises directly.
+       cannot complete these promises directly (but see the section on
+       cancelation below).
 
 
    2. Resolvers
@@ -165,7 +169,7 @@
 
    Internally, resolvers are the exact same objects as the promises they
    complete, even though the resolver is exposed as a reference of a different
-   type by [lwt.mli].
+   type by [lwt.mli]. For details on why, see section "Type system abuse" below.
 
 
    3. Callbacks
@@ -180,7 +184,8 @@
    Callbacks come in two flavors: regular callbacks and cancel callbacks. The
    only material differences between them are that:
 
-   - cancel callbacks are only called if a promise is canceled, and
+   - regular callbacks are always called when a promise is completed, but cancel
+     callbacks are called, in addition, only if the promise is canceled, and
    - all cancel callbacks of a promise are called before any regular callback
      is called.
 
@@ -198,8 +203,9 @@
    This chaining of promise completions through callbacks can be seen as a kind
    of promise dependency graph, in which the nodes are pending promises, and the
    edges are callbacks. During the completion loop, Lwt starts at some initial
-   promise that is getting completed by the user, and completes a bunch of
-   dependent promises.
+   promise that is getting completed by the user, and recursively completes all
+   dependent promises. The graph is modified: completed promises are removed
+   from it.
 
    Some of these dependencies are explicit to Lwt, e.g. the callbacks registered
    by [Lwt.bind]. Others are not visible to Lwt, because the user can always
@@ -254,11 +260,14 @@
 
    The Lwt core deliberately doesn't do I/O. The completion loop stops running
    once no promises can be completed immediately. It has to be restarted later
-   by some external process.
+   by some surrouding I/O loop. This I/O loop typically keeps track of pending
+   promises that represent blocked or in-progress I/O; other pending promises
+   that indirectly depend on I/O are not explicitly tracked. They are retained
+   in memory by references captured inside callbacks.
 
    On Unix and Windows, a separate top-level loop, typically [Lwt_main.run], is
    necessary to repeatedly call [select], [epoll], or [kevent], and complete
-   promises that are blocked on I/O.
+   blocked I/O promises.
 
    In JavaScript, references to promises are retained by JavaScript code, which
    is, in turn, triggered by the JS engine. In other words, the top-level loop
@@ -1718,7 +1727,15 @@ struct
 
     | Pending p_callbacks ->
       let p'' = new_pending ~how_to_cancel:(Propagate_cancel_to_one p) in
-      (* The result promise is a fresh pending promise. *)
+      (* The result promise is a fresh pending promise.
+
+         Initially, trying to cancel this fresh pending promise [p''] will
+         propagate the cancelation attempt to [p] (backwards through the promise
+         dependency graph). If/when [p] is resolved, Lwt will call the user's
+         callback [f] below, which will provide a new promise [p'], and [p']
+         will become a proxy of [p'']. At that point, trying to cancel [p'']
+         will be equivalent to trying to cancel [p'], so the behavior will
+         depend on how the user obtained [p']. *)
 
       let saved_storage = !current_storage in
 
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -113,7 +113,7 @@ type +'a result = ('a, exn) Result.result
 
   let state_of_result = function
     | Result.Ok x -> Resolved x
-    | Result.Error e -> Failed e
+    | Result.Error exn -> Failed exn
 
 let make_value v = Result.Ok v
 let make_error exn = Result.Error exn
@@ -126,11 +126,11 @@ include Public_types
 module Basic_helpers =
 struct
 
-let rec underlying t =
-  match t.state with
-    | Unified_with t' ->
-      let t'' = underlying t' in if t'' != t' then t.state <- Unified_with t''; t''
-    | Resolved _ | Failed _ | Pending _ -> t
+let rec underlying p =
+  match p.state with
+    | Unified_with p' ->
+      let p'' = underlying p' in if p'' != p' then p.state <- Unified_with p''; p''
+    | Resolved _ | Failed _ | Pending _ -> p
 
 let repr t = underlying (to_internal_promise t)
 
@@ -141,9 +141,9 @@ open Basic_helpers
 
 module Sequence_associated_storage =
 struct
-  type 'a key = {
+  type 'v key = {
     id : int;
-    mutable store : 'a option;
+    mutable value : 'v option;
   }
 
   let next_key_id = ref 0
@@ -151,15 +151,15 @@ struct
   let new_key () =
     let id = !next_key_id in
     next_key_id := id + 1;
-    { id = id; store = None }
+    { id = id; value = None }
 
   let current_storage = ref Storage_map.empty
 
   let get key =
     try
       Storage_map.find key.id !current_storage ();
-      let value = key.store in
-      key.store <- None;
+      let value = key.value in
+      key.value <- None;
       value
     with Not_found ->
       None
@@ -169,7 +169,7 @@ struct
     let data =
       match value with
         | Some _ ->
-            Storage_map.add key.id (fun () -> key.store <- value) save
+            Storage_map.add key.id (fun () -> key.value <- value) save
         | None ->
             Storage_map.remove key.id save
     in
@@ -214,34 +214,34 @@ struct
 
   let cleanup_throttle = 42
 
-  let remove_waiters l =
+  let remove_waiters ps =
     List.iter
-      (fun t ->
-         match (repr t).state with
+      (fun p ->
+         match (repr p).state with
            | Pending ({ regular_callbacks =
-                   Regular_callback_list_explicitly_removable_callback _; _ } as sleeper) ->
-               sleeper.regular_callbacks <- Regular_callback_list_empty
+                   Regular_callback_list_explicitly_removable_callback _; _ } as callbacks) ->
+               callbacks.regular_callbacks <- Regular_callback_list_empty
            | Pending ({regular_callbacks =
                    Regular_callback_list_empty
                  | Regular_callback_list_implicitly_removed_callback _
-                 | Regular_callback_list_concat _; _} as sleeper) ->
-               let removed = sleeper.cleanups_deferred + 1 in
+                 | Regular_callback_list_concat _; _} as callbacks) ->
+               let removed = callbacks.cleanups_deferred + 1 in
                if removed > cleanup_throttle then begin
-                 sleeper.cleanups_deferred <- 0;
-                 sleeper.regular_callbacks <- clean_up_callback_cells sleeper.regular_callbacks
+                 callbacks.cleanups_deferred <- 0;
+                 callbacks.regular_callbacks <- clean_up_callback_cells callbacks.regular_callbacks
                end else
-                 sleeper.cleanups_deferred <- removed
+                 callbacks.cleanups_deferred <- removed
            | Resolved _ | Failed _ | Unified_with _ ->
                ())
-      l
+      ps
 
   let add_regular_callback_list_node callbacks node =
     callbacks.regular_callbacks <- (match callbacks.regular_callbacks with
                           | Regular_callback_list_empty -> node
                           | Regular_callback_list_implicitly_removed_callback _
                           | Regular_callback_list_explicitly_removable_callback _
-                          | Regular_callback_list_concat _ as ws ->
-                            Regular_callback_list_concat (node, ws))
+                          | Regular_callback_list_concat _ as existing ->
+                            Regular_callback_list_concat (node, existing))
 
   let add_implicitly_removed_callback callbacks f =
     add_regular_callback_list_node callbacks (Regular_callback_list_implicitly_removed_callback f)
@@ -249,10 +249,10 @@ struct
   let add_explicitly_removable_callback_to_each_of ps f =
     let node = Regular_callback_list_explicitly_removable_callback f in
     List.iter
-      (fun t ->
-         match (repr t).state with
-           | Pending sleeper ->
-               add_regular_callback_list_node sleeper node
+      (fun p ->
+         match (repr p).state with
+           | Pending callbacks ->
+               add_regular_callback_list_node callbacks node
            | Resolved _ | Failed _ | Unified_with _ ->
                assert false)
       ps
@@ -280,48 +280,48 @@ struct
 
   exception Canceled
 
-let rec run_waiters_rec state ws rem =
-  match ws with
+let rec run_waiters_rec result fs rest =
+  match fs with
     | Regular_callback_list_empty ->
-        run_waiters_rec_next state rem
+        run_waiters_rec_next result rest
     | Regular_callback_list_implicitly_removed_callback f ->
-        f state;
-        run_waiters_rec_next state rem
+        f result;
+        run_waiters_rec_next result rest
     | Regular_callback_list_explicitly_removable_callback { contents = None } ->
-        run_waiters_rec_next state rem
+        run_waiters_rec_next result rest
     | Regular_callback_list_explicitly_removable_callback { contents = Some f } ->
-        f state;
-        run_waiters_rec_next state rem
-    | Regular_callback_list_concat (ws1, ws2) ->
-        run_waiters_rec state ws1 (ws2 :: rem)
+        f result;
+        run_waiters_rec_next result rest
+    | Regular_callback_list_concat (fs1, fs2) ->
+        run_waiters_rec result fs1 (fs2 :: rest)
 
-and run_waiters_rec_next state rem =
-  match rem with
+and run_waiters_rec_next result rest =
+  match rest with
     | [] ->
         ()
-    | ws :: rem ->
-        run_waiters_rec state ws rem
+    | fs :: rest ->
+        run_waiters_rec result fs rest
 
-let rec run_cancel_handlers_rec chs rem =
-  match chs with
+let rec run_cancel_handlers_rec fs rest =
+  match fs with
     | Cancel_callback_list_empty ->
-        run_cancel_handlers_rec_next rem
-    | Cancel_callback_list_callback (data, f) ->
-        current_storage := data;
+        run_cancel_handlers_rec_next rest
+    | Cancel_callback_list_callback (storage, f) ->
+        current_storage := storage;
         handle_with_async_exception_hook f ();
-        run_cancel_handlers_rec_next rem
+        run_cancel_handlers_rec_next rest
     | Cancel_callback_list_remove_sequence_node n ->
         Lwt_sequence.remove n;
-        run_cancel_handlers_rec_next rem
-    | Cancel_callback_list_concat (chs1, chs2) ->
-        run_cancel_handlers_rec chs1 (chs2 :: rem)
+        run_cancel_handlers_rec_next rest
+    | Cancel_callback_list_concat (fs1, fs2) ->
+        run_cancel_handlers_rec fs1 (fs2 :: rest)
 
-and run_cancel_handlers_rec_next rem =
-  match rem with
+and run_cancel_handlers_rec_next rest =
+  match rest with
     | [] ->
         ()
-    | chs :: rem ->
-        run_cancel_handlers_rec chs rem
+    | fs :: rest ->
+        run_cancel_handlers_rec fs rest
 
   let run_callbacks callbacks result =
     (match result with
@@ -334,9 +334,9 @@ and run_cancel_handlers_rec_next rem =
   let complete p result =
     let t = repr p in
     match t.state with
-      | Pending sleeper ->
+      | Pending callbacks ->
           t.state <- result;
-          run_callbacks sleeper result
+          run_callbacks callbacks result
       | Resolved _ | Failed _ | Unified_with _ ->
           assert false
 
@@ -351,7 +351,7 @@ end
 let queued_callbacks = Queue.create ()
 
   let enter_completion_loop () =
-    let snapshot = !current_storage in
+    let storage_snapshot = !current_storage in
     let already_wakening =
       if !currently_in_completion_loop then
         true
@@ -360,9 +360,9 @@ let queued_callbacks = Queue.create ()
         false
       end
     in
-    (already_wakening, snapshot)
+    (already_wakening, storage_snapshot)
 
-  let leave_completion_loop (already_wakening, snapshot) =
+  let leave_completion_loop (already_wakening, storage_snapshot) =
     if not already_wakening then begin
       while not (Queue.is_empty queued_callbacks) do
         let closure = Queue.pop queued_callbacks in
@@ -370,26 +370,26 @@ let queued_callbacks = Queue.create ()
         run_callbacks M.callbacks M.state
       done;
       currently_in_completion_loop := false;
-      current_storage := snapshot
+      current_storage := storage_snapshot
     end else
-      current_storage := snapshot
+      current_storage := storage_snapshot
 
   (* See https://github.com/ocsigen/lwt/issues/48. *)
   let abandon_wakeups () =
     if !currently_in_completion_loop then leave_completion_loop (false, Storage_map.empty)
 
-  let run_in_completion_loop sleeper state =
+  let run_in_completion_loop callbacks result =
     let ctx = enter_completion_loop () in
-    run_callbacks sleeper state;
+    run_callbacks callbacks result;
     leave_completion_loop ctx
 
   let wakeup_result r result =
     let t = underlying (to_internal_resolver r) in
     match t.state with
-      | Pending sleeper ->
+      | Pending callbacks ->
           let state = state_of_result result in
           t.state <- state;
-          run_in_completion_loop sleeper state
+          run_in_completion_loop callbacks state
       | Failed Canceled ->
           ()
       | Resolved _ | Failed _ | Unified_with _ ->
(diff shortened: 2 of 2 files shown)


src/core/lwt.ml
```diff
@@ -51,7 +378,45 @@ struct
...
+     potential state change for readers of the code. *)
+
+  (* Note:
+
+     The phantom types used in the GADT are declared as abstract types. They
+     must be declared outside the module, for OCaml's type system to know that
+     they are definitely distinct types. If they are declared inside this
+     module, OCaml sees them through an inferred signature in the rest of
+     [lwt.ml]. That signature does not guarantee that they are distinct, which
+     defeats the purpose of the GADT in pattern matching. *)
```
Drup: While this note is partially correct, you can get rid of the problem by giving a constructor for your phantom types (and expose the constructor):
```ocaml
type underlying = Underlying
```

This is a better way to prove to the typechecker that underlying is different to every other types.

Drup:
I'm not fond of the "Unification" vocabulary. I think something in term of "Aliasing" would be better. In any cases, the basic idea is that it's an edge in the DAG of dependencies between tasks, which is probably easier to explain.

aantron:
Agreed, I struggled with that. I encourage reviewers to challenge/improve the vocabulary. We can probably come up with something better than I can come up with alone :)

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+
+       For example, [Lwt.bind]. These promises only complete when some sequence
+       of "preceding" promises completes. The user cannot complete these
+       promises directly.
+
+     - Concurrent composition
+
+       For example, [Lwt.join] or [Lwt.choose]. These promises only complete
+       when all or one of a set of "preceding" promises complete. The user
+       cannot complete these promises directly.
```
edwintorok: How about cancelation?

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+
+   2. Resolvers
+
+   Resolvers are given to the user by [Lwt.wait] and [Lwt.task], and can be used
+   by the user to resolve the corresponding promises. Note that this means the
+   user only ever gets resolvers for initial promises.
+
+   Internally, resolvers are the exact same objects as the promises they
+   resolve, even though the resolver is exposed as a reference of a different
+   type by [lwt.mli].
```
edwintorok: Maybe add a reference to 'Type system abuse' which explains why.

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+
+
+   3. Callbacks
+
+   ...are attached by Lwt to pending promises, and are run by Lwt if/when those
+   promises complete. These callbacks are not directly exposed through
+   [lwt.mli] -- they are a low-level mechanism. For example, to implement
+   [Lwt.bind p f], Lwt attaches a callback to [p] that does some internal Lwt
+   book-keeping, and then calls [f] if [p] resolved, and does something else if
```
edwintorok: The usage of 'resolved' here is correct, but see above for confusion with resolvers.

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+   [p] failed.
+
+   Callbacks come in two flavors: regular callbacks and cancel callbacks. The
+   only material differences between them are that:
+
+   - cancel callbacks are only called if a promise is canceled, and
+   - all cancel callbacks of a promise are called before any regular callback
+     is called.
```
edwintorok: The regular callbacks are always called, right?

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+   Completing a pending promise triggers its callbacks, and those might complete
+   more pending promises, triggering more callbacks, etc. This behavior is the
+   *completion loop*. Lwt has some machinery to avoid stack overflow and other
+   unfortunate situations during this loop.
+
+   This chaining of promise completions through callbacks can be seen as a kind
+   of promise dependency graph, in which the nodes are pending promises, and the
+   edges are callbacks. During the completion loop, Lwt starts at some initial
+   promise that is getting completed by the user, and completes a bunch of
+   dependent promises.
```
edwintorok: I assume that the dependency graph changes during this process, and the edges are removed once the callbacks are (about to be) executed by `add_implicitly_removed_callback`? I think it would be useful to mention this, otherwise you get worried on what happens if you keep binding on `Lwt.return ()` ...

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+   - The user can create an initial promise, then fail it through its resolver.
+     The search is redundant because it would find only the same initial promise
+     to cancel.
+   - The user can create a trivial promise by calling [Lwt.fail Lwt.Canceled].
+     The search is again redundant; in this case it would find nothing to
+     cancel.
+
+   Note that there is a quirk: only promises created by [Lwt.task] are
+   susceptible to being canceled by [Lwt.cancel], but the user can manually
+   cancel initial promises created by both [Lwt.task] and [Lwt.wait].
```
edwintorok: What is the semantics of `Lwt.bind` wrt. cancelation?

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+   initiated by the user without access to a resolver. This is important for
+   reasoning about state changes in the implementation of Lwt, and is referenced
+   in some implementation detail comments.
+
+
+   6. No I/O
+
+   The Lwt core deliberately doesn't do I/O. The completion loop stops running
+   once no promises can be completed immediately. It has to be restarted later
+   by some external process.
```
edwintorok: Does this mean that all 'pending' promises are added to some global queue where this external loop can find it?

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+   terms of the internal type.
+
+   Casting from an ['a t] to an internal promise produces a reference for
+   which the state is "unknown": this is simulated with a helper GADT, which
+   encodes existential types. There are several similar casts, which are used
+   to document possible state changes between the time a promise is created,
+   and the later time it is used in a callback. You can see these casts in
+   action in [Lwt.bind]. The cast syntax is pretty light, and, besides being
+   commented in [bind], all such casts are documented in modules [Public_types]
+   and [Basic_helpers].
```
edwintorok: How can we check that those casts do not break assumptions made by the compiler/flambda?

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+
+   * Documentation
+
+   The documentation begins with an Overview of major concepts and components.
+   This overview puts everything into context. You don't have to read the whole
+   thing. The overview begins with basic concepts, moves on to advanced ones,
+   and then gets into the truly esoteric. You can read about each concept on an
+   as-needed basis. However, once you have read the whole overview, you will be
+   aware of *everything* that is needed to understand, and work with, the core
+   of Lwt.
```
edwintorok: IIUC the change in terminology (e.g. 'resolver' instead of 'wakener') comes from moving to 'promises' instead of 'threads'. Might be useful to add some table at the end on what the equivalent of the old `lwt.mli` terminology is in the new one for those who got used to the old terminology already?

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+
+       For example, [Lwt.join] or [Lwt.choose]. These promises only complete
+       when all or one of a set of "preceding" promises complete. The user
+       cannot complete these promises directly.
+
+
+   2. Resolvers
+
+   Resolvers are given to the user by [Lwt.wait] and [Lwt.task], and can be used
+   by the user to resolve the corresponding promises. Note that this means the
```
talex5: FWIW, here's the E terminology:

http://www.erights.org/elib/concurrency/refmech.html

- There, a "resolved" value can be either "fulfilled" or "broken".

(the E model is more complex than the Lwt one because it also handles remote promises)

raphael-proust:
Re: “unification”, this makes some of the documentation confusing. Specifically when “unification” is used to talked about the types. E.g., the documentation of `identical`

>     (* Checks physical equality ([==]) of two internal promises. Unlike [==], does
>     not force unification of their invariants. *)

I like “a proxy”, “to proxying” as an alternative. Or “a stand-in”, “to stand-in”. Basically, it should not share vocabulary with other common concepts in ML. So “alias” should be avoided: “type aliasing”, “alias pattern”.

aantron:
Thanks @edwintorok, @Drup, @copy, @talex5, @raphael-proust, and everyone else who gave, or will give, this PR a look. I'm going to leave it in this state for a little while, before adding commits to address the comments. This is both for me to take a break, and for reviewers to have a chance to look at the PR without it changing under them.

domsj:
fyi ran all the alba test suites against this branch and had no errors, except that on some jenkins slaves building the docker container stumbled on this:

```
#=== ERROR while installing lwt.dev ===========================================#
# opam-version         1.2.2
# os                   linux
# command              make build
# path                 /home/jenkins/OPAM/4.03.0/build/lwt.dev
# compiler             4.03.0
# exit-code            2
# env-file             /home/jenkins/OPAM/4.03.0/build/lwt.dev/lwt-245-3bd1cf.env
# stdout-file          /home/jenkins/OPAM/4.03.0/build/lwt.dev/lwt-245-3bd1cf.out
# stderr-file          /home/jenkins/OPAM/4.03.0/build/lwt.dev/lwt-245-3bd1cf.err
### stdout ###
# grep -v '^#' setup.ml > setup_dev.ml
# ocamlfind ocamlopt -o setup-dev.exe -linkpkg -package ocamlbuild,oasis.dynrun setup_dev.ml || \
#   ocamlfind ocamlc -o setup-dev.exe -linkpkg -package ocamlbuild,oasis.dynrun setup_dev.ml || true
# rm -f setup_dev.*
# ./setup-dev.exe -configure 
# Makefile:71: recipe for target 'setup.data' failed
### stderr ###
# File "setup_dev.ml", line 1, characters 0-2:
# Error: Comment not terminated
# File "setup_dev.ml", line 1, characters 0-2:
# Error: Comment not terminated
# make: ./setup-dev.exe: Command not found
# make: *** [setup.data] Error 127
```

https://github.com/openvstorage/alba/compare/test-new-lwt
https://github.com/openvstorage/alba/blob/17f68c5a359208be116163388ead73c8545e7e1f/docker/ubuntu-16.04/Dockerfile

Do note that the travis build - which always starts from scratch - worked just fine. So probably some caching of another docker layer was screwing this up?
(It's not bothering me - I can work around it by cleaning up some docker images - so feel free to ignore this error.)

aantron:
Thanks.

That's interesting. I'm concerned about the duplicate lines

```
```

raphael-proust:
The section about cancelation alone makes this PR worth it. I was always bugged by the explanation that it “cancels the deepest cancelable thread” (as if there was only one; with no explanation of ‘deepest’).

src/core/lwt.ml
raphael-proust: I think this is covered in `lwt.mli`. I don't think it needs explaining in the overview of the code here. Maybe, the bind function could have additional comments about that.

src/core/lwt.ml
raphael-proust: IIUC, this is for the scheduler/IO-loop to manage. Typically, when you do something like `Lwt_unix.read ()` it
- creates a promise (`p`),
- then it prepares a callback (`c`) that, when triggered, completes the promise,
- then it calls `select`, and it sets `c` to be called when `select` is triggered when uses `select` returns.

Note: it can be select or it can be some other similar mechanism.  
Note: the upcoming algebraic effects might make it easier to integrate Lwt to an IO-loop.  
Note: the callback `c` is not a callback set by `Lwt`, it is set by `Lwt_unix` (in this example, other libs can do something similar)

This is from what I remember from discussions about Lwt several years ago. It might be inaccurate, in which case, I'd be interested in what actually happens.

I think that there should be a reference like “check the documentation of the file Lwt_unix.ml for an explanation of one IO-loop” and then a detailed explanation in that file.

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+   7. Promise "unification"
+
+   In [Lwt.bind : 'a t -> ('a -> 'b t) -> 'b t], the outer ['b t] is created by
+   [bind] first, and returned to the user. The inner ['b t] is created by the
+   user later, and then returned to [bind]. At that point, [bind] needs to make
+   the inner and outer ['b t]s behave identically.
+
+   This is accomplished by "unifying" them, which simply means: making one of
+   the promises point to the other. One of the promises thus becomes a "proxy,"
+   and the other is its "underlying" promise.
```
raphael-proust: As stated in a comment in the PR thread, “unification” is confusing because it's also applies to types in ML. I prefer “proxy” and “proxying”.

domsj:
```
That's interesting. I'm concerned about the duplicate lines

# File "setup_dev.ml", line 1, characters 0-2:
# Error: Comment not terminated

If it happens again, could you install with --keep-build-dir, and show setup_dev.ml, if it's not too challenging? It's okay if you don't, maybe we can reproduce this in some other way later.
```

I was able to reproduce the problem in said container.
Summary: grep 2.24 doesn't like the unicode chars in Jérémie's name, updating to grep 2.25 makes the problem go away.

```
jenkins@cf9ecf7f974f:~/OPAM/4.03.0/build/lwt.dev$ cat setup.ml 
(*
 * setup.ml
 * --------
 * Copyright : (c) 2011, Jérémie Dimino <jeremie@dimino.org>
 * Licence   : BSD3
 *
 * This file is a part of lwt.
 *)

(* OASIS_START *)
#use "topfind";;
#require "oasis.dynrun";;
open OASISDynRun;;
(* OASIS_STOP *)

let () =
  InternalInstallPlugin.lib_hook :=
    fun (cs, bs, lib, dn) ->
      match lib.OASISTypes.lib_findlib_name with
        | Some "unix" ->
            (cs, bs, lib, dn, ["src/unix/lwt_config.h"; "src/unix/lwt_unix.h"])
        | _ ->
            (cs, bs, lib, dn, [])
;;

let () = setup ();;
jenkins@cf9ecf7f974f:~/OPAM/4.03.0/build/lwt.dev$ grep -v '^#' setup.ml > setup_dev.ml
jenkins@cf9ecf7f974f:~/OPAM/4.03.0/build/lwt.dev$ cat setup_dev.ml 
(*
 * setup.ml
 * --------
Binary file setup.ml matches
jenkins@cf9ecf7f974f:~/OPAM/4.03.0/build/lwt.dev$ grep --version
grep (GNU grep) 2.24
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and others, see <http://git.sv.gnu.org/cgit/grep.git/tree/AUTHORS>.
```
```
sudo apt update
sudo apt upgrade
```
```
jenkins@cf9ecf7f974f:~/OPAM/4.03.0/build/lwt.dev$ grep --version
grep (GNU grep) 2.25
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and others, see <http://git.sv.gnu.org/cgit/grep.git/tree/AUTHORS>.
jenkins@cf9ecf7f974f:~/OPAM/4.03.0/build/lwt.dev$ grep -v '^#' setup.ml > setup_dev.ml
jenkins@cf9ecf7f974f:~/OPAM/4.03.0/build/lwt.dev$ cat setup_dev.ml 
(*
 * setup.ml
 * --------
 * Copyright : (c) 2011, Jérémie Dimino <jeremie@dimino.org>
 * Licence   : BSD3
 *
 * This file is a part of lwt.
 *)

(* OASIS_START *)
open OASISDynRun;;
(* OASIS_STOP *)

let () =
  InternalInstallPlugin.lib_hook :=
    fun (cs, bs, lib, dn) ->
      match lib.OASISTypes.lib_findlib_name with
        | Some "unix" ->
            (cs, bs, lib, dn, ["src/unix/lwt_config.h"; "src/unix/lwt_unix.h"])
        | _ ->
            (cs, bs, lib, dn, [])
;;

let () = setup ();;
```

c-cube:
I haven't taken the time to read the whole file, but the prelude is truly brilliant. Hat off to @aantron.

smondet:
@aantron 

I did the

    opam pin add lwt https://github.com/ocsigen/lwt.git#wat

and built both [Ketrew](https://github.com/hammerlab/ketrew/) (which uses `Lwt_unix` *and* `Lwt_js`) and [Coclobas](https://github.com/hammerlab/coclobas/).

Ran a few tests, all seems to be working well!

PS: Ketrew needs https://github.com/hammerlab/ketrew/pull/537 to built with Lwt ≥ 3.0.0.

_tags
```diff
@@ -9,6 +9,8 @@ not <src/ssl/*>: safe_string
 # to depend on how Ocamlbuild internally handles lists of warn() tags.
 <src/camlp4/*.ml> or <src/ppx/*.ml>: warn(-4)
 <src/camlp4/*.ml>: warn(-3)
+# Warning 4 is disabled due to https://caml.inria.fr/mantis/view.php?id=7451.
+<src/core/lwt.ml>: warn(-4)
```
avsm: Looks like this is fixed in OCaml 4.05 -- can this warning be re-enabled on that version?

aantron: Yes, but I guess it will require messing with `myocamlbuild.ml`, to pass the flag conditionally. Maybe we can just add a comment here in `_tags`, saying that this is only needed as long as Lwt supports OCaml < 4.05 (which might be a long while).

There is also the possibility that the jbuilder port will preempt having to do it with Ocamlbuild, but I don't know yet if this is easy to do with jbuilder.

aantron:
Thanks again to the reviewers! I promise to reply and adjust the file :)

My current plan for this PR:

- There have been [several](https://github.com/ocsigen/lwt/projects/1) apparent stability issues reported in (the old) `lwt.ml` recently. I've created the [Lwt Stability](https://github.com/ocsigen/lwt/projects/1) project to track them. The GitHub UI is a bit user-unfriendly by not showing the project description; it can be found [here](https://github.com/ocsigen/lwt/projects).
- These stability issues all involve `lwt.ml` apparently failing under considerable load/stress. I want to diagnose (and fix) the failures, with the old `lwt.ml` that was being used.
- Ideally, we will reproduce the failures with a small suite of stress tests inside the Lwt repo (#372).
- We will then subject the new `lwt.ml` from this PR to the stress tests as well.
- The new `lwt.ml` passing should give us considerable additional confidence in merging it.

So, this PR will be merged some time after the stability project is drained of issues.

Depending on how long that looks to be taking, I may do a few rounds of incorporating feedback sooner than that :)

Please let me know if you have any notes, objections, or other thoughts.

cc @copy @smondet @domsj as reporters of issues in the stability project.

aantron:
Ok, the entire [stability project](https://github.com/ocsigen/lwt/projects/1) turned out to be (probably) caused by the [Skylake CPU bug](https://medium.com/ahrefs/skylake-bug-a-detective-story-ab1ad2beddcd). So, nothing is blocking work to finish off this PR. I've tried to address all the comments.

- Switched "unification" vocabulary to "proxying".
- Gave constructors to the phantom types, so they can be moved into an internal module.
- Fixed typos.
- Regarding the warning suppression (ping @avsm), I think we have to keep it. Since there is no `_tags` anymore, I will probably add it as an `[@@@ocaml.warning]` extension point inside `lwt.ml` in the upcoming merge commit.
- Regarding the promise state terminology, I am now thinking that the right thing to do is to fully adopt JS terminology, i.e. completing becomes resolving, resolving becomes fulfilling, etc. I'll open a separate issue about that, and that issue will also include an item to write a table containing the mapping between old terminology, new terminology, and terminology in other popular languages or systems. For this PR, I've gone through the comments and tried to make sure the current transient terminology is self-consistent. A quirk (already observed by @edwintorok) is that what *resolvers* do is allow *completing* promises. After switching to JS terminology, *resolvers* will *resolve* promises. The JS terminology issue doesn't have to result in JS terminology; debate is welcome, but whatever the final terminology is, we will make everything as consistent and obvious as possible.
- Regarding anything that is already explained in the `.mli` file, the `.mli` file will be rewritten as part of creating the new manual. I agree that several things from the comments in the new `lwt.ml` should, or are, explained there. However, the `.mli` file is currently a source of uncertainty, and I'd like to revisit `lwt.ml` once `lwt.mli` is written. I will adjust the comments then to reduce duplication, and everyone is welcome to participate in revisiting `lwt.ml` as well :)

This PR passes all the unit tests, and I believe it has also passed several stress tests by @smondet, @domsj, and @copy. I think we should merge it relatively soon. That will eliminate the chance of further bit rot, and dually eliminate uncertainty about working on `lwt.ml`/unblock several issues in it.

When merging the PR, I will incorporate the work on more accurate exception messages by @raphael-proust that has been added to `master` while the PR was open.

For whatever is left over, and upon new ideas, everyone is welcome to make further issues and PRs to improve `lwt.ml`. As it says inside:

```
* Please edit the code and the docs!

This code is meant to be readable, and to be edited. If you are reading
something, and think there is a much better way to express it, please go
ahead and open a pull request to the Lwt repository at

 https://github.com/ocsigen/lwt
```

I'll probably remove the word "much" from that. It sounds like a barrier.

So... :)

src/core/lwt.ml
aantron: Indeed, I've added a comment to `bind`: https://github.com/ocsigen/lwt/blob/d1abbf586f3eb82a9b0275fc204fdf85d728249a/src/core/lwt.ml#L1732-L1738

src/core/lwt.ml
aantron: I've clarified that not every pending promise is tracked in a global queue (it doesn't have to be global, it's just anything outside the core). Only the promises that represent blocked I/O. The rest of the pending promises are just retained in memory through references inside callbacks registered on those I/O promises, and transitively.

@raphael-proust: Indeed, I mentioned `Lwt_main.run` in the paragraph below this one, so that the reader can visit the `Lwt_unix` instance of this outer I/O loop. However, that I/O loop is not yet well-documented. I plan to change that, and I think we can add a clearer reference once there is something to read there.

src/core/lwt.ml
aantron: I honestly don't know if they do, but I ran code with casts through some performance tests with Flambda, and everything seemed fine. We'll have to:

1. Keep track of Flambda development (I've already started keeping closer track of compiler development, because it would have saved me like a week of time on the Skylake bug, not to mention others involved in debugging that on and off for Lwt).
2. Run future stress tests (#372) with Flambda turned on as well.

Even though there is talk of breaking `Obj.magic` and the like by Flambda, I don't think it can actually do that, as enough code uses it. If it does, I'm guessing there will be flags and/or annotations to disable parts of Flambda, which would also give Lwt time to find better ways of expressing things.

Finally, I should also note that the casting here is worse, but not much worse, than in `lwt.ml` before the PR (which used `%identity` a lot). In the worst case, we can return to the previous level of casting by abandoning the GADT, or maybe doing something smarter with it.

src/core/lwt.ml
```diff
@@ -1,9 +1,9 @@
-(* Lightweight thread library for OCaml
+(* OCaml promise library
```
aantron: I agree, but saving for a future issue. I want to also start a history of Lwt question thread on Discuss, to find out what people know. I heard it said that Lwt started in 1995 (??!?!?) as a module in some bigger project, perhaps even before Ocsigen.

src/core/lwt.ml
```diff
@@ -22,1307 +22,2746 @@
...
+       these promises manually, through the resolvers returned by those
+       functions.
+
+     - Sequential composition
+
+       For example, [Lwt.bind]. These promises only complete when some sequence
+       of "preceding" promises completes. The user cannot complete these
+       promises directly.
+
+     - Concurrent composition
```
aantron: `join` and `choose` are like conjunction and disjunction, yes, but they are both about completing tasks simultaneously. How would `bind` fit into this? It seems to me like `bind`, `join`, and `choose` are part of a wider model, and that's where the distinction between scheduling sequentially and scheduling concurrently comes in. I was under the impression that terminology is fairly standard, except sometimes they are called serial and parallel, and there are pretty standard symbols ⊕ and ⊗ for them.

src/core/lwt.ml
Drup: If you want to play code archeology, I would suggest looking for jérôme's code inside the [unison](https://www.cis.upenn.edu/~bcpierce/unison/) project.

aantron: Thanks for the link.

aantron:
(macOS build failures were caused by some upstream breakage on macOS, that was fixed in `master` by 0e63890511bea3f0b4f145e2332a714fda4894d1 on May 16, so it should be fine on merge, and I will check that before putting the merge into `master`).

src/core/lwt.ml
```diff
@@ -22,1307 +22,2777 @@
...
+     https://gist.github.com/aantron/97b58520d5bb4858ccac6f54700a24d7
+
+   The signatures are unusual: big comments are absent. They are moved into the
+   modules, so that they are hidden by code folding when you (the reader!) are
+   not interested in those modules.
+
+
+   * Documentation
+
+   The documentation begins with an Overview of major concepts and components.
```
avsm: nit: Overview -> overview

aantron: I actually wrote that deliberately, but looking back on it months later, I have to agree :) Changing...

src/core/lwt.ml
```diff
@@ -22,1307 +22,2777 @@
...
+     checker that these types are distinct from each other. Warning 37, "unused
+     constructor," therefore has to be temporarily suppressed. *)
+
+  [@@@ocaml.warning "-37"]
+
+  type underlying = Underlying_and_this_constructor_is_not_used
+  type proxy = Proxy_and_this_constructor_is_not_used
+
+  type completed = Completed_and_this_constructor_is_not_used
+  type pending = Pending_and_this_constructor_is_not_used
```
chambart: You should mark those types as private to enforce it.

src/core/lwt.ml
```diff
@@ -22,1307 +22,2777 @@
...
+  type ('a, 'u, 'c) promise = {
+    mutable state : ('a, 'u, 'c) state;
+  }
+
+  and (_, _, _) state =
+    | Resolved : 'a                  -> ('a, underlying, completed) state
+    | Failed   : exn                 -> ( _, underlying, completed) state
+    | Pending  : 'a callbacks        -> ('a, underlying, pending)   state
+    | Proxy    : ('a, _, 'c) promise -> ('a, proxy,      'c)        state
+
```
chambart: It is possible to avoid one indirection (and associated allocations) here.

```
type (_,_, _, _) kind =
    | Resolved : ('a,'a, underlying, completed) state
    | Failed   : (exn, _, underlying, completed) state
    | Pending  : ('a callbacks, 'a, underlying, pending)   state
    | Proxy    : (('a, 'b, 'c) promise, 'a, proxy,      'c)        state

type (_,_,_) promise = Promise : {
  mutable kind : ('value, 'a, 'u, 'c) kind;
  mutable value : 'value;
} -> ('a, 'u, 'c) promise
```

Of course the problem is that mutating the value requires both fields to be updated at once. You can't express that in the type system, hence you need to cheat some more with the type system to write a set function like:
```
val set : ('a, 'u, 'c) promise -> ('value, 'a, 'u, 'c) kind -> 'value -> unit
```

I'm not proud of this kind of suggestion, but in the case of Lwt the cost of traversing those pointers and allocating the state is far from negligible.

Drup: The problem of that proposition is that, additionally to the set trick, it requires inline records. Lwt is still compatible with ocaml 4.02, inline records come from 4.03.

src/core/lwt.ml
aantron: Very nice, I've done so.

This is actually enforcing even what *prevented* me from making the promise record type private – that the record would become immutable even inside the module `Main_internal_types`. `private` in a module is perfect for phantom types, though.

src/core/lwt.ml
aantron: This also reminds me of https://github.com/ocsigen/lwt/pull/94#issuecomment-59821934. We will have to keep it mind for the future.

aantron:
It's finally in! Thanks to all reviewers (whether commenting or not!) and stress testers!

aantron:
Hopefully, it's now open season on improving `lwt.ml` :)

Issue #350, opened 2017-05-10
Transient test suite failure: Unix.EBADF on close


emillon:
Hi,

From time to time the test suite fails with `Fatal error: exception Unix.Unix_error(Unix.EBADF, "close", "")`. I'm noticing it when building `pkcs11` on travis through [ocaml-ci-scripts](https://github.com/ocaml/ocaml-ci-scripts), which seems to run tests on dependencies as well.

Manually restarting it usually works, but this happens a lot (I'd say in 10% of builds)
so it should be easy to reproduce.

Here is the relevant build log excerpt:

```
#=== ERROR while installing lwt.3.0.0 =========================================#
# opam-version         1.2.2
# os                   linux
# command              make test
# path                 /home/opam/.opam/4.03.0/build/lwt.3.0.0
# compiler             4.03.0
# exit-code            2
# env-file             /home/opam/.opam/4.03.0/build/lwt.3.0.0/lwt-11742-736596.env
# stdout-file          /home/opam/.opam/4.03.0/build/lwt.3.0.0/lwt-11742-736596.out
# stderr-file          /home/opam/.opam/4.03.0/build/lwt.3.0.0/lwt-11742-736596.err
### stdout ###
# [...]
# rm -rf bisect*.out
# rm -rf _coverage/
# ./setup.exe -test 
# Running 110 tests for library "core".
Done. 0 test(s) skipped.
# Running 55 tests for library "unix".
(35/55) Running test "establish_server: order of exceptions" from suite "lwt_io"Running 1 tests for library "preemptive".
Done. 0 test(s) skipped.
# Running 19 tests for library "ppx".
Done. 0 test(s) skipped.
### stderr ###
# W: Cannot find source file matching module 'Lwt_unix' in library lwt-unix.
# W: Use InterfacePatterns or ImplementationPatterns to define this file with feature "source_patterns".
# W: Cannot find source file matching module 'Lwt_unix' in library lwt-unix.
# W: Use InterfacePatterns or ImplementationPatterns to define this file with feature "source_patterns".
# Fatal error: exception Unix.Unix_error(Unix.EBADF, "close", "")
# W: Test 'unix' fails: Command '/home/opam/.opam/4.03.0/build/lwt.3.0.0/_build/tests/unix/main.native' terminated with error code 2
# E: Failure("Tests had a 25.00% failure rate")
# make: *** [Makefile:42: test] Error 1
[lwt: ocamlfind remove] Command started
```

[...]

PR #355, opened 2017-05-15
Test Lwt_io.with_connection at a lower level

emillon:
As reported in #350, `Test_lwt_io` uses a fragile a test helper to close file descriptors in order to trigger errors in "context manager"-like functions. Those are functions that take a function in argument channel parameters: `with_connection` and `establish_server`.

Since the tear-down part of these functions is not exposed, it is hard to test. This exposes the tear-down part of `with_connection` as `with_close_connection` (in the `.mli` file but marked as nonpublic API and hidden in a stop comment). This function is tested and the corresponding fragile test is removed.

Similar tests exist for `establish_server`, but I just only deleted the fragile tests. Shall I extract and test its teardown part as well? It's a bit different since it handles `no_close` and runs `Lwt.async_exception_hook` by hand, but it should be possible to handle it in the same way.

Thanks!

### DIFF
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1399,9 +1399,7 @@ let open_connection ?fd ?in_buffer ?out_buffer sockaddr =
       Lwt_unix.close fd >>= fun () ->
       Lwt.fail exn)
 
-let with_connection ?fd ?in_buffer ?out_buffer sockaddr f =
-  open_connection ?fd ?in_buffer ?out_buffer sockaddr >>= fun (ic, oc) ->
-
+let with_close_connection f (ic, oc) =
   (* If the user already tried to close the socket and got an exception, we
      don't want to raise that exception again during implicit close. *)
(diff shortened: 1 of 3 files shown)


aantron:
Thanks for this.

> Similar tests exist for establish_server, but I just only deleted the fragile tests. Shall I extract and test its teardown part as well? It's a bit different since it handles no_close and runs Lwt.async_exception_hook by hand, but it should be possible to handle it in the same way.

I think it's okay to skip this for now. `Lwt_io.establish_server` is probably going to undergo some changes (though the API under that name hopefully won't be affected). That code probably needs to be rewritten anyway, and we can address testing it then.

I'm not sure why the Homebrew builds failed, but I am rerunning them now.

aantron:
Alright, the macOS/Homebrew failure is probably due to some kind of bit rot, and I don't think it has anything to do with the code in this PR, so I'm merging it. Thanks again.

emillon:
Great, thanks!

PR #358, opened 2017-05-17
unix: put fd non-blocking flag changes inside the lazy blocking value

### LINKED ISSUE #356: Lwt_unix.of_unix_file_descr should not eagerly call Unix.clear_nonblock
If `~blocking` is supplied to `Lwt_unix.of_unix_file_descr`, `Unix.clear_nonblock` will be called outside of the lazy value of the `Lwt_unix.file_descr` `blocking` field. If `blocking` is never forced (e.g. because the fd is only used for metadata operations), the resulting `fcntl` syscall is unnecessary. Additionally, on some operating systems like macOS, calling `fcntl` `F_SETFL` on certain fd types (like symlinks opened with `O_SYMLINK`) is an error (`ENOTTY` in that OS's case for some reason). This can cause a change from `?blocking:None` to `?blocking:(Some true)` to suddenly produce errors when encapsulating these fds.

The relevant section of code seems to be
<https://github.com/ocsigen/lwt/blob/master/src/unix/lwt_unix.cppo.ml#L324-L342>.

I'd be happy to prepare a PR if you agree with my analysis.

dsheets:
This design makes me mildly nauseous but with this patch it's merely really bad instead of catastrophic. Fixes #356.

### DIFF
--- a/src/unix/lwt_unix.cppo.ml
+++ b/src/unix/lwt_unix.cppo.ml
@@ -304,16 +304,16 @@ let is_blocking ?blocking ?(set_flags=true) fd =
         | Some state, false ->
             lazy(Lwt.return state)
         | Some true, true ->
-            Unix.clear_nonblock fd;
-            lazy(Lwt.return_true)
+            lazy(Unix.clear_nonblock fd;
+                 Lwt.return_true)
         | Some false, true ->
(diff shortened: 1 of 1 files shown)


aantron:
Thanks!

PR #362, opened 2017-05-19
Make the function cst_to_constr static in lwt_unix_unix.c

diml:
It is already publicly exposed [here](https://github.com/ocaml/ocaml/blob/trunk/otherlibs/unix/cst2constr.h), causing link error in some cases. The latter could probably be used directly now but I didn't check since when it's available.


aantron:
I think just making it `static` is fine. Thanks!

Issue #364, opened 2017-05-19
The paths of Lwt's installed cmi files


aantron:
@diml in https://github.com/ocaml/opam-repository/pull/9228#issuecomment-302756447:

> qcow-tool fails to build in one the datakit build. I check, it's because mirage-block-unix uses Lwt_unix but lwt.unix is missing from the jbuild. It's not detected at build time since the cmi files for lwt.unix are installed in the same directory as the cmi files for lwt.
>
> /cc @aantron, it'd be better to install the cmi for the various sub-libraries in sub-directories, to avoid problems like these. jbuilder installs the files for a library with findlib name a.b.c in <lib-dir>/a/b/c.

~~This seems like a problem with jbuilder being incompatible with non-jbuilder installs; on the other hand, maybe we can just~~ move these `.cmi` files in the next release, ~~and avoid friction~~.

cc @djs55 @rgrinberg 

EDIT: instructions for testing are [below](https://github.com/ocsigen/lwt/issues/364#issuecomment-303116926).

Drup:
The recent oasis feature called "FindlibDirectory" should allow to do that quite easily.

aantron:
So I guess this is a matter of

1. Adding `FindlibDirectory` to each Findlib library package in `_oasis` that has a `FindlibParent` ([example](https://github.com/ocsigen/lwt/blob/cf5580b6362a05087227f6b889a3a6e6094baee8/_oasis#L132-L136)).
2. Finding out what version of [OASIS](https://github.com/ocaml/oasis) implements `FindlibParent`, either by reading its [changelog](https://github.com/ocaml/oasis/blob/master/CHANGES.txt), or somehow through blame in that repo.
3. Making sure Lwt's [`opam` file](https://github.com/ocsigen/lwt/blob/cf5580b6362a05087227f6b889a3a6e6094baee8/opam/opam#L43) requires at least that version of OASIS.

diml:
FTR, jbuilder is compatible with non-jbuilder installs. It simply reads META files like ocamlfind. The issue I was talking about in the PR is the following:

1. create a file referencing Lwt_unix
2. compile it with `ocamlfind ocamlc -package lwt -c ...`
3. link an exectuable with `ocamlfind ocamlc -package lwt -o ...`

[...]

Issue #365, opened 2017-05-20
Port the build system to jbuilder + Configurator (?)


aantron:
cc @rginberg

@diml: We have a question: would current jbuilder be able to build Lwt?

c-cube:
If you find a good solution for cppo and the docs, I'm interested too.

rgrinberg:
Yes, cppo is easy to handle. Especially since we only have a couple of files being preprocessed.

For docs, we need only wait for https://github.com/janestreet/jbuilder/pull/74 to be finished. Which is should be done before 1.0 is released.

diml:
Yh, there are [two methods](http://jbuilder.readthedocs.io/en/latest/quick-start.html#using-cppo) shown in the manual for cppo.

Otherwise yh jbuilder should be able to build Lwt without problems.

andrewray:
```
opam pin add -k git lwt https://github.com/andrewray/lwt#jbuild
```

or clone and

```
jbuilder runtest
```

I've done a good chunk of the grunt work, leaving just the difficult bits!

* No enabling/disabling of the sub-libraries (though the discover tools do run).  Everything is assumed on. 
 Exactly how this can/should be expressed through jbuilder needs some discussion.

* I've reorganised the discovery build process a bit, however, it is still basically the same code.  One approach might be to modify the discover script to generate sexps (like configurator does)

* A couple of generated modules are not hidden (Lwt_config,  Lwt_unix_jobs_generated at least).  I think the directory organisation will have to change to work around that.

* Lwt_glib/ssl/react are called Lwt.glib/ssl/react.  This just needs new opam files at the top level and a couple of minor changes to the jbuild files.

* I have found 1 issue in utop - Lwt_main isn't automatically found.

* The ppx is based on ocaml-migrate-parsetree

aantron:
> leaving just the difficult bits!

😆 

> - No enabling/disabling of the sub-libraries (though the discover tools do run). Everything is assumed on.
Exactly how this can/should be expressed through jbuilder needs some discussion.

Some info:

`Lwt_ssl`, `Lwt_glib`, and `Lwt_react` should always be off in the main build, I just didn't bother editing it to remove them. They were factored out in #335. These (obviously) correspond to `--disable-ssl`, `--disable-glib`, and `--disable-libev`.

We want to do the same thing to the PPX (#338) and Camlp4 (#370).

I don't know yet enough jbuilder to recommend a specific course of action based on this. This is FYI.

[...]

PR #374, opened 2017-05-23
Port to jbuilder

andrewray:
#365 

This splits lwt into multiple opam packages, and ties them back together with an empty `lwt` package with appropriate dependancies and a special META file to map to the old package names.

<br/>

EDIT: Note from @aantron:
- See the [summary below](https://github.com/ocsigen/lwt/pull/374#issuecomment-307212409).
- This will be **merged** as #391.


lwt-core.opam
```diff
@@ -0,0 +1,25 @@
...
+]
+homepage: "https://github.com/ocsigen/lwt"
+doc: "https://ocsigen.org/lwt/manual/"
+bug-reports: "https://github.com/ocsigen/lwt/issues"
+license: "LGPL with OpenSSL linking exception"
+dev-repo: "https://github.com/ocsigen/lwt.git"
+build: [ [ "jbuilder" "build" "-p" name "-j" jobs ] ]
+build-test: [ [ "jbuilder" "runtest" ] ]
+depends: [
+  "ocamlfind" {build & >= "1.5.0"}
```
rgrinberg: Is findlib still explicitly necessary? jbuilder doesn't really use it anywhere.

lwt-core.opam
```diff
@@ -0,0 +1,25 @@
...
+  "Jérôme Vouillon"
+  "Jérémie Dimino"
+]
+homepage: "https://github.com/ocsigen/lwt"
+doc: "https://ocsigen.org/lwt/manual/"
+bug-reports: "https://github.com/ocsigen/lwt/issues"
+license: "LGPL with OpenSSL linking exception"
+dev-repo: "https://github.com/ocsigen/lwt.git"
+build: [ [ "jbuilder" "build" "-p" name "-j" jobs ] ]
+build-test: [ [ "jbuilder" "runtest" ] ]
```
rgrinberg: You should add `"-p" name` in the build-test as well.

lwt-ppx.opam
```diff
@@ -0,0 +1,29 @@
...
+doc: "https://ocsigen.org/lwt/manual/"
+bug-reports: "https://github.com/ocsigen/lwt/issues"
+license: "LGPL with OpenSSL linking exception"
+dev-repo: "https://github.com/ocsigen/lwt.git"
+build: [ [ "jbuilder" "build" "-p" name "-j" jobs ] ]
+build-test: [ [ "jbuilder" "runtest" ] ]
+depends: [
+  "ocamlfind" {build & >= "1.5.0"}
+  "jbuilder" { build & >= "1.0+beta9" }
+  "lwt-core"                            # is this really needed, or only sensible?
```
rgrinberg: I would remove it if it's not entirely necessary. It will give opam an opportunity to parallelize the builds more.

lwt-ppx.opam
```diff
@@ -0,0 +1,29 @@
...
+  "Jérôme Vouillon"
+  "Jérémie Dimino"
+]
+homepage: "https://github.com/ocsigen/lwt"
+doc: "https://ocsigen.org/lwt/manual/"
+bug-reports: "https://github.com/ocsigen/lwt/issues"
+license: "LGPL with OpenSSL linking exception"
+dev-repo: "https://github.com/ocsigen/lwt.git"
+build: [ [ "jbuilder" "build" "-p" name "-j" jobs ] ]
+build-test: [ [ "jbuilder" "runtest" ] ]
```
rgrinberg: This build-test needs to be fixed as well.

lwt-preemptive.opam
```diff
@@ -0,0 +1,27 @@
+opam-version: "1.2"
+name: "lwt"
+version: "dev"
```
rgrinberg: IMO you should remove the version/name from the opam

src/ppx/.merlin
```diff
@@ -1,4 +1,8 @@
...
-
-REC
\ No newline at end of file
+B ../../_build/default/src/ppx
+FLG -ppx "../../_build/default/.ppx/ppx_tools_versioned.metaquot_404/ppx.exe --as-ppx --cookie 'library-name="ppx_lwt"'"
+FLG -w -40
+PKG compiler-libs
+PKG compiler-libs.common
+PKG ocaml-migrate-parsetree
+PKG ppx_tools_versioned
```
rgrinberg: .merlin file should be deleted

META.lwt.template
```diff
@@ -0,0 +1,38 @@
+# JBUILDER_GEN
+
+version = "dev"
```
rgrinberg: version here should be removed I believe

lwt-core.opam
andrewray: Yeah.  I added the package stanza to all the runtest aliases then forgot this bit!

lwt-core.opam
andrewray: Not sure.  Jbuilder says it uses it if available, but I am not sure what difference it atually makes.

META.lwt.template
andrewray: Yup, and possibly substituted eventually by topkg or similar.

andrewray:
> How come the old build system isn't being deleted? Do we still need it for something.

I've been using the _oasis file as a reference.  It can disappear soon.

> I wonder if lwt-syntax is such a great name. Perhaps something like lwt-camlp4 or even lwt-camlp4-deprecated to signify to people that this package shouldn't really be used.

Whatever's prefered.  It's disturbingly easy to refactor jbuild files...(in fact I used had lwt-camlp4, then changed it to match the old naming).

lwt-core.opam
rgrinberg: If ocamlfind is installed it will prefer `$ ocamlfind printconf path` rather than just looking into opam directly. IMO, reducing dependence on findlib is a good thing as it's just a needless layer of indirection. But it's a minor issue.

andrewray: Ok, makes sense.  Sounds like I should get rid of it.

aantron:
> > I wonder if lwt-syntax is such a great name. Perhaps something like lwt-camlp4 or even lwt-camlp4-deprecated to signify to people that this package shouldn't really be used.
>
> Whatever's prefered. It's disturbingly easy to refactor jbuild files...(in fact I used had lwt-camlp4, then changed it to match the old naming).

I agree with @rgrinberg that moving away from `lwt.syntax` is desirable. I'm not sure how that interacts with legacy code that expects `lwt.syntax`, and the vague goal of not naming an OPAM package differently from a Findlib package, but no new users should be affected by such difficulties.

rgrinberg:
> and the vague goal of not naming an OPAM package differently from a Findlib package, but no new users should be affected by such difficulties.

It's contrary to this particular goal, but IMO it's still worth thinking about. I feel uneasy about this dead weight being carried around with lwt and being named so inconspicuously. @hhugo opted to name the camlp4 extension `js_of_ocaml-camlp4` fyi.

andrewray:
FYI;

| new opam pkg | old findlib name |
|-|-|
| lwt-core | lwt |
| lwt-unix | lwt.unix |
| lwt-preemptive | lwt.preemptive |
| lwt-log | lwt.log |
| lwt-ppx | lwt.ppx |
| lwt-syntax | lwt.syntax |
| lwt-react | lwt.react |
| lwt-simple-top | lwt.simple-top |
| lwt_glib | lwt_glib |
| lwt_ssl | lwt_ssl |
| lwt_react | lwt_react |

In addition, the findlib names for the syntax extension package break down into;

| new findlib name | old findlib name |
|-|-|
| lwt-syntax | lwt-syntax |
| lwt-syntax.log | lwt.syntax.log |
| lwt-syntax.options | lwt.syntax.options |

Note that in the new scheme, the package called `lwt` doesn't refer to any code and simply installs a META file which maps the new names to the old names.  It should probably come with a depreciation notice.

If you want to change the syntax name to `lwt-camlp4[-depreciated]`, we can do so for now without breaking (too much) code by keeping the appropriate mapping.

I note one further inconsistency which is that I have used `lwt-xxx` whereas the old external packages use `lwt_xxx`.  I prefer the former, but don't mind that much either way.  I think the whole namespace should be as consistent as possible, though, so it should be fixed.  Again the META mapping trick should keep things fairly compatible whatever is chosen.

Drup:
If we have a compatibility layer that rewires everything, we might as well take this occasion to change the names that we don't like, such as `lwt.syntax`.

lwt-ppx.opam
Drup: I disagree, `lwt-ppx`'s META should mention that `lwt-core` is needed at runtime, which only make sense if `lwt-core` is an opam dep.

lwt-preemptive.opam
Drup: Please keep the version, it's better for pinning.

avsm:
omg, this is amazing! :-) See https://github.com/mirage/mirage/issues/818#issuecomment-303756041

rgrinberg:
Hmm configurator seems to depend on base. Is it acceptable to have that as a build time dependency for Lwt?

andrewray:
> Hmm configurator seems to depend on base. Is it acceptable to have that as a build time dependency for Lwt?

It's gone now anyway.

The discover script is rather more sophisticated than configurator at the moment.

andrewray:
Given @diml's comment I refactored the jbuild files to model the current lwt package organisation.  Fewest changes possible and all that.  This is easy to change and can be addressed seperately.

I have tested with ocaml 4.02.3, 4.03.0, and 4.04.0 on Ubuntu and it seems OK.  Related issues;

* the jbuild-workspace feature doesn't work due to the ppx for some reason. 
* lambda-term doesn't compile on MacOSX for a reason I don't understand yet.
* no Windows, BSD or Android testing

The main remaining issue is how to properly expose configuration of `lwt.unix` to its discover script.  The question is how to tell jbuilder about the conf-libev opam package.

A minor unresolved issue is how to hide the Lwt_config and Lwt_unix_jobs_generated modules.

andrewray:
By the way, jbuilder organises sub-packages into their own directories so it should fix #364 (and break some dependent packages which are to some extent broken anyway - see bad_packages.md).

aantron:
Thanks for the ongoing work!

> A minor unresolved issue is how to hide the Lwt_config and Lwt_unix_jobs_generated modules.

It would be nice to hide these, but I think not critical. I think we can assume that nobody has modules with these names, and we can check that in OPAM.

jbuild-workspace.dev
```diff
@@ -0,0 +1,3 @@
+(context ((switch 4.02.3)))
+(context ((switch 4.03.0)))
+(context ((switch 4.04.0)))
```
aantron: Maybe 4.04.1?

aantron:
@andrewray you can add `[skip appveyor]` to your commit messages to avoid triggering AppVeyor until ready. I'm getting a bunch of AppVeyor spam, which I don't actually mind, but in case you do, see previous sentence :)

andrewray:
It's actually what I am working on at the moment!

andrewray:
Ok, so progress;

* I rebased the code to the latest commit, and sorted the conflicts.

* The mac build problem was just opam leaving some old files behind when pinning.  It works fine.

* Configuration is now available by running src/util/configure.ml with the options -use-libev, -use-pthread, -android-target, -libev-default.  I think this provides the appropriate options for development testing.  If not run at all, the same defaults as _oasis had should be calculated.

* The modules which should be hidden live on.  There might be a fix from jbuilder in the future, but it doesn't seem to be a big issue.

* workspace builds/tests across compiler versions work with the latest dev version of jbuilder. 

* Travis and Appveyor support has been updated.  Comments to follow...

* Documentation generation on Jenkins is broken.  I think we could knock up some jbuilder rules to run ocamldoc appropriately, but I think this will have to wait until jbuilder supports meta-programming features if we don't want to write them out per file.

Regarding Travis - apart from its usual flakiness, everything bar the 4.05 build works (due to jbuilder itself not building).

On appveyor, the cygwin port works and the mingw64 port builds but can't run its tests.  There is some sort of path issue I think.  It's strange that jbuilder works when called from opam, but doesn't work when called directly from the shell.  If we figure out why that is, then I suspect we can fix this issue.

rgrinberg:
> Documentation generation on Jenkins is broken. I think we could knock up some jbuilder rules to run ocamldoc appropriately, but I think this will have to wait until jbuilder supports meta-programming features if we don't want to write them out per file.

@diml just merged his odoc support into master so perhaps we can simply replace the old docs with that: http://jbuilder.readthedocs.io/en/latest/api-doc.html

andrewray:
Unfortunately I think the stuff being built on jenkins is related to the stuff that goes up on the oscigen website which jbuilder wont support.

Drup:
The stuff that goes to the ocsigen website is generated by using [a custom plugin for ocamldoc](https://github.com/ocsigen/wikidoc) that generates wikicreole syntax and added to [this repo](https://github.com/ocsigen/ocsigen.org-data). It's a bit involved and clunky, but it has significant advantages (versioning of APIs, integrated manual, richer markup) that odoc does not have just yet. 

I'm not sure what's @aantron's plan, but I would tend to think we should try to keep it (especially since I think it's not *that* hard to do it as an additional script, in the style of odig).

aantron:
@andrewray 

> It's actually what I am working on at the moment!

Apologies, I was responding to some earlier commits :)

> workspace builds/tests across compiler versions work with the latest dev version of jbuilder.

Does this mean we need to wait for it to be released? I'm assuming that this is an optional facet of the new Lwt build system, so no.

> Regarding Travis - apart from its usual flakiness, everything bar the 4.05 build works (due to jbuilder itself not building).

Probably not a blocking issue, as long as jbuilder gets 4.05 support before 4.05 actually releases :) I hope that's safe to assume.

> On appveyor, the cygwin port works and the mingw64 port builds but can't run its tests. There is some sort of path issue I think. It's strange that jbuilder works when called from opam, but doesn't work when called directly from the shell. If we figure out why that is, then I suspect we can fix this issue.

Not sure by simple inspection.

Should you want to connect to one of the AppVeyor VMs, you can try downloading an RDP client for your platform, and following [these](https://www.appveyor.com/docs/how-to/rdp-to-build-worker/) instructions. Add those lines to `appveyor.yml` in your branch. I think the environment variable is a new thing, you should be able to get by without it.

I've done this many times, and spent many, many hours RDPd into AppVeyor VMs for Lwt and other projects :)

> Unfortunately I think the stuff being built on jenkins is related to the stuff that goes up on the oscigen website which jbuilder wont support.

Yes. I think that is also the main reason for keeping the Jenkins build around at this point.

> I'm not sure what's @aantron's plan, but I would tend to think we should try to keep it (especially since I think it's not that hard to do it as an additional script, in the style of odig).

The new manual (so far a myth, #304) will eventually be hosted independently, and odoc is likely to play a role in its generation, but I had intended to retain an instance of it in the Ocsigen hosted docs – unless the people from Ocsigen are okay with them not being there anymore. I'm not immediately sure who to contact for opinion on that. @Drup, @vasilisp?

aantron:
And I agree with @Drup, having (tried to) work on that doc generation briefly, that it's probably not hard to do as a separate process from the main build system.

andrewray:
> workspace builds/tests across compiler versions work with the latest dev version of jbuilder.
> Does this mean we need to wait for it to be released? I'm assuming that this is an optional facet of the new Lwt build system, so no.

It's a neat dev feature that allows you to build and test a project against multiple compiler versions at once.  It had a small bug related to ppx now fixed.

> Regarding Travis - apart from its usual flakiness, everything bar the 4.05 build works (due to jbuilder itself not building).
> Probably not a blocking issue, as long as jbuilder gets 4.05 support before 4.05 actually releases :) I hope that's safe to assume.

I think that's most likely.

> On appveyor, the cygwin port works and the mingw64 port builds but can't run its tests. There is some sort of path issue I think. It's strange that jbuilder works when called from opam, but doesn't work when called directly from the shell. If we figure out why that is, then I suspect we can fix this issue.
> Not sure by simple inspection.
> 
> Should you want to connect to one of the AppVeyor VMs, you can try downloading an RDP client for your platform, and following these instructions. Add those lines to appveyor.yml in your branch. I think the environment variable is a new thing, you should be able to get by without it.
> 
> I've done this many times, and spent many, many hours RDPd into AppVeyor VMs for Lwt and other projects :)

I have one of those IE test VMs installed which I can test with, so I hopefully don't need RDP.

I am pretty sure this issue is well understood by the likes of @fdopen ...

vasilisp:
> The new manual (so far a myth, #304) will eventually be hosted independently, and odoc is likely to play a role in its generation, but I had intended to retain an instance of it in the Ocsigen hosted docs – unless the people from Ocsigen are okay with them not being there anymore. I'm not immediately sure who to contact for opinion on that. @Drup, @vasilisp?

I think we all (= Ocsigen team) agree that we eventually need to migrate away from the current site/doc infrastructure. Despite its fancy features, it is just cumbersome to maintain, and quite unmaintained in practice. We had better streamline our processes and focus on the interesting aspects of our projects.

I would very much like to see statically-generated docs under `ocsigen.github.io/docs` (or something), and an almost empty, trendy, Bootstrap-y `ocsigen.org` that only contains buzzwords and links to the relevant places.

Therefore, If you have a reasonable plan for generating static HTML from jbuilder, I would be supportive. If you have ideas for handling existing wikidoc markup in a not-too-tedious way, we will happily borrow your ideas for the other Ocsigen projects :).

andrewray:
So, 2 green ticks for appveyor and travis.

* For the failing mingw run I just ran opam with `-t` to run its build test.  I don't know why running in opam works, but a direct command doesn't - but it doesn't matter - the result is the same.

* I have temporarily commented out testing on 4.05.

* There was a consistently failing test on 4.02.3+libev (time out running the unix test).   To try and find the problem I made the tests verbose - and for some reason that has fixed it.  Ho-hum.

I have also added a makefile and some rules to generate documentation various ways

* make doc - builds the shiny new odoc documentation through jbuilder
* make doc-api-html - builds ocamldoc documentation
* make doc-api-wiki - builds the wikidoc documentation

I was unable to understand the jenkins build infrastructure enough to actually integrate the code and fix that test.  Perhaps someone who knows how those scripts work can sort this.

I can't think of anything else that needs to be done.

andrewray:
> Therefore, If you have a reasonable plan for generating static HTML from jbuilder, I would be supportive. If you have ideas for handling existing wikidoc markup in a not-too-tedious way, we will happily borrow your ideas for the other Ocsigen projects :).

Generating odoc documentation with jbuilder very simple.

For the wiki stuff I just nabbed `js_of_ocaml/doc/Makefile` and modified it.  It was pretty easy as well, so long as you can just glob the mli files.

aantron:
Very impressive! 🎉  🎉  🎉 

For Jenkins, I think the error is triggered by [this code in Lwt's `.jenkins.sh`](https://github.com/andrewray/lwt/blob/d827388f0779b1fef11d5349b67db1ffa64e6555/.jenkins.sh#L10-L11). It should probably be changed to call the new target, then [copy](https://github.com/andrewray/lwt/blob/d827388f0779b1fef11d5349b67db1ffa64e6555/.jenkins.sh#L10-L11) the `.wiki` files to the places given by the environment variables you see there.

More broadly, for Jenkins, we should probably keep only a latest OCaml build in Jenkins (4.04 at the moment), and use it only to generate the `ocsigen.org` docs, using this new targets. cc @vasilisp. I guess we still want a PR build, so this comment is for builds of `master`.

For testing, I think Travis covers everything that is done in Jenkins now, but is much easier for contributors and maintainers to deal with (e.g. in this PR). This was why I originally added it (plus, it has Mac VMs).

<br/>

Thanks again. I'll give this a thorough review in some days, as I have a backlog of work to do. You're encouraged to take a nice, well-deserved break :) and thank you very much for taking this task on!

_opam-files/META.lwt.template
```diff
@@ -0,0 +1,69 @@
...
+
+package "react" (
+  requires = "lwt-react"
+  description = "[compatibility] maps to lwt-react"
+  exists_if = "../lwt-react/lwt_react.cma"
+)
+
+package "syntax" (
+  requires = "camlp4 lwt-camlp4 lwt.camlp4.options"
+  exists_if = "../lwt-camlp4/lwt_camlp4.cma"
```
avsm: Is this relative path and `exists_if` necessary here?  Might as well just make them always exist, since the new package scheme is based around explicit OPAM dependencies anyway.

andrewray: About half way through the PR I was taught that splitting up the packages wasn't necessary, so I decided to keep the current package/sub-package organisation.

Whether splitting up the package is desirable, is a bigger conversation, and generally speaking I think a positive move, so instead of removing the various opam files I put them in this `_opam-files` directory.  With these files, and some grepping and editing of jbuild files, you can radically change the package organisation in 10 minutes which I think is an awesome jbuilder feature.

Specifically, this lwt.opam file was designed to model the current lwt package on top of a potential new set of packages.  The exists_if lets it expose these new packages *if installed*, which it otherwise doesn't know the status of.  There are issues with this scheme - while it helps compatibility, doing `opam remove lwt` doesn't actually uninstall lwt - just the package name mappings.  I also find the relative directory hijinks rather smelly.

andrewray:
> For Jenkins, I think the error is triggered by this code in Lwt's .jenkins.sh. It should probably be changed to call the new target, then copy the .wiki files to the places given by the environment variables you see there.

Actually it's the line before, `setup-dev.exe`.  I couldn't figure out from the jenkins logs how this was built with the wikidoc package.  Also, I don't know where the script that does this comes from.


> More broadly, for Jenkins, we should probably keep only a latest OCaml build in Jenkins (4.04 at the moment), and use it only to generate the ocsigen.org docs, using this new targets. cc @vasilisp. I guess we still want a PR build, so this comment is for builds of master.


Like the odoc tool, I belive the wikidoc package requires 4.03.

aantron:
> Actually it's the line before, setup-dev.exe.

That's what I meant, that was a pasting error.

I believe this is what invokes wikidoc: https://github.com/ocsigen/lwt/blob/428592a1c73ef33a34c5756cc7ed0ea64dbdad52/myocamlbuild.ml#L147-L179

aantron:
Not sure why that Travis row is hanging now. I don't see any relevant change since the last time it passed successfully. I restarted it several times, with the same outcome. I think we can assume that it's not a problem, and all the builds are working – thanks!

aantron:
Well, looking through the build history, that row has been flaky for a while.

andrewray:
I made a virtualbox ubuntu precise vm, installed ocaml through the @avsm ppa and went through the steps to compile everything and I can repeat the Travis failure.

It's difficult to see, as I think somehow jbuilder is suppressing some of the output of the test run, but the test runs up to;

```
(29/53) Running test "establish_server_1: shutdown: client closes first" from suite "lwt_io"
```

where it hangs.  I compiled up the current master as well and it has the same behaviour when I run the unix tests by hand.

```
andyman@precisely:~/dev/lwt-ref$ gdb ./_build/tests/unix/main.native 
GNU gdb (Ubuntu/Linaro 7.4-2012.04-0ubuntu2.1) 7.4-2012.04
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
For bug reporting instructions, please see:
<http://bugs.launchpad.net/gdb-linaro/>...
Reading symbols from /home/andyman/dev/lwt-ref/_build/tests/unix/main.native...done.
(gdb) run
Starting program: /home/andyman/dev/lwt-ref/_build/tests/unix/main.native 
warning: no loadable sections found in added symbol-file system-supplied DSO at 0x7ffff7ffa000
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Running 53 tests for library "unix".
(1/53) Running test "openfile: O_CLOEXEC" from suite "lwt_unix"[New Thread 0x7ffff6447700 (LWP 26808)]
(29/53) Running test "establish_server_1: shutdown: client closes first" from suite "lwt_io"^C
Program received signal SIGINT, Interrupt.
0x00007ffff71e4dd3 in epoll_wait () from /lib/x86_64-linux-gnu/libc.so.6
(gdb) bt
#0  0x00007ffff71e4dd3 in epoll_wait () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff7bd12f2 in ?? () from /usr/lib/libev.so.4
#2  0x00007ffff7bd3733 in ev_run () from /usr/lib/libev.so.4
#3  0x00000000004a38bf in ev_loop (flags=<optimised out>, loop=0x78a3d0)
    at /usr/include/ev.h:820
#4  lwt_libev_loop (val_loop=<optimised out>, val_block=<optimised out>)
    at src/unix/lwt_libev_stubs.c:126
#5  0x0000000000435b19 in camlLwt_engine__fun_1968 ()
    at src/unix/lwt_engine.ml:187
#6  0x0000000000438328 in camlLwt_main__run_1018 () at src/unix/lwt_main.ml:40
#7  0x000000000043433a in camlTest__loop_tests_1037 () at tests/test.ml:79
#8  0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#9  0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#10 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#11 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#12 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#13 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#14 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#15 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#16 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#17 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#18 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#19 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#20 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#21 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#22 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#23 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#24 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#25 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#26 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#27 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#28 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#29 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#30 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#31 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#32 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#33 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#34 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#35 0x000000000043436c in camlTest__loop_tests_1037 () at tests/test.ml:80
#36 0x000000000042b0be in camlMain__entry () at tests/unix/main.ml:44
#37 0x0000000000428249 in caml_program ()
#38 0x00000000004beb2e in caml_start_program ()
#39 0x00000000004beca0 in caml_system__code_end ()
#40 0x0000000000000000 in ?? ()
```

I think there is a problem here, but I dont think jbuilder introduced it.

bad_packages.md
```diff
@@ -0,0 +1,43 @@
+A few errors found by randomly installing some lwt packages.
+
+All seem to point to libraries using lwt.unix but only linking to lwt.
```
aantron: Agreed, thanks for checking this. I don't think this blocks the PR. We should remove this file from the branch; I've copied it out locally (or we can save it in a gist). I'll do builds of reverse dependencies and notify the package maintainers after the PR.

I'll give some thought to whether this means we have to do this conversion softer until 4.0.0. But, regardless, if we really have to, we can, I think, make this transition smoother by adding some symlinks after install in 3.1.0, or moving installed files and patching `META`, ugly as that is. In any case, we can decide that after merge.

_opam-files/lwt-core.opam
```diff
@@ -1,30 +1,24 @@
 opam-version: "1.2"
-name: "lwt_ssl"
-version: "1.0.1"
+name: "lwt"
```
aantron: `lwt-core` might be too confusing with `core-lwt` already a package in OPAM, but we can think about that later.

.gitignore
```diff
@@ -1,8 +1,4 @@
 _build
-src/unix/lwt_config.ml
-src/unix/lwt_config.h
-src/unix/lwt_unix_jobs_generated.ml
-src/unix/jobs-unix/
 setup.data
```
aantron: I guess the `setup.*` files can also be removed from `.gitignore`.

README.md
```diff
@@ -119,7 +119,7 @@ version,
 opam source --dev-repo --pin lwt
 ```
 
-This will also install the development dependency OASIS.
+This will also install the development dependency jbuilder.
```
aantron: `jbuilder` isn't a development-only dependency like OASIS is/was, so we can remove this entire sentence. The current `README.md` has a completely different contributing section anyway, though.

lwt.opam
```diff
@@ -0,0 +1,47 @@
...
+messages: [
+  "For module Lwt_ssl, please install package lwt_ssl"
+  {ssl:installed & !lwt_ssl:installed}
+  "For module Lwt_glib, please install package lwt_glib"
+  {lablgtk:installed & !lwt_glib:installed}
+  "For module Lwt_react, please install package lwt_react"
+  {react:installed & !lwt_react:installed}
+]
+post-messages: [
+  "Lwt+jbuilder has some major breaking changes!!!"
```
aantron: I don't think this is true anymore :)

lwt.opam
```diff
@@ -0,0 +1,47 @@
...
+dev-repo: "https://github.com/ocsigen/lwt.git"
+build: [ 
+  [ "ocaml" "src/util/configure.ml" "-use-libev" "%{conf-libev:installed}%" ]
+  [ "jbuilder" "build" "-p" name "-j" jobs ] 
+]
+build-test: [ [ "jbuilder" "runtest" "-p" name ] ]
+depends: [
+  "ocamlfind" { build & >= "1.5.0" }
+  "jbuilder" { build & >= "1.0+beta9" }
+  "ocaml-migrate-parsetree"
```
aantron: Do we need to depend on omp explicitly? I think `ppx_tools_versioned` pulls that in. However, I'm not yet familiar with the best practices for these packages.

.travis.yml
```diff
@@ -8,8 +8,8 @@ matrix:
       env: COMPILER=4.03
     - os: linux
       env: COMPILER=4.04
-    - os: linux
-      env: COMPILER=4.05
+    #- os: linux
+    #  env: COMPILER=4.05
```
andrewray: I think the bug may actually be related to a problem with 4.05+beta3, and it will work with 4.05+trunk

bad_packages.md
andrewray: To support camlp4 properly (for the time being) we will need to hand maintain a META file anyway.

Personally, I would just break those packages as they are strictly wrong, and can cause upstream problems for their users anyway.  But maybe that's just not playing very nicely...

lwt.opam
andrewray: Not sure, but you are correct about the implicit dependency, so I removed the explicit one here.

aantron:
Thanks! *Almost* everything looks good. There is only **one** problem remaining: I still get the error message

    Error: Unbound module Camlp4

on a fresh switch, due to the spurious Camlp4 dependency, which I guess you realize. I do see that dummy `META` file added by OPAM, which does seem like an upstream issue. I think we need to work around this somehow, as for some users, the system `opam` and/or `ocaml` packages install system Camlp4. This seems to be true at least for me on Homebrew.

Also, if you know who to cc for the upstream issue, please cc them. Otherwise, I will take a look later.

<br/><hr/>

Note: this doesn't block the PR, but js_of_ocaml 2.8.4 doesn't build with development Lwt now, apparently due to the use of `dev` as the version in `META`. This seems to be due to js_of_ocaml using additional version checking, instead of relying on OPAM constraints:

```
# malformed input: no version number
# Js_of_ocaml requires Lwt version 2.3.0 at least.  Please upgrade.
```

Working around that, js_of_ocaml 2.8.4 build fails with:

```
# File "log/lwt_log_js.mli", line 32, characters 14-28:
# Error: Unbound module Lwt_log
```

If this is another dependency issue, we might have to edit `META` even more heavily. I didn't test with js_of_ocaml `master`, so I'm not ready to open an issue on it, but cc @hhugo.

bad_packages.md
aantron: I agree with breaking those packages, and in this case, I think there is no technical reason not to do it (e.g. this alone won't create any inconsistent dependency diamonds). But since we still have to maintain the `META`, and warn reverse dependency maintainers about various other things, I think it's not too much extra work to do this nicely :)

andrewray:
> Thanks! Almost everything looks good. There is only one problem remaining: I still get the error message
> 
> Error: Unbound module Camlp4
> on a fresh switch, due to the spurious Camlp4 dependency, which I guess you realize. I do see that dummy META file added by OPAM, which does seem like an upstream issue. I think we need to work around this somehow, as for some users, the system opam and/or ocaml packages install system Camlp4. This seems to be true at least for me on Homebrew.
> 
> Also, if you know who to cc for the upstream issue, please cc them. Otherwise, I will take a look later.

I opened issue ocaml/opam-repository#9440 to discuss this, though whether ocamlfind, camlp4 or opam packaging in general are the cause is an open question.

The only way jbuilder knows not to build a sub-package is if it cannot find all the ocamlfind dependencies (and you say optional in the jbuild file).  The only workaround I can think of would be to move the camlp4 package to a separate library.

Interestingly, I don't get quite the same error as you - it does find camlp4, but the version is all wrong and it fails.

andrewray:
> Working around that, js_of_ocaml 2.8.4 build fails with:
> 
> # File "log/lwt_log_js.mli", line 32, characters 14-28:
> # Error: Unbound module Lwt_log
> If this is another dependency issue, we might have to edit META even more heavily. I didn't test with js_of_ocaml master, so I'm not ready to open an issue on it, but cc @hhugo.

This is a rather confusing naming convention;  `Lwt_log` is part of `lwt.unix`.  `js_of_ocaml` is compiling with `lwt.log` but not `lwt.unix`, so same old problem (though arguably an understandable mistake).

aantron:
> The only workaround I can think of would be to move the camlp4 package to a separate library.

Ok, what I'm proposing is not pretty, but can we wrap some kind of script around jbuilder, split the jbuilder build system into two (one for `lwt.syntax`, one for everything else) and do our own detection? We will maintain this janky thing until 4.0.0, when `lwt.syntax` will become its own package and/or the various upstream projects will have fixed the problem on their end.

> This is a rather confusing naming convention; Lwt_log is part of lwt.unix. js_of_ocaml is compiling with lwt.log but not lwt.unix, so same old problem (though arguably an understandable mistake).

Fair enough. I guess we have our first maintainer ping. I'll ping again when actually addressing this issue.

> Interestingly, I don't get quite the same error as you - it does find camlp4, but the version is all wrong and it fails.

I think the version would be correct in my case, since Homebrew installed 4.04.0 as my system compiler (and corresponding Camlp4), and I've been testing in a 4.04.1 switch.

andrewray:
> The only workaround I can think of would be to move the camlp4 package to a separate library.
> Ok, what I'm proposing is not pretty, but can we wrap some kind of script around jbuilder, split the jbuilder build system into two (one for lwt.syntax, one for everything else) and do our own detection? We will maintain this janky thing until 4.0.0, when lwt.syntax will become its own package and/or the various upstream projects will have fixed the problem on their end.

Since we getting janky....I've put a not too terrible hack in which can disable camlp4 by writing a `jbuild-ignore` file from `src/util/configure.ml`.  The opam file can then coordinate compilation of lwt.syntax.

src/util/configure.ml
```diff
@@ -30,6 +32,17 @@ let main () =
...
+  (* '-use-camlp4 false' will write a jbuild-ignore file directing jbuilder to ignore
+   *                     the camlp4 directory. 
+   * '-use-camlp4 true' (or none) will write an empty file which does nothing 
+   *                     except to potentially overwrite any previous version. 
+   * 
+   * This is a workaround required to overcome some weird camlp4 packaging behaviour
+   * where ocamlfind sometimes installs a dummy META file for it even though it's 
+   * not actually installed. *)
+  let f = open_out "src/jbuild-ignore" in
+  (if !use_camlp4 = Some(false) then Printf.fprintf f "camlp4" else ());
```
aantron: I think `camlp4` should be off by default; that might also help with the overall review comment.

aantron:
The diff looks and works good now. Thanks!

What's left is the history rewrite for blame and squashing. I made a list of the renamed files that git seems to misjudge in the net diff:

```
src/ssl/opam -> lwt_react.opam
src/react/opam -> lwt_ssl.opam
```

I noticed that we lost the version constraints in `lwt_react` on `lwt` (`>= 3.0.0`) and `react` (`>= 1.0.0`); hadn't seen that because of this misjudging by git. Perhaps adding them back in will be enough for git to get these files right.

And the infamous

```
src/ppx/ppx_lwt.ml -> src/ppx/ppx_lwt.ml
```

`ppx_lwt.ml` should be traceable to `ppx_lwt_ex.ml` instead. If fixing `lwt_react.opam`is helpful, this should be the only file that needs special attention in deleting/renaming. I think everything else can be squashed into one commit. I don't know if you'd like to keep the original history of this PR visible. If you wish, we can copy out all the detailed commit messages out into a big comment; but they were already helpful to me during review, and some of my comments are replying to them.

If you rebase (please :)), I think you will get a conflict in `README.md`. That one should probably just be resolved by adopting whatever is in `master`.

I'm keeping an instance of the `jbuild` branch, as it is now, locally, so reviewing the delete/rename/squash/rebase should be easy as long as the net diff is about the same (hopefully changing only in the constraints of `lwt_react` and the omission of changes to `README.md`). I'll also manually check the blame chain for `ppx_lwt.ml`. That should be about all.

aantron:
After merge, I'll update affected issues, open any follow-on issues (e.g. cleaning up the packaging and build by 4.0.0 along the lines of https://github.com/ocsigen/lwt/pull/374#issuecomment-303498536), check the Ocsigen doc generation, and fix some absolutely trivial questions and nits that don't really affect anything.

g2p:
Did something just break the Lwt syntax extension?
I've switched to this branch for the express purpose of using it with jbuilder.
Sorry if I'm misusing it somehow.

andrewray:
> Did something just break the Lwt PPX syntax extension?
> I've switched to this branch for the express purpose of using it with jbuilder, and it was working fine a few commits ago.
> Sorry if I'm misusing it somehow.

Hmmm...I can't think what apart from the custom META file might have changed.  Can you provide an error messgage, ocaml version, jbuilder version etc?

g2p:
> Hmmm...I can't think what apart from the custom META file might have changed. Can you provide an error messgage, ocaml version, jbuilder version etc?

I get:

    # Error: Uninterpreted extension 'lwt'.

When building my code that uses lwt.ppx. The `jbuild` file contains

    (preprocess (pps (lwt.ppx cstruct.ppx)))

which used to work.

g2p:
Somehow upgrading cstruct from 2.4.1 to 3.0.0 breaks parsing lwt syntax.
No idea why though.

andrewray:
I think the ppx issue was with cstruct and has been fixed.

Also good to see you're working with Ocaml 4.05, which addresses another outstanding issue!

aantron:
Ok, I think the PPX issue doesn't block this PR then. Thanks for finding and diagnosing it :)

Regarding 4.05, I'll edit the Travis build shortly after merging, and we'll have testing on it again.

aantron:
Summary of what happened in this PR:

In a massive effort, @andrewray ported Lwt to jbuilder, but *without* splitting Lwt up into multiple OPAM packages, as originally stated. The Lwt project will do that in follow-on work over the next few months, mainly to avoid breaking projects that depend on Lwt (see #293 regarding Lwt's soft breakage policy).

- The Lwt build is now faster: 5s, compared to 15s.
- Lwt should now be composable with other jbuilder projects.

This achievement is not to be understated: Lwt is one of the few OCaml projects tested on Cygwin and Win32, in addition to having many legacy quirks that we can only eliminate slowly. Thanks @andrewray!

Several things are temporarily broken by this PR, which we will fix later (and definitely before the next release). See the message of commit 2163fb1f19826d4a38e4cd4583b75e8b2c9b9921 for details.

This PR will be merged as #391, which has a simplified history for `master`.

aantron:
FYI the ocsigen.org manual seems to have survived this PR :) If any minor issues develop, we can fix them later.

avsm:
I just wanted to express my appreciation for all the hard work that went into this PR.  We've been porting a large chunk of the Mirage package repository with a view to using a unified `jbuilder build`, and Lwt being ported gives us a large boost in this direction. Thank you!

Issue #367, opened 2017-05-20
Update comments at the top of all code files


aantron:
The first three lines of `.mli` and `.ml` files in Lwt currently look like this ([example](https://github.com/ocsigen/lwt/blob/cf5580b6362a05087227f6b889a3a6e6094baee8/src/core/lwt.ml)):

```ocaml
(* Lightweight thread library for OCaml
 * http://www.ocsigen.org/lwt
 * (Module|Interface) Lwt
```

we want them to look like this ([example](https://github.com/ocsigen/lwt/blob/fdff5c09d47d2b020d4998ebed922acf383a2e9d/src/core/lwt.ml)):

```ocaml
(* OCaml promise library
 * http://www.ocsigen.org/lwt
```

- "Lightweight thread" terminology is deprecated. We especially want to avoid it in `.mli` files, since people read those – though most people probably don't care about the file headers.
- Repeating the module name (which is based on the file name) in the header is redundant and fragile.

The header for `src/core/lwt.ml` is already being changed in #354, but the rest need to be done.

jsthomas:
I'll work on this issue.

aantron:
Done in #376. Thanks @jsthomas!

PR #376, opened 2017-05-24
Updated comments at the top of all code files

jsthomas:
This change addresses Issue #367. It updates the header comments in all `.ml/.mli` files to remove references to lightweight threads in favor of promises. Module/Interface/Program names are also removed.


aantron:
Wow, thanks! I didn't fully realize there were that many files, but I guess it all adds up.

PR #368, opened 2017-05-22
enhance/fix invalid argument exception string payload

domsj:

PR #382, opened 2017-05-28
Accurate exceptions

### LINKED ISSUE #369: Accurate messages in exceptions raised by Lwt
For example, trying this in the REPL

```
let _, r = Lwt.wait ();;
Lwt.wakeup r ();;
Lwt.wakeup r ();;
```

i.e. calling `Lwt.wakeup r ()` when `r` is already resolved, raises `Invalid_argument "Lwt.wakeup_result"`. This is misleading, because `Lwt.wakeup_result` is a different function in Lwt. `Lwt.wakeup` just happens to be implemented in terms of it. This message can send a reader to the wrong part of their code. The message is generated [here](https://github.com/ocsigen/lwt/blob/master/src/core/lwt.ml#L334).

Some other instances of this are being fixed in #368, but there are many more in Lwt.

To do this, one should search Lwt for `invalid_arg`, `fail_invalid_arg`, `failwith`, `fail_with`, `Failure`, `Invalid_argument`, and make sure that the string argument to each instance of those actually reflects the API function that the user called. For helpers that get called from multiple API functions, that means the API function name will have to be passed to the helper.

cc @domsj

raphael-proust:
Fixes #369

### DIFF
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -320,7 +320,7 @@ let state_of_result
 let make_value v = Result.Ok v
 let make_error e = Result.Error e
 
-let wakeup_result t result =
+let wakeup_gen fun_name t result =
   let t = repr_rec (wakener_repr t) in
   match t.state with
     | Sleep sleeper ->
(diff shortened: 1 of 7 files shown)


aantron:
LGTM right now. Let me know when you are done making changes, and I'll merge.

raphael-proust:
It's good for me.

It could be improved, but I think such improvements are low priority. I'll check the issue tracker and see if there are other things I can tackle.

aantron:
Agreed. Thanks for the thorough work!

I'll take care of merging this into #354, once that PR is being merged.

Related #368.

Issue #385, opened 2017-05-31
[easy] Tests for the simpler modules of Lwt (multiple PRs)


aantron:
This issue is about writing new test cases for some of the simpler helper modules of the Lwt [core library](#about-core-library):

- [x] ~~[`Lwt_result`](https://ocsigen.org/lwt/dev/api/Lwt_result) ([`src/core/lwt_result.ml`](https://github.com/ocsigen/lwt/blob/428592a1c73ef33a34c5756cc7ed0ea64dbdad52/src/core/lwt_result.ml#L25))~~ – ~~not tested~~ 100% coverage by @jsthomas.
- [x] ~~[`Lwt_condition`](https://ocsigen.org/lwt/dev/api/Lwt_condition) ([`src/core/lwt_condition.ml`](https://github.com/ocsigen/lwt/blob/428592a1c73ef33a34c5756cc7ed0ea64dbdad52/src/core/lwt_condition.ml#L29))~~ – ~~not tested~~ 100% coverage by @jsthomas.
- [x] ~~[`Lwt_list`](https://ocsigen.org/lwt/dev/api/Lwt_list) ([`src/core/lwt_list.ml`](https://github.com/ocsigen/lwt/blob/428592a1c73ef33a34c5756cc7ed0ea64dbdad52/src/core/lwt_list.ml#L22))~~ – [~~22%~~ ~~59%~~ 100% coverage](https://ocsigen.github.io/lwt/coverage/file0009.html) by @ryanslade.
- [x] ~~[`Lwt_mvar`](https://ocsigen.org/lwt/dev/api/Lwt_mvar) ([`src/core/lwt_mvar.ml`](https://github.com/ocsigen/lwt/blob/428592a1c73ef33a34c5756cc7ed0ea64dbdad52/src/core/lwt_mvar.ml#L29))~~ – [~~33%~~ 100% coverage](https://ocsigen.github.io/lwt/coverage/file0014.html) by @jsthomas.
- [ ] [`Lwt_pool`](https://ocsigen.org/lwt/dev/api/Lwt_pool) ([`src/core/lwt_pool.ml`](https://github.com/ocsigen/lwt/blob/428592a1c73ef33a34c5756cc7ed0ea64dbdad52/src/core/lwt_pool.ml#L23)) – not tested. 2 APIs, 74 lines.

Each of these is a separate small project.

Not included in this issue are:

- `Lwt` itself. This got thorough test cases recently in #339.
[...]

PR #392, opened 2017-06-11
Lwt_list: More tests.

ryanslade:
Also formatted the code using ocp-indent and renamed the tests so they
have descriptive names instead of just numbers.

### DIFF
--- a/tests/core/test_lwt_list.ml
+++ b/tests/core/test_lwt_list.ml
@@ -153,31 +153,31 @@ let test_rev_map f =
   return (after = [6;4;2])
 
 let suite = suite "lwt_util" [
-    test "0"
+    test "iter_p"
       (fun () ->
          test_iter Lwt_list.iter_p [1;0;1];
          test_exception Lwt_list.iter_p;
          return true);
 
-    test "1"
+    test "iter_s"
       (fun () ->
          test_iter Lwt_list.iter_s [1;0;0];
          test_exception Lwt_list.iter_s;
          return true);
 
-    test "2"
+    test "map_p"
       (fun () ->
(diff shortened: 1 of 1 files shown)


ryanslade:
One other thing, can we translate the comment in test_exception into English?

Drup:
Here's the translation

```
Is this the desired behavior ? 
We might wish for iter and map to pass their parameters in Lwt.apply.
Another solution would be to have two binds, one tail recursive, the other not.
```

Tbh, I'm not sure this comment has any purpose.

aantron:
Thanks!

It seems that comment can be deleted indeed.

Could you split the commit into two, one that does whitespace and renamings, and another that adds the new tests? The reason is that it will make it both easier to review now (not *that* important, but nice), and also easier to read the history later (more important). For the lines that changed only whitespace, a person reading the blame will see something like "whitespace and renamings" and be able to immediately recognize the commit doesn't change logic, and click past it to the previous blame.

ryanslade:
@aantron I've split the changes into three commits, please take another look.

aantron:
Great, thank you!

This brings the coverage of `Lwt_list` up to 100%.

We also probably need to test everything against exceptions raised by `f`, and check that the `_s` functions really run `f` in series, and the `_p` functions really run it in parallel. After solving #347 (currently being worked on), we should probably also add tests checking that these functions are stack-safe. I'll open a new issue about all of this. I think we can check off the `Lwt_list` line in #385 :)

aantron:
The follow-on is #395. Also, cherry-picked these commits onto `working-coverage`, which we need until Bisect_ppx is running on the new jbuilder build system.

Issue #328, opened 2017-03-26
Proposal: (softly) replace Lwt_unix by libuv/uwt


aantron:
This diagram from the libuv docs sketches the logical structure of libuv. `Lwt_unix` has basically the same structure.

![architecture](https://cloud.githubusercontent.com/assets/12073668/24334315/b3aacc70-122d-11e7-930a-d8fd57dd8bac.png)

*(source: http://docs.libuv.org/en/v1.x/_images/architecture.png)*

However, libuv is maintained by more people, and has a larger user base, so one can expect it to be of higher quality, especially over time. Also, the Lwt Unix binding's Windows side really doesn't get enough attention.

I propose that we eventually somehow "bless" [`uwt`](https://github.com/fdopen/uwt), help maintain it, and/or merge it into Lwt, and start discouraging use of the current `Lwt_unix` and the rest of the current Unix binding.

`uwt` already seems to have work-alike modules for most (all?) of the Unix binding, which might make this process easier.

This would, however, introduce a new external dependency.

There are some other possibilities, like mio, but I haven't really looked at them yet. Discussion very welcome.

cc @fdopen

Drup:
> However, libuv is maintained by more people, and has a larger user base, so one can expect it to be of higher quality, especially over time. Also, the Lwt Unix binding's Windows side really doesn't get enough attention.

While that might be true, the same certainly cannot be said about `uwt` itself.

c-cube:
This sounds like a lot of work if you don't want to break compat. `Lwt_unix` maybe hard to maintain (I have looked once or twice upon the internal type definitions for processes, and was surprised by the amount of magic required), but it provides a relativement good and comprehensive API which might not be easy to reproduce. Note I'm mostly interesting in `Lwt_process`, which provides timeouts and the likes.

Maybe having `lwt.uv` or `lwt.uwt` as a new sublibrary would be a proper migration path. Especially if tests that ensure that both `lwt.uv` and `lwt.unix` behave the same can be written.

rgrinberg:
The arguments why a libuv base should be favored are convincing to me but I agree with c-cube that this does not have to mean removing the classical `lwt.unix`.

I'm also a little concerned about the story of Lwt_engine. Today Lwt_engine seems to be the preferred way of supporting various event loops, but that clearly isn't enough to support something like libuv (at least the way uwt does it). Seems strange to have an API for custom event loops that can't support even Lwt's own needs (libuv support).

mfp:
[...]

PR #399, opened 2017-06-12
[WIP] Initial implementation of getcwd

raphael-proust:


src/unix/lwt_unix.cppo.ml
```diff
@@ -1149,6 +1149,14 @@ let chdir name =
   else
     run_job (Jobs.chdir_job name)
 
+external getcwd_job : unit -> string job = "lwt_unix_getcwd_job"
+
+let getcwd () =
+  if Sys.win32 then
+    Lwt.return (Unix.getcwd ())
```
raphael-proust: That's inspired by the other functions in the module. I don't think it's optimal. It'll probably get better if #328 gets resolved?

tests/unix/test_lwt_unix.cppo.ml
```diff
@@ -677,11 +677,28 @@ let bind_tests = [
         (fun () -> Lwt_unix.close socket));
 ]
 
+let dir_tests = [
+    test "getcwd"
+    (fun () ->
+      Lwt_unix.getcwd () >>= fun (_:string) ->
+      Lwt.return_true
```
aantron: One thing that we can do for the basic test, is compare the result of `Lwt_unix.getcwd` to the result of `Unix.getcwd` from the stdlib.

tests/unix/test_lwt_unix.cppo.ml
```diff
@@ -677,11 +677,28 @@ let bind_tests = [
         (fun () -> Lwt_unix.close socket));
 ]
 
+let dir_tests = [
```
raphael-proust: Ideally we would have another test such as

```
    test "get other cwd"
    (fun () ->
      Lwt_unix.getcwd () >>= fun here ->
      Lwt_unix.chdir somewhere >>= fun () ->
      Lwt_unix.getcwd () >>= fun there ->
      Lwt_unix.chdir there >>= fun () ->
      Lwt.return (here = there)
    );
```

But:

1. How to choose `somewhere`?
2. How to deal with failures which leave the rest of the tests in an unexpected directory?

I think there are things in setup.ml about directories to run tests in. I'll investigate.

tests/unix/test_lwt_unix.cppo.ml
raphael-proust: Good idea. I'll do that.

src/unix/lwt_unix.cppo.ml
aantron: Yep, this is what is meant by realtively poor Windows support :/ A lot of it will get better when #328 is resolved. However, libuv doesn't include a non-blocking `getcwd`, so we will still need to have it threaded, and address Windows. I believe uwt does that to a great extent. The first thing to do for that would be to check whether the underlying Windows function can block or not – if it can't, then we don't have anything to worry about.

IMO, it's okay to put this off until later.

tests/unix/test_lwt_unix.cppo.ml
aantron: You should probably work with the jbuilder build system is `master`, because we won't be able to merge anything that uses `setup.ml`. The only reason I'm aware of at this point to use the older build system is coverage, but I hope to have that solved in a couple days.

One way to address this is:

1. We have this helper: https://github.com/ocsigen/lwt/blob/ba4deb79166235cfe32d5515dd99adc56793274d/tests/test.mli#L52. I think it's okay not to clean up the temporary directory, though I'm not fully sure how this interacts with the new jbuilder build system.
2. Ignore this :) With the test written using the helper, it shouldn't be a problem.

aantron: And `lwt_unix_unix.c` was renamed by the jbuilder port to `lwt_unix_unix.h`, so this is causing the merge conflict. Hopefully, git can resolve that for you automatically :)

raphael-proust: I'll rebase this branch on the most recent upstream master.

raphael-proust:
I messed up the history on this one. I made a clean version. I'll make a new PR and close this one.

(The clean version is the same without the spurious merge and it's on top of current master.)

PR #403, opened 2017-06-14
Getcwd

### LINKED ISSUE #342: getcwd in lwt
I'm using lwt in a project and I have no previous experience with lwt, but is there a function in lwt for grabbing the current working directory? If not then is it safe to use `Sys.getcwd` with lwt while remaining async?

Aside from that command, is it safe to assume that any command in Unix, Sys, etc. that lwt did not wrap is safe to use in lwt while remaining async?

raphael-proust:
A clean history for the same commits as the previous PR.

Fixes #342


src/unix/lwt_unix_unix.h
```diff
@@ -3008,6 +3009,41 @@ CAMLprim value lwt_unix_bind_job(value fd, value address)
...
+
+static value result_getcwd(struct job_getcwd *job)
+{
+    LWT_UNIX_CHECK_JOB(job, job->result == NULL, "getcwd");
+    value result = caml_copy_string(job->result);
+    lwt_unix_free_job(&job->job);
+    return result;
+}
+
+CAMLprim value lwt_unix_getcwd_job()
```
aantron: This function should take an argument `value unit` (see elsewhere in the file). Technically, it's not necessary, since this is C and we don't try to use the argument, but OCaml is passing that argument to this function when it calls it (and its value is zero, though it may be encoded as 1, I forgot how `()` is actually encoded).

src/unix/lwt_unix_windows.h
```diff
@@ -575,5 +575,6 @@ LWT_NOT_AVAILABLE3(unix_writev)
 LWT_NOT_AVAILABLE3(unix_writev_job)
 LWT_NOT_AVAILABLE1(unix_iov_max)
 LWT_NOT_AVAILABLE2(unix_bind_job)
+LWT_NOT_AVAILABLE1(getcwd_job)
```
aantron: Should be `unix_getcwd_job`, I believe.

aantron:
Canceling the Travis macOS build, so it doesn't hold up the next build – so ignore the upcoming Travis "failure" :)

aantron:
Wow, AppVeyor hates us today. The build sometimes hangs, and we haven't had time to debug it. It's not because of this PR. I restarted the build.

aantron:
😓  thanks!

PR #400, opened 2017-06-13
Added .ocp-indent rules and fixed lwt_stream indentation

Richard-Degenne:
This PR is a first step towards a prettification of the Lwt code base, as per issue #393.

The `.ocp-indent` file will enable future edits and developments to follow sane indentation rules.

Issue #393, opened 2017-06-11
[easy] Fix whitespace, formatting, prettify the code


aantron:
When code doesn't look good, it is difficult to read and understand. We want Lwt to be easy to work on.

Basically, if you run into any strange indentation, please go ahead and fix it :) Here are some ~~immediate~~ examples from the recent past: [[~~1~~](https://github.com/ocsigen/lwt/blob/1eb217a6ac9bf86d522e5e31c1d1a10f53e59df3/src/unix/lwt_io.ml#L1044-L1050)] [[~~2~~](https://github.com/ocsigen/lwt/blob/1eb217a6ac9bf86d522e5e31c1d1a10f53e59df3/src/unix/lwt_io.ml#L1279-L1281)] [[~~3~~](https://github.com/ocsigen/lwt/blob/1eb217a6ac9bf86d522e5e31c1d1a10f53e59df3/src/unix/lwt_unix.cppo.mli#L75-L79)] [[~~4~~](https://github.com/ocsigen/lwt/blob/1eb217a6ac9bf86d522e5e31c1d1a10f53e59df3/src/unix/lwt_unix.cppo.ml#L484)].

It's recommended not to do this to a file that has an [open, recent pull request](https://github.com/ocsigen/lwt/pulls), to avoid a merge conflict. "Recent" is probably less than a couple months old. Also, check any [in-progress](https://github.com/ocsigen/lwt/labels/in%20progress) or [easy](https://github.com/ocsigen/lwt/labels/easy) issues for an idea of what is being worked on. On the other hand, don't worry too much – in case of a bad conflict, maintainers exist to make sure nobody has to do too much merging work :)

In particular, at the moment, `lwt.ml` (#354) probably shouldn't be touched, although you can edit the *new* one's whitespace by opening pull requests against the `wat` branch that #354 comes from.

Some pointers:

- Lwt uses 80-column lines. Comment lines should use up as much of that as possible.
- Two-space tabs (tabs as spaces!).
- I obviously prefer the formatting as in #354 ([example](https://github.com/ocsigen/lwt/blob/fdff5c09d47d2b020d4998ebed922acf383a2e9d/src/core/lwt.ml#L1710)), but any improvement is welcome.
- ~~If you know how to use [ocp-indent](https://github.com/OCamlPro/ocp-indent), teach us/me, or tell what it's capable of :)~~ Thanks @Richard-Degenne, @vasilisp!
- If you want to discuss what the whitespace *should* be, please go ahead!

[...]

Issue #408, opened 2017-06-14
Pick a C function as a model, and document it really well


aantron:
Adding a C call is currently an involved process, rife with potential snags, even for what should be simple work. We should pick a (relatively!) easy function, such as `getcwd`, and really document it right inside the code. A subset of things that should be covered:

- https://github.com/ocsigen/lwt/issues/342#issuecomment-307655393 (errno and the opaque Lwt macros).
- https://github.com/ocsigen/lwt/pull/403#pullrequestreview-44083154 (re. system calls not implemented on Win32).
- https://github.com/ocsigen/lwt/pull/403#discussion_r122006161 (passing unit).

PR #409, opened 2017-06-15
Indented most of the source code

Richard-Degenne:
This PR is a follow-up of PR #400 about issue #393.

Most of the `ml` files of the source folder have undergone an `ocp-indent` command, followed by a quick review. Some remarks about this:

- Globally, the indentation was consistant within files, but varied a bit between files;
- Most of the changes are on pattern matches;
- I've left `core/lwt.ml` and `unix/lwt_unix.cppo.ml` untouched, so as not to cause conflicts with ongoing PRs;
- `ocp-indent`ing `mli` files wasn't really necessary, and it kept breaking comments indentation;
- `ocp-indent` got the camlp4 and ppx files pretty wrong, and I indented them mostly by hand. Since they are quite obscure to me, these ones may require some extra attention.

### DIFF
--- a/src/preemptive/lwt_preemptive.ml
+++ b/src/preemptive/lwt_preemptive.ml
@@ -74,16 +74,16 @@ struct
   let get t =
     let rec await_value t =
       match t.cell with
-        | None ->
-            Condition.wait t.cv t.m;
-            await_value t
-        | Some v ->
-            t.cell <- None;
-            Mutex.unlock t.m;
-            v
+      | None ->
+        Condition.wait t.cv t.m;
+        await_value t
+      | Some v ->
+        t.cell <- None;
+        Mutex.unlock t.m;
+        v
     in
-      Mutex.lock t.m;
-      await_value t
+    Mutex.lock t.m;
+    await_value t
 
   let set t v =
     Mutex.lock t.m;
@@ -136,10 +136,10 @@ let make_worker () =
 (* Add a worker to the pool: *)
 let add_worker worker =
   match Lwt_sequence.take_opt_l waiters with
-    | None ->
-        Queue.add worker workers
-    | Some w ->
-        Lwt.wakeup w worker
+  | None ->
+    Queue.add worker workers
+  | Some w ->
+    Lwt.wakeup w worker
 
 (* Wait for worker to be available, then return it: *)
 let get_worker () =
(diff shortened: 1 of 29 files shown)


raphael-proust:
AFAIAC, there are no PRs locking `unix/lwt_unix.cppo.ml` right now.

There are probably going to be some changes in the short term (e.g., #408) but there are no PR for these.

aantron:
Thanks! I'll review this in a bit. Just want to say that I'm minutes or hours away from submitting a PR for #408, but it won't touch the `.ml` file, at least for now. So, I think you can add `lwt_unix.cppo.ml` as @raphael-proust suggests (thanks again) :)

src/preemptive/lwt_preemptive.ml
```diff
@@ -207,20 +207,20 @@ let detach f args =
...
-        (* Put back the worker to the pool: *)
-        add_worker worker
-      else begin
-        decr threads_count;
-        (* Or wait for the thread to terminates, to free its associated
-           resources: *)
-        Thread.join worker.thread
-      end;
-      Lwt.return_unit)
+       if worker.reuse then
```
aantron: IMO ocp-indent gets this kind of thing wrong. It shouldn't cause an extra space if you have to add a `(` to a function (or any other expression). Is it easy to fix in the ocp-indent settings? If not, it's fine – no need to go through all these files. The PR does much more good than harm, even with this nit.

src/unix/lwt_log.ml
```diff
@@ -244,47 +244,47 @@ let syslog ?(template="$(date) $(name)[$(pid)]: $(section): $(message)") ?(paths
...
-        syslog_socket := Some x;
-        Lwt.return x
+      syslog_connect paths >>= fun x ->
+      syslog_socket := Some x;
+      Lwt.return x
   in
   make
     ~output:(fun section level lines ->
-               Lwt_mutex.with_lock mutex
+      Lwt_mutex.with_lock mutex
```
aantron: I'm really glad things like this are getting fixed.

src/preemptive/lwt_preemptive.ml
Richard-Degenne: Hmm, I guess it does this to align everything two spaces after the `fun` keyword, which is one column after the beginning of the line. I agree it looks kinda wrong. I'll see what I can do.

src/util/configure.ml
```diff
@@ -22,26 +22,26 @@ let args = [
...
-   *                     the camlp4 directory.
-   * '-use-camlp4 true' will write an empty file which does nothing
-   *
-   * This is a workaround required to overcome some weird camlp4 packaging behaviour
-   * where ocamlfind sometimes installs a dummy META file for it even though it's
-   * not actually installed. *)
-  let f = open_out "src/jbuild-ignore" in
-  (if !use_camlp4 = Some(true) then () else Printf.fprintf f "camlp4");
-  close_out f
+    let f = open_out "src/unix/lwt_config" in
```
aantron: Looks like ocp-indent got this wrong. I wonder why? I don't think this should be indented more than the previous line (am I missing something?).

aantron:
So most of it looks good – thanks for doing this. The only other nit I have is that the glib commit should be titled something like ``Fixed `glib` indentation`` like all the others are, so people can click past it in blame.

src/preemptive/lwt_preemptive.ml
Richard-Degenne: Available configuration doesn't seem to provide an option for this behavior.

Is it worth opening an issue on their side?

aantron: Sure, they might be able to answer it, fix it, or teach me/us why the ocp-indent way is the right way :) Also cc @vasilisp for any experience with this.

I don't think this blocks the PR though, unless you want to wait for it.

src/util/configure.ml
Richard-Degenne: There's a similar issue reported [here](https://github.com/OCamlPro/ocp-indent/issues/229). I'll keep an eye on it.

aantron: Thanks, subscribed to that issue.

aantron:
Ok, looks great. I'm happy to have this in. I'll merge once the builds finish, but they will, of course, succeed.

PR #410, opened 2017-06-15
Less brutal C stubs – document getcwd stubs as model functions

### LINKED ISSUE #408: Pick a C function as a model, and document it really well

aantron:
This is towards making the writing of C stubs less of a painful experience.

The comments are a draft at this point. I've reached a state of stupor proof-reading it, so I'm posting it for fresh eyes, and I'll take a look with a clearer head tomorrow or later.

This is based on  *my* side of the recent experience in #342, #399, #403, which were done by @raphael-proust. So, @raphael-proust, you're welcome to, and I would be grateful if you, comment on anything here. Also, if you want to make bigger alterations, you can PR against the `getcwd-model` branch, or expand on this stuff in a separate PR after this one is merged.

Of course, reviews are welcome from everyone else as well :)

Once we figure out decent comments for `getcwd`, I will write a guide in [`CONTRIBUTING.md`](https://github.com/ocsigen/lwt/blob/master/doc/CONTRIBUTING.md) on adding C stubs. The guide will link to `getcwd` for people to use as a reference.

EDIT: Resolves #408.

### DIFF
--- a/src/unix/lwt_unix_unix.h
+++ b/src/unix/lwt_unix_unix.h
@@ -21,7 +21,12 @@
  * 02111-1307, USA.
  */
 
-/* Unix (non windows) version of stubs. */
+/* Unix (non-Windows) version of Lwt C stubs.
+
+   Implementing an Lwt C stub can be a bit challenging. See lwt_unix_getcwd_job
+   (search for it in your text editor) for a well-documented "model"
+   function, including conceptual documentation, practical considerations,
+   common pitfalls, etc. */
 
 #define ARGS(args...) args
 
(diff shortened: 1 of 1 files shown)


aantron:
And by expand, I mean including throw away and rewrite, as needed :)

src/unix/lwt_unix_unix.h
```diff
@@ -3009,38 +3014,200 @@ CAMLprim value lwt_unix_bind_job(value fd, value address)
...
+       Throughout Lwt, blocking C calls run in worker threads can't write
+       directly into OCaml strings, because the OCaml garbage collector might
+       move the strings after the pointer has already been passed to the call,
+       but while the call is still blocked. Bigarrays don't have this problem,
+       so pointers into them are passed to blocking C calls, saving a copy.
+
+       For jobs that return integers or other kinds of values, it is necessary
+       to use the various `Int_val`, `Long_val` macros, etc. See
+
+         http://caml.inria.fr/pub/docs/manual-ocaml/intfc.html#sec415 */
```
raphael-proust: Maybe start with a generic recommendation (e.g., “Convert the result of the job to its OCaml representation”) and only after  that talk about the specific of this particular job (strings) and other values (bigarrays, ints, etc.).

src/unix/lwt_unix_unix.h
```diff
@@ -3009,38 +3014,200 @@ CAMLprim value lwt_unix_bind_job(value fd, value address)
...
 
+/* In the case of `Lwt_unix.getcwd`, the argument is `()`, which is represented
+   in C by one argument, which we conventually call `unit`. OCaml always passes
+   the same value for this argument, and we don't use it. */
 CAMLprim value lwt_unix_getcwd_job(value unit)
 {
+    /* Allocate the `job_getcwd` on the OCaml heap. Inside it, store its size,
+       and pointers to `worker_getcwd` and `result_getcwd`. Arguments must be
+       stored manually after the macro is called, but in the case of `getcwd`,
+       there are no arguments to initialize.
```
raphael-proust: Point at a function that has arguments.

src/unix/lwt_unix_unix.h
```diff
@@ -3009,38 +3014,200 @@ CAMLprim value lwt_unix_bind_job(value fd, value address)
...
+
+       The first argument is the name of the variable to be craeted to store
+       the pointer to the job `struct`, i.e.
+
+         struct job_getcwd *job = ...
+
+       The last argument is the number of bytes of storage to reserve in memory
+       immediately following the `struct`. This is for fields such as
+       `char data[]` at the end of the struct. It is typically zero. */
```
raphael-proust: Point to a function for which it isn't zero.

src/unix/lwt_unix_unix.h
```diff
@@ -3009,38 +3014,200 @@ CAMLprim value lwt_unix_bind_job(value fd, value address)
...
+      able to run them.
+    - Store the C call arguments, or references to them, so they can be accessed
+      by `worker_FOO` when it runs in the worker thread.
+    - Store the C call results, so they can be accessed by `result_FOO` in the
+      main thread.
+    - Be something that can be placed in queues, or manipulated otherwise.
+
+    So, there are three functions and a `struct` to define. It is important that
+    their names fit the pattern given above, because it is expected by Lwt job
+    helper macros.
```
aantron: Yeah, I was torn between that order and this one. In fact, this is what contributed the most to my writing fatigue :/ The back to functions is meant to be just a conclusion. I thought the functions (in particular, running in the worker thread) motivate the existence of the struct much more than the other way around, but it also makes sense to describe things in the order the code depends on them. This is what I can't address for a day or more :p

src/unix/lwt_unix_unix.h
```diff
@@ -3009,38 +3014,200 @@ CAMLprim value lwt_unix_bind_job(value fd, value address)
...
+    doesn't call the ones that aren't really defined.
+
+    The `LWT_NOT_AVAILABLEx` macros are used to define dummy symbols. `x` is the
+    number of arguments the stub takes. For example, the `getcwd` job is
+    currently not defined on Windows. For this reason, `lwt_unix_windows.h` has
+    `LWT_NOT_AVAILABLE1(unix_getcwd_job)`. The `lwt_` prefix is left off.
+
+    A typical way to discover that you need to use this kind of macro is to
+    submit a PR to the Lwt repo, and wait for the CI build bots to tell you. We
+    have pretty thorough CI coverage of the systems Lwt supports.
```
aantron: Ideally, yes, but in practice I would forget those macros, even after setting up the CI and watching it fail on my code and others' code in the past. But that's a comment on whether the documentation while make the CI-driven development obsolete. I agree the documentation itself should be changed :)

src/unix/lwt_unix_unix.h
```diff
@@ -3009,38 +3014,200 @@ CAMLprim value lwt_unix_bind_job(value fd, value address)
...
+
+
+    See inline comments in the implementation of the `getcwd` job below for
+    other details. */
 
 /*
 In the OCaml sources, getcwd is set as either
 - 4096 (in asmrun/spacetime.c, byterun/sys.c)
 - PATH_MAX, MAXPATHLEN, or 512 (in otherlibs/unix/getcwd.c)
 */
```
aantron: Yeah, it's far. I'm worried about the length of that line. Another option is to move it down immediately below the struct.

aantron:
Thanks for the review. I'll address it with modifications in a day (maybe two, tomorrow is going to be busy). You're welcome to interrupt that with a PR, if that is/was/becomes your intent :)

src/unix/lwt_unix_unix.h
raphael-proust: I'll give a stab at writing something.

I think that introduces the struct first also makes sense because we are addressing OCaml programmers which are used to type-before-functions.

src/unix/lwt_unix_unix.h
raphael-proust: Yes, just below the struct would also be good.

aantron:
@raphael-proust Ok, after an excessive delay (sorry), I made modifications to address your helpful comments above.

raphael-proust:
I think it looks good. We can see if that helps the next person along and maybe improve them further based on their feedback.

aantron:
Thanks for provoking this, reviewing it, and working on it!

Issue #402, opened 2017-06-13
build error


raphael-proust:
With the HEAD at `ba4deb79166235cfe32d5515dd99adc56793274d` I get the following error when trying to build.

```
$ jbuild clean
$ jbuild build
    […]
    ocamlc src/unix/lwt_unix_jobs_generated.{cmi,cmo,cmt} (exit 2)
(cd _build/default && /usr/bin/ocamlc.opt -w -40 -g -bin-annot -I /home/bnwr/.opam/system/lib/bytes -I /home/bnwr/.opam/system/lib/result -I /usr/lib/ocaml -I src/core -I src/logger -no-alias-deps -I src/unix -o src/unix/lwt_unix_jobs_generated.cmo -c -impl src/unix/lwt_unix_jobs_generated.ml)
File "src/unix/lwt_unix_jobs_generated.ml", line 1:
Error: I/O error: src/unix/lwt_unix_jobs_generated.ml: No such file or directory

Waiting for 1 job to finish.
$ jbuilder build
      ocamlc src/logger/lwt_log.cma
      ocamlc src/core/lwt.cma
    ocamlopt src/core/lwt_condition.{cmx,o}
      ocamlc src/ppx/ppx_lwt.{cmi,cmti}
      ocamlc src/unix/lwt_unix.{cmi,cmti}
   gen_stubs […]
    ocamldep src/unix/lwt_unix.depends.ocamldep-output
Internal error, please report upstream including the contents of _build/log.
Description: "Lwt_config" not found in map <dependency graph in _build/default/src/unix>
Backtrace:
Raised at file "src/import.ml", line 17, characters 43-65
Called from file "src/module_compilation.ml", line 44, characters 24-63
Called from file "src/build_system.ml", line 300, characters 18-35
Called from file "src/build_system.ml", line 276, characters 8-25
Called from file "src/build_system.ml", line 276, characters 8-25
Called from file "src/build_system.ml", line 276, characters 8-25
Called from file "src/build_system.ml", line 323, characters 17-47
Called from file "src/future.ml", line 88, characters 40-45
Called from file "src/future.ml", line 83, characters 65-70
Called from file "src/future.ml", line 37, characters 9-12

Waiting for 2 jobs to finish.
```

After that, all the build end with the same error and the same backtrace.

I'll also get in touch with the jbuilder team, it says to do so.

raphael-proust:
This happens if lwt is pinned to my local development repository (`$HOME/code/lwt`).

[...]

PR #413, opened 2017-06-18
add a check for left over oasis files

andrewray:
provide hints for cleaning up the repository

### DIFF
--- a/src/util/configure.ml
+++ b/src/util/configure.ml
@@ -20,8 +20,43 @@ let args = [
   "-use-camlp4", arg_bool use_camlp4, " when true enable camlp4 syntax extension";
 ]
 
+let oasis_files = [
+  "setup-dev.exe";
+  "setup.exe";
+  "setup.data";
+  "setup.log";
+  "src/unix/lwt_config.h";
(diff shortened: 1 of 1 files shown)


src/util/configure.ml
```diff
@@ -20,8 +20,43 @@ let args = [
...
+build system and run distclean ie
+
+$ git checkout 30c3fd86f9ddebc98e54276100a59c79436b174f
+$ make clean distclean
+$ git checkout master 
+$ git clean -ffd
+"
+
+let oasis_sanity_check () = 
+  if List.fold_left (fun b f -> b || (Sys.file_exists f)) false oasis_files then
```
aantron: Nit, but I believe this is better implemented by [`List.exists`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html#VALexists). This doesn't block the PR (it doesn't really matter), so I'll merge it in some hours, but if you'd actually like to address this, go ahead :p

aantron:
Related to #402.

aantron:
Thanks for taking care of this!

PR #416, opened 2017-06-23
Formatted C files

Richard-Degenne:
This PR is a continuation of the work started in PRs #400 and #409 about issue #393.

I added a `.clang-format` file enforcing Google-style formatting for C files. I also applied this style to Lwt's C codebase.

Overall, I'm happy with the way the files were indented, even if I find the 80-column limit too strict sometimes ([example](https://github.com/ocsigen/lwt/compare/master...Richard-Degenne:c-formatting?expand=1#diff-da2f75024a2915ba1c1e888053121f95R135)).

`clang-format` has quite a lot of options and if you have remarks on what the files look like, please go ahead, I should be able to find the right configuration. :)

### DIFF
--- a/src/unix/lwt_unix.h
+++ b/src/unix/lwt_unix.h
@@ -24,20 +24,21 @@
 
 #define CAML_NAME_SPACE
 
-#include <lwt_config.h>
 #include <caml/mlvalues.h>
 #include <caml/unixsupport.h>
+#include <lwt_config.h>
+
 #include <caml/socketaddr.h>
 
 /* The macro to get the file-descriptor from a value. */
 #if defined(LWT_ON_WINDOWS)
-#  define FD_val(value) win_CRT_fd_of_filedescr(value)
+#define FD_val(value) win_CRT_fd_of_filedescr(value)
 #else
-#  define FD_val(value) Int_val(value)
+#define FD_val(value) Int_val(value)
 #endif
 
 /* Macro to extract a libev loop from a caml value. */
-#define Ev_loop_val(value) *(struct ev_loop**)Data_custom_val(value)
+#define Ev_loop_val(value) *(struct ev_loop **)Data_custom_val(value)
 
 /* +-----------------------------------------------------------------+
    | Utils                                                           |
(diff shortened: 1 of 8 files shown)


aantron:
Thanks for looking into this!

I haven't reviewed it in any detail yet, but from the CI, it looks like `clang-format` may have been a little lossy :(

Richard-Degenne:
I ran the tests on my machine and had no errors. I'll look into it.

Richard-Degenne:
Nailed it. I think I messed with `lwt_unix_stubs.c` when browsing it. :man_facepalming:

aantron:
:scream: <- that should be a reaction option :)

aantron:
It looks pretty good.

> `clang-format` has quite a lot of options and if you have remarks on what the files look like, please go ahead, I should be able to find the right configuration. :)

Two things jump out:

- I think switching to 4-space indentation would be nice, as this is (or was?) a typical standard for C.
- It would be good to leave the curly braces on function headers (only) alone as they are now, i.e. on new lines.

aantron:
Thanks for adjusting the rules. The C files look a *lot* better than their state before this PR, IMO.

I believe two things remain:

- [ ] Apply the formatting to our `.h` files:
    ```
    ./src/unix/lwt_unix.h
    ./src/unix/lwt_unix_unix.h
    ./src/unix/lwt_unix_windows.h
    ```
- [ ] Squash this history down a bit. The main goal is to avoid doing two large whitespace changes back-to-back to the same files, since we really mean for it to be one. The original history (one commit to introduce `.clang-format` and one to apply the changes) is good, or any other history that achieves the goal. I can do this boring stuff, let me know if you'd like me to.

src/unix/lwt_unix.h
```diff
@@ -24,20 +24,20 @@
 
 #define CAML_NAME_SPACE
 
-#include <lwt_config.h>
 #include <caml/mlvalues.h>
-#include <caml/unixsupport.h>
 #include <caml/socketaddr.h>
+#include <caml/unixsupport.h>
+#include <lwt_config.h>
```
aantron: Argh, [this build failure](https://ci.appveyor.com/project/aantron/lwt/build/515/job/itjy9yrwatj5m9p7#L446) suggests that either `caml/unixsupport.h` or `lwt_config.h` has to manually be moved above `caml/socketaddr.h`. This is an upstream issue due to one of the headers not including the other one internally, but we have to work around it :/

Richard-Degenne: This is really weird, I don't have these errors when building on my machine.

Richard-Degenne: A trick is to have separate "paragraphs", clang-format will order includes by-paragraph. I've amended the commit. :)

aantron: Thanks, I'm learning :)

aantron:
Much appreciated. I find these files much easier to read now, hopefully others will too.

PR #420, opened 2017-06-25
Add a script to post-process lwt.install

andrewray:
This flattens the installation directory for the 3.1.0 release and
hopefully helps maintains compatibility with various packages that
dont link with lwt.XXX sublibraries correctly.

We use a post-process step rather than a custom hand-written install
file so that jbuilder can work out for us exactly which packages
were built.

The opam build step calls the post-processing script after running
jbuilder.

The META file is lightly modified to remove the directory
specifications.  They have been commented out for now so they can
be easily added back.

Finally, the install and uninstall steps in the Makefile were modified
to use opam-installer.  Invoking "jbuilder install" after the
post-processing step didn't seem to work.

### DIFF
--- a/META.lwt.template
+++ b/META.lwt.template
@@ -13,7 +13,7 @@ archive(native) = "lwt.cmxa"
 plugin(byte) = "lwt.cma"
 plugin(native) = "lwt.cmxs"
 package "log" (
-  directory = "log"
+  #directory = "log"
   version = "dev"
   description = "Logger for Lwt"
   requires = "bytes lwt result"
@@ -24,7 +24,7 @@ package "log" (
   exists_if = "lwt_log.cma"
 )
 package "ppx" (
-  directory = "ppx"
+  #directory = "ppx"
   version = "dev"
   description = "Lwt PPX syntax extension"
   requires(ppx_driver) = "compiler-libs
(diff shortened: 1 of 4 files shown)


Drup:
Is it really a problem to break packages that don't link properly, given that the fix is backward compatible ?

aantron:
@andrewray Thanks for looking into this!

@Drup We *will* break those packages 4.0.0, but in a civil fashion with forewarning. I believe we don't intend to keep this `.ml` script beyond that, but it serves a very useful purpose for the coming months.

For breaking now, yes, I believe it is a problem.

- It would cause unnecessary friction and disappointment for users that want to use 3.1.0 and/or just Lwt with Jbuilder, and are also using packages, including transitively, that haven't yet updated with the fix (i.e. we piss off users that are not "at fault", so to speak). We want to avoid: "if you want a feature or bugfix in 3.1.0, but also use package `XYZ`, sucks for you."
- We don't want to surprise users with a breaking change in a minor release.
- We don't want to force downstream maintainers to update their broken packages on our schedule.
- We want to announce this problem ahead of time to maintainers and users of packages in the "deep ecosystem" outside OPAM, and the best way I know of reaching them so far is through a message on an OPAM release.
- Et cetera...

Basically, a few hours of @andrewray's (thanks!) and my time is probably going to save a much larger amount of time over the whole community, especially when you consider the effect of stress and annoyance from dealing with software maintenance surprises.

PR #430, opened 2017-07-01
Use `readdir` instead of deprecated `readdir_r` to avoid warnings

raphael-proust:
Fixed all C warning for unix on my machine. I'll have a look at the windows ones (although I'm not sure how much help I'd be there).

It passes tests but I don't trust myself and because it's C I don't trust the compiler to shout at me when I'm wrong. A second pair of eyes is definitely welcome.

I should also probably rebase to avoid a separate commit for cosmetics.

### DIFF
--- a/src/unix/lwt_unix_unix.h
+++ b/src/unix/lwt_unix_unix.h
@@ -1835,19 +1835,21 @@ static void worker_readdir_n(struct job_readdir_n *job)
 {
     long i;
     for (i = 0; i < job->count; i++) {
-        struct dirent *ptr;
-        int result = readdir_r(job->dir, &job->entries[i], &ptr);
+        errno = 0;
+        struct dirent *entry = readdir(job->dir);
 
         /* An error happened. */
-        if (result != 0) {
-            job->error_code = result;
+        if (entry == NULL && errno != 0) {
+            job->error_code = errno;
             return;
         }
 
         /* End of directory reached */
-        if (ptr == NULL) break;
-    }
+        if (entry == NULL && errno == 0) break;
 
+        /* All is good */
+        job->entries[i] = *entry;
+    }
     job->count = i;
     job->error_code = 0;
 }
--- a/src/unix/lwt_unix_unix.h
+++ b/src/unix/lwt_unix_unix.h
@@ -32,6 +32,7 @@
 
 #include <caml/unixsupport.h>
 #include <caml/version.h>
+#include <dirent.h>
 #include <poll.h>
 #include <sys/resource.h>
 #include <sys/time.h>
@@ -1777,30 +1778,35 @@ struct job_readdir {
     struct lwt_unix_job job;
     DIR *dir;
     struct dirent *entry;
-    struct dirent *ptr;
-    int result;
+    int error_code;
 };
 
 static void worker_readdir(struct job_readdir *job)
 {
-    job->entry = lwt_unix_malloc(dirent_size(job->dir));
-    job->result = readdir_r(job->dir, job->entry, &job->ptr);
+    // From the man page of readdir
+    // If the end of the directory stream is reached, NULL is returned and
+    // errno is not changed. If an error occurs, NULL is returned and errno
+    // is set appropriately. To distinguish end of stream and from an error,
+    // set errno to zero before calling readdir() and then check the value of
+    // errno if NULL is returned.
+    errno = 0;
+    job->entry = readdir(job->dir);
+    job->error_code = errno;
 }
 
 static value result_readdir(struct job_readdir *job)
 {
-    int result = job->result;
-    if (result) {
-        free(job->entry);
-        lwt_unix_free_job(&job->job);
-        unix_error(result, "readdir", Nothing);
-    } else if (job->ptr == NULL) {
-        free(job->entry);
+    LWT_UNIX_CHECK_JOB(job, job->entry == NULL && job->error_code != 0, "readdir");
+    if (job->entry == NULL) {
+        // From the man page
+        // On success, readdir() returns a pointer to a dirent structure.
+        // (This structure may be statically allocated; do not attempt to
+        // free(3) it.)
         lwt_unix_free_job(&job->job);
         caml_raise_end_of_file();
     } else {
         value name = caml_copy_string(job->entry->d_name);
-        free(job->entry);
+        // see above about not freeing dirent
         lwt_unix_free_job(&job->job);
         return name;
     }
--- a/src/unix/lwt_unix_unix.h
+++ b/src/unix/lwt_unix_unix.h
@@ -1811,6 +1811,13 @@ CAMLprim value lwt_unix_readdir_job(value val_dir)
 struct job_readdir_n {
     struct lwt_unix_job job;
     DIR *dir;
+    /* count serves two purpose:
+       1. Transmit the maximum number of entries requested by the programmer.
+          See `readdir_n`'s OCaml side documentation.
+       2. Transmit the number of actually read entries from the `worker` to
+          the result parser.
+          See below `worker_readdir_n` and `result_readdir_n`
+    */
     long count;
     int error_code;
     char *entries[];
--- a/src/unix/lwt_unix_unix.h
+++ b/src/unix/lwt_unix_unix.h
@@ -841,8 +841,6 @@ CAMLprim value lwt_unix_mcast_set_ttl(value fd, value ttl)
 #define VAL_MCAST_ACTION_ADD (Val_int(0))
 #define VAL_MCAST_ACTION_DROP (Val_int(1))
 
-#define GET_INET_ADDR(v) (*((struct in_addr *)(v)))
-
 CAMLprim value lwt_unix_mcast_modify_membership(value fd, value v_action,
                                                 value if_addr, value group_addr)
 {
@@ -2364,10 +2362,7 @@ struct job_gethostbyname {
 };
 
 CAMLexport value alloc_inet_addr(struct in_addr *inaddr);
-#define GET_INET_ADDR(v) (*((struct in_addr *)(v)))
-
 CAMLexport value alloc_inet6_addr(struct in6_addr *inaddr);
-#define GET_INET6_ADDR(v) (*((struct in6_addr *)(v)))
 
 static value alloc_one_addr(char const *a)
 {
--- a/src/unix/lwt_unix_unix.h
+++ b/src/unix/lwt_unix_unix.h
@@ -1849,7 +1849,7 @@ static void worker_readdir_n(struct job_readdir_n *job)
         }
 
         /* All is good */
-        job->entries[i] = strdup(entry->d_name);
+        job->entries[i] = name;
     }
     job->count = i;
     job->error_code = 0;
--- a/src/unix/lwt_unix_unix.h
+++ b/src/unix/lwt_unix_unix.h
@@ -1826,11 +1826,20 @@ static void worker_readdir_n(struct job_readdir_n *job)
         /* An error happened. */
         if (entry == NULL && errno != 0) {
             job->error_code = errno;
+            job->count = i - 1;
             return;
         }
 
         /* End of directory reached */
         if (entry == NULL && errno == 0) break;
+
+        /* readdir is good */
+        char *name = strdup(entry->d_name);
+        if (name == NULL) {
+            job->error_code = errno;
+            job->count = i - 1;
+            return;
+        }
 
         /* All is good */
         job->entries[i] = strdup(entry->d_name);
@@ -1845,6 +1854,8 @@ static value result_readdir_n(struct job_readdir_n *job)
     CAMLlocal1(result);
     int error_code = job->error_code;
     if (error_code) {
+        long i;
+        for (i = 0; i < job->count; i++) free(job->entries[i]);
         lwt_unix_free_job(&job->job);
         unix_error(error_code, "readdir", Nothing);
     } else {
@@ -1853,6 +1864,7 @@ static value result_readdir_n(struct job_readdir_n *job)
         for (i = 0; i < job->count; i++) {
             Store_field(result, i, caml_copy_string(job->entries[i]));
         }
+        for (i = 0; i < job->count; i++) free(job->entries[i]);
         lwt_unix_free_job(&job->job);
         CAMLreturn(result);
     }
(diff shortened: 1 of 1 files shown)


aantron:
It looks plausible, but due to past concerns about thread safety of `readdir`, we should probably read the man pages carefully and look in the source code of the standard libraries of at least Linux, macOS, FreeBSD. Links to Linux and macOS sources can be found [here](https://github.com/ocsigen/lwt/issues/342#issuecomment-295795683). I can do this later, as part of a thorough review.

Some extra testing may also be in order, but we can push that to a separate issue.

Also found this Rust issue, still open: https://github.com/rust-lang/rust/issues/34668

raphael-proust:
When researching this, I got  to https://lwn.net/Articles/696475/ which states:

> * In the current POSIX.1 specification (POSIX.1-2008), readdir(3) is
> not required to be thread-safe. However, in modern
> implementations (including the glibc implementation), concurrent
> calls to readdir(3) that specify different directory streams are
> thread-safe.

This raises the same concerns as the issue you linked to.

------------------------------------------------

On my linux, the man page states:

```
       ┌──────────┬───────────────┬──────────────────────────┐
       │Interface │ Attribute     │ Value                    │
       ├──────────┼───────────────┼──────────────────────────┤
       │readdir() │ Thread safety │ MT-Unsafe race:dirstream │
       └──────────┴───────────────┴──────────────────────────┘

       In the current POSIX.1 specification (POSIX.1-2008), readdir() is not required to be thread-safe.  However, in modern implementations (including the glibc implementation), concurrent calls to readdir() that specify
       different directory streams are thread-safe.  In cases where multiple threads must read from the same directory stream, using readdir() with external synchronization is still preferable to the use of the deprecated
       readdir_r(3) function.  It is expected that a future version of POSIX.1 will require that readdir() be thread-safe when concurrently employed on different directory streams.
```

aantron:
I found [this article](http://elliotth.blogspot.com/2012/10/how-not-to-use-readdirr3.html), which states that `readdir` is as thread-safe as we need it to be. It's from 2012, which is good – I was feeling uneasy that `readdir` might have only become thread-safe recently on some odd system.

Going to look around a bit more.

src/unix/lwt_unix_unix.h
```diff
@@ -1779,30 +1762,36 @@ struct job_readdir {
...
 static void worker_readdir(struct job_readdir *job)
 {
-    job->entry = lwt_unix_malloc(dirent_size(job->dir));
-    job->result = readdir_r(job->dir, job->entry, &job->ptr);
+    // From the man page of readdir
+    // If the end of the directory stream is reached, NULL is returned and
+    // errno is not changed. If an error occurs, NULL is returned and errno
+    // is set appropriately. To distinguish end of stream and from an error,
+    // set errno to zero before calling readdir() and then check the value of
+    // errno if NULL is returned.
```
aantron: +1 for including this comment!

src/unix/lwt_unix_unix.h
```diff
@@ -1779,30 +1762,36 @@ struct job_readdir {
...
-        unix_error(result, "readdir", Nothing);
-    } else if (job->ptr == NULL) {
-        free(job->entry);
+    LWT_UNIX_CHECK_JOB(job, job->entry == NULL && job->error_code != 0,
+                       "readdir");
+    if (job->entry == NULL) {
+        // From the man page
+        // On success, readdir() returns a pointer to a dirent structure.
+        // (This structure may be statically allocated; do not attempt to
+        // free(3) it.)
```
aantron: +2

src/unix/lwt_unix_unix.h
```diff
@@ -1872,7 +1863,7 @@ CAMLprim value lwt_unix_readdir_n_job(value val_dir, value val_count)
     long count = Long_val(val_count);
     DIR *dir = DIR_Val(val_dir);
 
-    LWT_UNIX_INIT_JOB(job, readdir_n, dirent_size(dir) * count);
+    LWT_UNIX_INIT_JOB(job, readdir_n, sizeof(struct dirent) * count);
```
aantron: It's good to get rid of the size estimation and whatnot. I believe this is not necessary with `readdir`, but see the other comment about copying out the strings.

For reference on how much of a pain this size estimation was, there is #292. Among other things, it fixed a buffer overflow bug in https://github.com/ocsigen/lwt/pull/292/commits/dbc42f26bd48873b3dbd864ab53e828d555f2111, where the janky `dirent` size computation recommended online could give a number smaller than what macOS could actually write, etc.

See also https://bugs.chromium.org/p/chromium/issues/detail?id=457759.

I sometimes wonder how anything in the world works, with such OS API issues :p Then again, that's why we have 0-days all over the place.

src/unix/lwt_unix_unix.h
```diff
@@ -1831,19 +1820,21 @@ static void worker_readdir_n(struct job_readdir_n *job)
...
             return;
         }
 
         /* End of directory reached */
-        if (ptr == NULL) break;
-    }
+        if (entry == NULL && errno == 0) break;
 
+        /* All is good */
+        job->entries[i] = *entry;
```
aantron: I think this is incorrect as a way to copy the filename, because the `d_name` field is not specified to have any particular size; on a platform where it is defined as `d_name[]`, copying 
the structure will not copy the "true" contents of that field. From [Linux `readdir(3)`](http://man7.org/linux/man-pages/man3/readdir.3.html):

> Warning: applications should avoid any dependence on the size of the
       d_name field.  POSIX defines it as char d_name[], a character array
       of unspecified size, with at most NAME_MAX characters preceding the
       terminating null byte ('\0').

 I think we need to allocate a string and do `strcpy` (or just call `strdup`), and replace the flat array of `struct dirent`s with an array of pointers to these copied strings. Of course, they have to be freed in the `_result` function, or, upon failure, in the worker, if that is more convenient.

Unfortunately, this also means two copies of the string on the Lwt side, one into this array of ours, and one as part of `caml_copy_string`. I doubt this is a big deal. It's not necessary, but for extra thoroughness it's good to run the benchmarks [linked here](https://github.com/ocsigen/lwt/pull/292#issue-190602425); I can also do it on macOS on an updated version of this code.

aantron:
So replacing `readdir_r` by `readdir` seems like the right way on all major Unix platforms (Linux, macOS, FreeBSD, Android, iOS). See inline comments for small remaining issues.

Rebasing also is a good idea, but not necessary – I think we will do a squash-merge at the end of this PR, and that should take care of that first commit being there.

src/unix/lwt_unix_unix.h
raphael-proust: I almost want to use`caml_copy_string` there. It sort of breaks the pattern of “`worker` does the C/system part and `result` does the conversion.” And on top of that, I'm not sure what happens if you call `caml_copy_string` and then abandon those values because of an error. Do you need to free them?

Actually… see other comment.

I see little point in copying the string twice: once in C and then once to convert to OCaml layout. Any opinions about it? Any comments on the paragraph above?

aantron: Off the top of my head, I think we can't call `caml_copy_string` from a fully-preemptive worker thread that does not play nice with OCaml's global runtime lock. Should probably also clarify that in the recent `getcwd` docs.

> Actually… see other comment.

I hope I didn't miss another comment :)

> I see little point in copying the string twice: once in C and then once to convert to OCaml layout. Any opinions about it? Any comments on the paragraph above?

I also see little point, but we may be forced to by the constraints of the runtime system :/ Ultimately, I think the cost of switching to the worker thread and back will tend to dominate the cost of copying bytes around.

raphael-proust: No other comment. I was worried about something else and then whilst writing the comment I realised that it wasn't a problem.

src/unix/lwt_unix_unix.h
aantron: This should now probably be `sizeof(char*)`.

src/unix/lwt_unix_unix.h
```diff
@@ -1860,7 +1851,7 @@ static value result_readdir_n(struct job_readdir_n *job)
         result = caml_alloc(job->count, 0);
         long i;
         for (i = 0; i < job->count; i++) {
-            Store_field(result, i, caml_copy_string(job->entries[i].d_name));
+            Store_field(result, i, caml_copy_string(job->entries[i]));
```
aantron: We should also free the `entries` that actually got allocated, both if the job fails and if it succeeds. For that, we need a valid `job->count` in all cases, but we don't have one in case of an error from `readdir` in the worker (because of the hard `return` in it, above).

src/unix/lwt_unix_unix.h
```diff
@@ -1824,26 +1813,28 @@ struct job_readdir_n {
...
             return;
         }
 
         /* End of directory reached */
-        if (ptr == NULL) break;
-    }
+        if (entry == NULL && errno == 0) break;
 
+        /* All is good */
+        job->entries[i] = strdup(entry->d_name);
```
aantron: Should check the result of this for `NULL`, and probably set a custom error code (`ENOMEM`?) and break out of the loop to ensure a valid count of entries previously allocated successfully. See other comments on the count and `free`.

src/unix/lwt_unix_unix.h
raphael-proust: That's the thing I had doubts about. But then, the previous code did not try to deallocate any of the `dirent`s. Was there a memory leak before? What's the difference in the code that makes that change necessary?

aantron: IIRC: the code before this PR allocated a flat array of `dirent`s, i.e. if it calculated that a `dirent` was 1024 bytes and needed to allocate 32 of them, it would allocate 32KB of space in one block. The strings were written directly into various parts of this single block. We controlled the maximum length of these strings when calling `readdir_r`.

The new code inserts a level of indirection, because we don't know how long the strings returned by `readdir` will be. So, for reading 32 entries, it is now allocating space for 32 8-byte pointers in one block, and each one of those pointers may then be set to the address of a separately-allocated string (by `strdup`). We need to free these separate string blocks.

src/unix/lwt_unix_unix.h
```diff
@@ -1826,12 +1826,21 @@ static void worker_readdir_n(struct job_readdir_n *job)
...
             return;
         }
 
         /* End of directory reached */
         if (entry == NULL && errno == 0) break;
 
+        /* readdir is good */
+        char *name = strdup(entry->d_name);
+        if (name == NULL) {
+            job->error_code = errno;
```
aantron: Good call here, I didn't even know that `strdup` actually sets `errno` (to `ENOMEM`) :p

src/unix/lwt_unix_unix.h
```diff
@@ -1811,6 +1811,13 @@ CAMLprim value lwt_unix_readdir_job(value val_dir)
 struct job_readdir_n {
     struct lwt_unix_job job;
     DIR *dir;
+    /* count serves two purpose:
+       1. Transmit the maximum number of entries requested by the programmer.
+          See `readdir_n`'s OCaml side documentation.
+       2. Transmit the number of actually read entries from the `worker` to
+          the result parser.
+          See below `worker_readdir_n` and `result_readdir_n`
+    */
```
aantron: Definitely :+1:

src/unix/lwt_unix_unix.h
```diff
@@ -1822,28 +1811,46 @@ CAMLprim value lwt_unix_readdir_job(value val_dir)
...
-        int result = readdir_r(job->dir, &job->entries[i], &ptr);
+        errno = 0;
+        struct dirent *entry = readdir(job->dir);
 
         /* An error happened. */
-        if (result != 0) {
-            job->error_code = result;
+        if (entry == NULL && errno != 0) {
+            job->error_code = errno;
+            job->count = i - 1;
```
aantron: I believe `job->count` should be set to `i`: if this fails on the first iteration, `i` is 0; 0 entries were read. If it fails on the second iteration, `i` is 1; 1 entry was read, etc.

aantron:
I hate that big red X. This is my first time using reviews with Approve/Request changes, the X looks like rejected...

src/unix/lwt_unix_unix.h
raphael-proust: Oh yes. The stop condition when freeing is `i < job->count` so it should definitely be `i`. Well spotted.

src/unix/lwt_unix_unix.h
raphael-proust: Oh. That makes sense.

src/unix/lwt_unix_unix.h
```diff
@@ -3199,7 +3199,12 @@ CAMLprim value lwt_unix_getcwd_job(value unit)
        The last argument is the number of bytes of storage to reserve in memory
        immediately following the `struct`. This is for fields such as
        `char data[]` at the end of the struct. It is typically zero. For an
-       example where it is not zero, see `lwt_unix_read_job` again. */
+       example where it is not zero, see `lwt_unix_read_job` again.
+
+       If the additional data is stored inline in the job struct, it is
+       deallocated with `lwt_unix_free_job`. If the additional data is for
+       pointers to additional structure, you must remember to deallocate it
+       yourself. For an example of this, see `readdir_n`.*/
```
aantron: :+1: Thanks, this is one less task in my local TODO list :)

src/unix/lwt_unix_unix.h
```diff
@@ -1822,28 +1811,46 @@ CAMLprim value lwt_unix_readdir_job(value val_dir)
...
+        /* readdir is good */
+        char *name = strdup(entry->d_name);
+        if (name == NULL) {
+            job->count = i;
+            job->error_code = errno;
+            return;
+        }
 
+        /* All is good */
+        job->entries[i] = strdup(entry->d_name);
```
aantron: I think you meant to assign `name` here, not call `strdup` again.

raphael-proust: Good catch!

raphael-proust:
Let me know if it's all good to go and I can squash the whole thing (or rebase it to look better).

aantron:
It should be good now, just waiting for the build. There's no need for you to squash it, unless you want to. I can use GitHub's squash-merge. Let me know if you want me to wait and let you doctor the history instead of me :p

aantron:
Thank you!

PR #433, opened 2017-07-04
Fix Lwt_unix.timeout/Lwt.sleep returning early

### LINKED ISSUE #427: Lwt_unix.timeout may return early
We are observing that Lwt_unix.timeout can sometimes return early.

The case that we were able to debug a bit more closer was doing a very lengthy writeout to the console (multiple megabytes), from an Lwt task. The next operation with an Lwt_unix.timeout raised Lwt_unix.Timeout early. Suppressing the writeout gave normal behavior.

Digging a bit deeper, I believe the following behavior in libev is causing this:
>The special problem of time updates
>
>Establishing the current time is a costly operation (it usually takes at least two system calls): EV
>therefore updates its idea of the current time only before and after "ev_run" collects new events, which 
>causes a growing difference between "ev_now ()" and "ev_time ()" when handling lots of events in one 
>iteration.
>
>The relative timeouts are calculated relative to the "ev_now ()" time. This is usually the right thing as this 
>timestamp refers to the time of the event triggering whatever timeout you are modifying/starting. If 
>you suspect event processing to be delayed and you need to base the timeout on the current time, use 
>something like this to adjust for this:
>
>ev_timer_set (&timer, after + ev_now () - ev_time (), 0.);
>
>If the event loop is suspended for a long time, you can also force an update of the time returned by 
>"ev_now ()" by calling "ev_now_update ()". 

See also [man ev](https://linux.die.net/man/3/ev)

When an Lwt.t already has a value or exception, a bind will simply continue execution. This may, in the absence of I/O, cause lengthy periods of not going back into the libev event loop, causing the ev_now time-base to become stale. The next Lwt_engine.on_timer call will then have a relative timeout based on the old time-base, making the timeout shorter than expected.

Reproducer usable in the Lwt testsuite:
```
  test "libev"
    (fun () ->
       let start = Unix.gettimeofday () in
       let stop = ref start in
       let t1 = catch (fun () ->
          let () = Unix.sleep 3 in
          Lwt_unix.timeout 1.;
         )
           (function
            | Lwt_unix.Timeout ->
              let () = stop := Unix.gettimeofday () in
              return ()
            | exn -> Lwt.fail exn
           )
       in
       bind t1 (fun () ->
           let () = Printf.printf "start = %f, stop= %f, delta=%f\n"
               start
               (!stop)
               ((!stop) -. start)
           in
           return ((!stop) -. start >= 4.))
    );
```
This traces:
start = 1498665441.874718, stop= 1498665444.876054, delta=3.001336

A potential fix could be this:
```
diff --git a/src/unix/lwt_libev_stubs.c b/src/unix/lwt_libev_stubs.c
index 66ccb4a..57ef001 100644
--- a/src/unix/lwt_libev_stubs.c
+++ b/src/unix/lwt_libev_stubs.c
@@ -191,6 +191,7 @@ CAMLprim value lwt_libev_timer_init(value loop, value delay, value repeat, value
 {
   CAMLparam4(loop, delay, repeat, callback);
   CAMLlocal1(result);
+  ev_now_update(Ev_loop_val(loop));
   /* Create and initialise the watcher */
   struct ev_timer* watcher = lwt_unix_new(struct ev_timer);
   if (Bool_val(repeat))
```
(Beware this is against a 2.5.0 version of Lwt).

stijn-devriendt:
These patches fix #427 and adds a test for it.
As mentioned in the issue discussion this patch should be fairly low overhead compared to doing a full-fledged ev_now_update().

### DIFF
--- a/src/unix/lwt_libev_stubs.c
+++ b/src/unix/lwt_libev_stubs.c
@@ -200,12 +200,14 @@ CAMLprim value lwt_libev_timer_init(value loop, value delay, value repeat,
                                     value callback) {
   CAMLparam4(loop, delay, repeat, callback);
   CAMLlocal1(result);
+
+  struct ev_loop* ev_loop = Ev_loop_val(loop);
   /* Create and initialise the watcher */
   struct ev_timer *watcher = lwt_unix_new(struct ev_timer);
   if (Bool_val(repeat))
-    ev_timer_init(watcher, handle_timer, Double_val(delay), Double_val(delay));
+    ev_timer_init(watcher, handle_timer, Double_val(delay) + ev_time() - ev_now(ev_loop), Double_val(delay));
   else
-    ev_timer_init(watcher, handle_timer, Double_val(delay), 0.0);
+    ev_timer_init(watcher, handle_timer, Double_val(delay) + ev_time() - ev_now(ev_loop), 0.0);
 
   /* Wrap the watcher into a custom caml value */
   result = caml_alloc_custom(&watcher_ops, sizeof(struct ev_timer *), 0, 1);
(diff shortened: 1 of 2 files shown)


aantron:
Thanks for including the test. It made reproducing on my system, to check, pretty easy.

PR #439, opened 2017-07-08
Testing: be sure we test thoroughly on Flambda

### LINKED ISSUE #436: Be more thorough about testing on Flambda
The old `lwt.ml` used `Obj.magic`, and the new `lwt.ml` (#354) uses it even more.

In the meantime, I realized that we might not actually be passing an optimization flag to the compiler in the Flambda build row in Travis. We should make sure to do that. I've done it manually a few times on one compiler, but that's obviously not suitable going forward.

We should also start testing on compilers before they reach beta, i.e. against `trunk`. We want to know about any potential problems as early as possible.

This was prompted by @bluddy on Discourse; see https://discuss.ocaml.org/t/lwt-core-refactored-and-documented-to-be-contributor-friendly/161/17 and subsequent comments.

aantron:
Passes `-O3` to the compiler when there is Flambda support. Before this PR, our Flambda builds were using an Flambda compiler, but perhaps the optimizations applied were [not aggressive enough](https://caml.inria.fr/pub/docs/manual-ocaml/flambda.html#sec478) for catching any potential problems.

Lwt [uses some `Obj.magic`][objmagic], so it's important we test thoroughly on Flambda, including on development versions of the compiler. In addition to `-O3`, this PR adds CI testing on 4.06.0+trunk+flambda.

`-O3` is passed to the compiler when the version is at least 4.03. These compilers accept the flag even if not configured with Flambda. At the same time, we have to avoid passing `-O3` to 4.02.3. That compiler exits with an error if given the flag. This conditional flag-passing is achieved by writing even more OCaml code in `src/core/jbuild` (see the diff). @diml, is there any nice way to avoid that in current Jbuilder?

I'd like to avoid turning all of our `jbuild` files into OCaml source, in part because we would have a lot of duplicate code in them. I guess if we have to go down that route with the `jbuild` files, we will have to write a separate generating pass. To limit the number of `jbuild` files turned to OCaml syntax in this PR, I added `-O3` to only the core (`src/core/*`) and the core tests (`test/core/*`). This should be enough for rudimentary testing of Flambda during compilation of `lwt.ml`, and during cross-module optimization when the test suite is compiled against `lwt.cmx`.

I confirmed that `-O3` is being passed, and working, by running `jbuilder --verbose` manually, and also by temporarily passing `-inlining-report` and checking the generated `*.inlining.*` files.

We don't seem to be having any problems from Flambda at the moment – at least none triggered by our tests :)

Resolves #436.

cc @bluddy (thanks for getting this rolling)

[objmagic]: https://discuss.ocaml.org/t/lwt-core-refactored-and-documented-to-be-contributor-friendly/161/24

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,9 @@ matrix:
     - os: linux
       env: COMPILER=4.04
     - os: linux
-      env: COMPILER=4.05
-    - os: linux
       env: COMPILER=4.05 FLAMBDA=yes
+    - os: linux
+      env: COMPILER=4.06 FLAMBDA=yes
     - os: osx
       env: COMPILER=system
     - os: osx
--- a/src/core/jbuild
+++ b/src/core/jbuild
@@ -16,10 +16,20 @@ let optional_bisect_stanza = {|
   (preprocess (pps (bisect_ppx)))
 |}
 
+let optional_flambda_flags = {|
+  (ocamlopt_flags (:standard -O3))
+|}
+
 let bottom = {|
   (libraries (bytes result))
   (flags (:standard -w +A-29))))
 |}
+
+let accumulate_content_if condition more_content content =
+  if condition then
+    content ^ more_content
+  else
+    content
 
 let () =
   let generating_coverage =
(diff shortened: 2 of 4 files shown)


diml:
It's not possible to do that without using the OCaml syntax at the moment, however https://github.com/janestreet/jbuilder/pull/153 allows to use `(:include ..,.)` with `ocamlopt_flags`

aantron:
It's a definite improvement. It's still slightly awkward, though. For a project with many `jbuild` files, the `:include` has to be done in each one. It's much better than converting all files to OCaml syntax, though.

PR #440, opened 2017-07-10
Issue #385: Increase test coverage for Lwt_result

jsthomas:
This change adds tests to bring coverage for `Lwt_result` up to 100%.

### DIFF
--- a/tests/core/main.ml
+++ b/tests/core/main.ml
@@ -25,4 +25,5 @@ Test.run "core" [
   Test_lwt_list.suite;
   Test_lwt_switch.suite;
   Test_lwt_mutex.suite;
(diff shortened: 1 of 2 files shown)


aantron:
Don't worry about the failed CI builds. We will remove the `working-coverage` branch after this PR, since #438 makes it unnecessary for the Lwt core, and we can just use `master`. The fatal warning messing up CI is fixed in `master` by 4557ea18f17fcee8912cd60f9fe43d2b38602f4a.

aantron:
Thanks! I cherry-picked this into `master` as 99f556b347567f92a1b290915881692b94b060b4. Also want to add a link to #385 (and from it).

PR #444, opened 2017-07-12
Improve `Lwt.finalize` documentation

### LINKED ISSUE #442: [easy] Clarify Lwt.finalize documentation
From https://discuss.ocaml.org/t/lwt-finalize-and-failure-in-the-cleanup-code/525:

> The documentation for Lwt.finalize states:
>
> > `finalize f g` returns the same result as `f ()` whether it fails or not. In both cases, `g ()` is executed after `f`.
>
> Yet the following example demonstrates that this is not the case:
>
> ```ocaml
> Lwt.finalize
>   (fun () -> Lwt.fail (Failure "1"))
>   (fun () -> Lwt.fail (Failure "2"))
> |> state;;
> ```
>
> This gives a state with `Failure "2"`, not the `Failure "1"` so the failure comes from the execution of g. Is it a bug in the documentation?

Indeed, if `g ()` fails, that behavior controls the behavior of `finalize`, no matter the outcome of `f ()`. The documentation does not mention this, and states unconditionally that `finalize` returns the same result as `f ()`.

The documentation is found here: https://github.com/ocsigen/lwt/blob/99f556b347567f92a1b290915881692b94b060b4/src/core/lwt.mli#L164-L166

Richard-Degenne:
Fixes #442.

### DIFF
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -162,8 +162,10 @@ val try_bind : (unit -> 'a t) -> ('a -> 'b t) -> (exn -> 'b t) -> 'b t
       exception associated to [t ()]. *)
 
 val finalize : (unit -> 'a t) -> (unit -> unit t) -> 'a t
-  (** [finalize f g] returns the same result as [f ()] whether it
-      fails or not. In both cases, [g ()] is executed after [f]. *)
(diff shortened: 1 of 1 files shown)


aantron:
Thank you!

PR #447, opened 2017-07-14
Npick fix

### LINKED ISSUE #340: Lwt.npick is broken

ZackC:
#340 - removed the line that causes a list of pending promises to be canceled.

### DIFF
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -2534,7 +2534,6 @@ struct
         let callback in_completion_loop _result =
           let State_may_now_be_pending_proxy p = may_now_be_proxy p in
           let p = underlying p in
-          List.iter cancel ps;
           let State_may_have_changed p =
             finish_nchoose_or_npick_after_pending in_completion_loop p [] ps in
           ignore p
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -2418,34 +2418,32 @@ struct
   (* If [nchoose ps] or [npick ps] found all promises in [ps] pending, the
      callback added to each promise in [ps] eventually calls this function. The
      function collects promises in [ps] that have resolved, or finds one promise
-     in [ps] that has failed. It then completes [to_complete], the result
-     promise of [nchoose]/[npick], with the list of resolved promises, or the
-     state of the failed promise. *)
-  let rec finish_nchoose_or_npick_after_pending
-      in_completion_loop
-      (to_complete : ('a list, underlying, pending) promise)
+     in [ps] that has failed. The resolved promises are not completed to allow 
+     the unresolved promises to be cancelled first (making npick's behavior 
+     consistent with pick's behavior) *)
+     
+  let rec collect_resolved_nchoose_or_npick_after_pending
       (results : 'a list)
-      (ps : 'a t list) :
-        ('a list, underlying, completed) state_changed =
+      (ps : 'a t list) =
 
     match ps with
     | [] ->
-      complete in_completion_loop to_complete (Resolved (List.rev results))
+      (Resolved (List.rev results))
 
     | p::ps ->
       let Internal p = to_internal_promise p in
 
       match (underlying p).state with
       | Resolved v ->
-        finish_nchoose_or_npick_after_pending
-          in_completion_loop to_complete (v::results) ps
+        collect_resolved_nchoose_or_npick_after_pending
+          (v::results) ps
 
       | Failed _ as result ->
-        complete in_completion_loop to_complete result
+        result
 
       | Pending _ ->
-        finish_nchoose_or_npick_after_pending
-          in_completion_loop to_complete results ps
+        collect_resolved_nchoose_or_npick_after_pending
+          results ps
 
   let nchoose ps =
     (* If at least one promise in [ps] is found resolved, this function is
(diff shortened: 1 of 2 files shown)


ZackC:
From my understanding the build checks fail because of these tests in ./test/core/test_lwt.ml:   

(* The behavior of [p] tested here is a current bug in [Lwt.npick]. *)
  test "npick: pending, resolves" begin fun () ->
    let p1, _ = Lwt.task () in
    let p2, r = Lwt.task () in
    let p = Lwt.npick [p1; p2] in
    assert (Lwt.state p = Lwt.Sleep);
    Lwt.wakeup r "foo";
    Lwt.return
      (Lwt.state p1 = Lwt.Fail Lwt.Canceled &&
       Lwt.state p = Lwt.Fail Lwt.Canceled)
  end;

  (* This is the same bug as above. *)
  test "npick: pending, fails" begin fun () ->
    let p1, _ = Lwt.task () in
    let p2, r = Lwt.task () in
    let p = Lwt.npick [p1; p2] in
    Lwt.wakeup_exn r Exception;
    Lwt.return
      (Lwt.state p1 = Lwt.Fail Lwt.Canceled &&
       Lwt.state p = Lwt.Fail Lwt.Canceled)
  end;

ZackC:
New commit that moves the cancellation of promises to after some of them have the chance to complete.

ZackC:
Huh, the build produced an unable to install parsetree error.  That shouldn't be related to the changes I made.

aantron:
Indeed it's not, see https://github.com/let-def/ocaml-migrate-parsetree/issues/19#issuecomment-317189028. We can just ignore that build row for now; I'll list in `allowed_failures:` later.

src/core/lwt.ml
```diff
@@ -2534,9 +2534,9 @@ struct
         let callback in_completion_loop _result =
           let State_may_now_be_pending_proxy p = may_now_be_proxy p in
           let p = underlying p in
-          List.iter cancel ps;
           let State_may_have_changed p =
             finish_nchoose_or_npick_after_pending in_completion_loop p [] ps in
+          List.iter cancel ps;
```
aantron: I believe this reordering produces the correct behavior.

There is one weird thing about Lwt now, though:

- The [callback of `Lwt.pick`](https://github.com/ocsigen/lwt/blob/424e4c50ef84d359bf0389e9fdb37ef66cbd1c3f/src/core/lwt.ml#L2398-L2404) cancels the pending promises *before* completing its own ("final") promise.
- This modified callback of `Lwt.npick` cancels the pending promises *after* completing its own promise.

This means that code running in `on_cancel` and regular callbacks of the respective pending promises can observe different states of the final promise depending on which function is used. Likewise, code running in the callbacks of the final promises can observe different states of the (to-be) canceled promises, depending on which function is used.

I'm not sure if that's really a problem, since we don't document or guarantee either behavior, and I'm not aware of any pathological interactions due to picking one or the other, so one could argue that this doesn't need to be addressed right now.

However, it could cause problems for users refactoring their code. For completeness, this could be addressed by factoring out the [implicit final promise completion](https://github.com/ocsigen/lwt/blob/424e4c50ef84d359bf0389e9fdb37ef66cbd1c3f/src/core/lwt.ml#L2433) from `finish_nchoose_or_npick_after_pending`, so that function only collects the result, and the [callback](https://github.com/ocsigen/lwt/blob/424e4c50ef84d359bf0389e9fdb37ef66cbd1c3f/src/core/lwt.ml#L2534-L2541) can fully control the order of (1) figuring out the result (2) canceling the pending promises (3) setting the state of the final promise (the one returned by `npick` itself). Callbacks in both `npick` and `nchoose` will have to be modified to call `complete` explicitly.

test/core/test_lwt.ml
```diff
@@ -2361,7 +2374,7 @@ let npick_tests = [
        Lwt.state p = Lwt.Fail Lwt.Canceled)
   end;
 
-  (* This is the same bug as above. *)
```
aantron: This is fine, but FYI OCaml supports nested comments, so this can be reverted.

test/core/test_lwt.ml
```diff
@@ -2349,7 +2349,20 @@ let npick_tests = [
...
+  test "npick: all pending" begin fun () ->
+    let p1, r = Lwt.task () in
+    let p2, _ = Lwt.task () in
+    let p3 = Lwt.npick [p1; p2] in
+    Lwt.wakeup r ();
+    (* Expected: *)
+    Lwt.return (Lwt.state p3 = Lwt.Return [()])
+    (* Actual: *)
+    (*Lwt.return (Lwt.state p3 = Lwt.Fail Lwt.Canceled)*)
+  end;
```
aantron: Thanks for inserting this here!

src/core/lwt.ml
aantron: A secondary reason to prefer the ordering in `pick` is that function gets used a lot, so presumably that behavior is not problematic.

aantron:
@ZackC I wrote a small program to show the two behaviors, in part for my own education, but it may also be helpful for ensuring same ordering: https://gist.github.com/aantron/4db12b8e5ca92f57f7a223e594c833cc

aantron:
Thanks for fixing this!

ZackC:
No problem. Glad to help!

PR #450, opened 2017-07-18
Issue #385: Increase coverage for Lwt_condition

jsthomas:
As part of #385, this change increases unit test coverage of `Lwt_condition` to 100%.


aantron:
Excellent, thanks!

Issue #449, opened 2017-07-16
[easy] Delete Camlp4 from the manual introduction


aantron:
The Camlp4 syntax is deprecated, but we are still confusing users by [talking about it (at length) on the manual intro page (link; section "The syntax extension")](https://ocsigen.org/lwt/manual/). We should:

- Delete everything in the Camlp4 subsection of this.
- Either delete the Correspondence table for now, or adapt it to the PPX.
- Get rid of the Ppx subsection header – it's not necessary when there is no Camlp4.
- Get rid of the reference to "two" syntax extensions.

bobbypriambodo:
Hi Anton, I'd love to take this on to learn more about the ppx syntax. Questions, though:

1) Do you wish to keep the ppx syntax documentation [separated](https://ocsigen.org/lwt/3.0.0/api/Ppx_lwt)? If so, I would think keeping the correspondence table doesn't make sense in the manual (perhaps move it to the ppx page?).
2) I suppose we should add a note about the deprecation of camlp4?

aantron:
Thanks for looking at this.

1. I'm not sure yet. Ultimately, all the docs will get [rearranged](https://github.com/ocsigen/lwt/projects/2). However, we need good source material *somewhere*, and the table definitely seems worthwhile to have. I'd like to leave it up to you :)
2. The Camlp4 syntax is already [marked as deprecated](https://ocsigen.org/lwt/dev/api/Pa_lwt) on its own page, if that's what you mean (as opposed to Camlp4 itself). I think it's best to avoid mentioning it all elsewhere, to avoid drawing attention to it. I think we should act like it doesn't exist, but still retain its own docs for code that still uses it, and for readers that encounter code still using it.

bobbypriambodo:
I see. Yes that's what I meant. I think that makes sense.

Okay then, let me see what I can come up with. 🙂

PR #457, opened 2017-07-22
Remove camlp4 reference from the manual

bobbypriambodo:
Solves #449.

Removes camlp4 reference, since its use has been deprecated. The "Correspondence table" is also updated to use the Ppx syntax.

### DIFF
--- a/doc/manual.wiki
+++ b/doc/manual.wiki
@@ -684,30 +684,9 @@ val s' : int Lwt_stream.t = <abstr>
 ;;
 >>
 
-  {{{Lwt}}} also provides a syntax extension, in the package
-  {{{lwt.syntax.log}}}. It does not modify the language but
-  it replaces log statements of the form:
-
-<<code language="ocaml" |Lwt_log.info_f ~section "something happened: %s" msg
->>
-
-  by:
-
-<<code language="ocaml" |if Lwt_log.Section.level section <= Lwt_log.Info then
-  Lwt_log.info_f ~section "something happened: %s" msg
-else
-  Lwt.return ()
->>
-
-  The advantages of using the syntax extension are the following:
-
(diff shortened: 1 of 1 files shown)


bobbypriambodo:
There is still one reference left of camlp4 in [The logging facility](https://github.com/bobbypriambodo/lwt/blob/remove-camlp4-reference/doc/manual.wiki#L676-L699) section. I'm not sure what to do with it, should I just delete it altogether?

aantron:
Probably the best is to replace it by some kind of summary of, or reference to, the PPX logging support section https://ocsigen.org/lwt/3.0.0/api/Ppx_lwt#2_Logging.

aantron:
It looks good. The only thing I'd recommend is adding an example of a command line for triggering the PPX, i.e. with `ocamlfind ... -package lwt.ppx ...` (hope I didn't miss it if it's already there). We had the equivalent for Camlp4.

aantron:
And don't worry about the 4.06+trunk build failure – that's clearly an upstream problem. I'll move it to `allowed_failures` later in `master`.

aantron:
(And if you'd like to see more details on that upstream issue, https://github.com/let-def/ocaml-migrate-parsetree/issues/19#issuecomment-317189028).

bobbypriambodo:
I've updated the PR with the example to use `lwt.ppx` on `ocamlfind` and toplevel, and a reference to `Ppx_lwt` Logging section.

Sorry for the late commits, timezone issues 🙂

aantron:
Looks good. Thanks!

aantron:
Checked the generated version online (it's hard to do so ahead of time). It looks good as well.

bobbypriambodo:
That's good to hear, and thanks for letting me take this!

PR #460, opened 2017-07-29
cppo 1.1.0 introduced -V option

rgrinberg:
minimum bounds should reflect that

### DIFF
--- a/lwt.opam
+++ b/lwt.opam
@@ -24,7 +24,7 @@ build: [
 build-test: [ [ "jbuilder" "runtest" "-p" name ] ]
 
 depends: [
-  "cppo" {build}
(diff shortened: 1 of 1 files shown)


aantron:
Thanks. Is this affecting any code in the wild? Do we need to update already-released `opam` files?

rgrinberg:
It has affected me which is why I made this PR. But it's probably unlikely
to affect many users as it happens only when opam decides on some weird
plans by picking a very old cppo

rgrinberg:
Oh and I forgot to add that it definitely won't hurt to add this constraint
to old versions of lwt but it's kind of painful to do with opam-repository
so I will leave that to the experts.

PR #461, opened 2017-08-01
Bugfix: Adjust validation behavior of Lwt_pool.use

jsthomas:
This change adjusts the behavior of `Lwt_pool.use` to match the [manual's specifications](https://ocsigen.org/lwt/dev/api/Lwt_pool), per some discussion in Issue #385. 

Prior to this change, a pool element would not be validated if it was returned to the pool with threads waiting. After this change, `validate` is called whenever `Lwt_pool.use` releases an existing pool element (though we still assume that it is not necessary to validate elements that were newly created for the pool). I've included a single unit test that exercises this fix.

cc: @hcarty

### DIFF
--- a/src/core/lwt_pool.ml
+++ b/src/core/lwt_pool.ml
@@ -92,19 +92,10 @@ let replace_acquired p =
          p.count <- p.count - 1;
          Lwt.wakeup_later_exn wakener exn)
 
-let acquire p =
-  if Queue.is_empty p.list then
-    (* No more available member. *)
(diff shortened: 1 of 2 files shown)


hcarty:
Thank you for the fix!

aantron:
Excellent, thanks for fixing that. Reading `Lwt_pool` in more detail just keeps making me more and more worried :p Thanks for including that test. It made reviewing this much easier than it would otherwise have been.

PR #462, opened 2017-08-02
Add a license file

samoht:


samoht:
This is a tentative license file, as it seems that the opam metadata is a bit different from the header files. Not sure what the "openSSL" linking exception is.

aantron:
Lwt has two license files, [`doc/LICENSE`](https://github.com/ocsigen/lwt/blob/cfd04ecdfc3b259d2bca728a452f1bf1553023b7/doc/LICENSE) with a summary and [`doc/COPYING`](https://github.com/ocsigen/lwt/blob/cfd04ecdfc3b259d2bca728a452f1bf1553023b7/doc/COPYING) with the full text.

I'm also not sure what the OpenSSL linking exception really is, but `COPYING` includes this text:

```
the additional exemption that compiling, linking, and/or using OpenSSL is
allowed.
```

I don't know if it's actually relevant to how Lwt is used.

Lwt is weird. Most of it is LGPL, there is the exception, and some files are under the BSD license. Perhaps we should start a discussion at some point about simplifying the licensing. For comparison, there is also the relatively recent discussion about this in Bisect_ppx: https://github.com/aantron/bisect_ppx/issues/133.

@Drup, can you cc the right persons from Ocsigen for license issues?

Drup:
As far as ocsigen is concerned, cc @balat 
I think big contributors _used to_ sign a piece of paper giving licensing rights to INRIA (I did), but it's not really done anymore. I'm not sure if you will need to ping all the contributors ever (*gasp*) to change the license.

samoht:
Ha my license scrapping script and I haven't found the doc/COPYING file, so I will use this :-)

Feel free to close that PR then. Having some clarification on the choosen license would be useful, although it seems that the exception is due to openSSL itself: https://en.wikipedia.org/wiki/OpenSSL#Licensing

aantron:
That text seems much more relevant to GPL rather than LGPL, and also, now that we have a separate `lwt_ssl`, maybe only `lwt_ssl` needs a weird license. I'll leave this PR open while thinking about it, and maybe we will get some additional informed comments as well :)

aantron:
Closing this *as a PR*. It might be worthwhile to open a separate issue about the license later.

PR #464, opened 2017-08-04
Issue #385: Increase test coverage for Lwt_pool.

jsthomas:
This change increases test coverage for `Lwt_pool` to 100%; this is the final subtask in issue #385.


aantron:
Excellent, thank you!

For future readers (including myself), I want to note that what looks like the deleted test actually became `"validation exceptions are propagated to users"`.

Issue #336, opened 2017-04-10
Temporary file creation


copy:
The following functions would be useful:

```ocaml
val open_temp_file : 
  ?buffer:Lwt_bytes.t ->
  ?perm:Unix.file_perm -> unit -> output channel Lwt.t

val with_temp_file : 
  ?buffer:Lwt_bytes.t ->
  ?perm:Unix.file_perm -> (output channel -> 'b Lwt.t) -> 'b Lwt.t
```

In particular, `with_temp_file` should safely delete the file when the callback returns or fails.
Inspired by [open_temp_file](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Filename.html#VALopen_temp_file) and [mkstemp](http://man7.org/linux/man-pages/man3/mkstemp.3.html).

aantron:
Agreed in principle. I've written such Lwt helpers in my own projects as well.

jsthomas:
I'm working on this issue.

PR #467, opened 2017-08-26
Issue #336: Add temporary file creation capabilities to Lwt_io

jsthomas:
This change adds two new functions, `open_temp_file` and `with_temp_file` to `Lwt_io`, as was requested in #336.

@aantron, @copy: Please let me know if this was what you had in mind when #336 was opened/added to the milestone; I haven't worked in this module before, so I'm not sure if I'm breaking any important conventions.

### DIFF
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1354,6 +1354,29 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
     (fun () -> f ic)
     (fun () -> close ic)
 
+let temp_filename () =
+  let rec helper n =
+    let f = "lwt_io_temp_file_" ^ string_of_int n in
+    Lwt_unix.file_exists f >>= fun exists ->
+    if exists then helper (n + 1)
+    else Lwt.return f
+  in
+  helper 0
+
+let open_temp_file:
+  ?buffer : Lwt_bytes.t ->
+  ?perm : Unix.file_perm ->
+  unit -> output_channel Lwt.t
+  = fun ?buffer ?perm ->
+    fun () -> temp_filename () >>= fun fname ->
+      open_file ?buffer:buffer ?perm:perm ~mode:Output fname
+
+let with_temp_file ?buffer ?perm f =
+  open_temp_file ?buffer ?perm () >>= fun ic ->
+  Lwt.finalize
+    (fun () -> f ic)
+    (fun () -> close ic)
+
 let file_length filename = with_file ~mode:input filename length
 
 let close_socket fd =
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1491,9 +1491,7 @@ let open_temp_file:
 
 let with_temp_file ?buffer ?flags ?perm ?temp_dir ?prefix f =
   open_temp_file
-    ?buffer:buffer ?flags:flags
-    ?perm:perm ?temp_dir:temp_dir
-    ?prefix: prefix () >>= fun (fname, chan) ->
+    ?buffer ?flags ?perm ?temp_dir ?prefix () >>= fun (fname, chan) ->
   Lwt.finalize
     (fun () -> f (fname, chan))
     (fun () -> close chan >>= fun _ ->
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1456,15 +1456,17 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
 
 let open_temp_file:
   ?buffer : Lwt_bytes.t ->
+  ?flags : Unix.open_flag list ->
   ?perm : Unix.file_perm ->
   unit -> (string * output_channel) Lwt.t
-  = fun ?buffer ?perm ->
-    let f = Some [Unix.O_WRONLY; Unix.O_CREAT; Unix.O_TRUNC;
-                  Unix.O_NONBLOCK; Unix.O_EXCL] in
+  = fun ?buffer ?flags ?perm ->
+    let flags = match flags with
+    | None -> Some [Unix.O_WRONLY; Unix.O_CREAT; Unix.O_EXCL; Unix.O_CLOEXEC]
+    | Some l -> Some l in
     let rec helper n =
       let fname = "lwt_io_temp_file_" ^ string_of_int n in
       Lwt.catch
-        (fun () -> open_file ?buffer:buffer ?flags:f ?perm:perm ~mode:Output fname
+        (fun () -> open_file ?buffer:buffer ?flags:flags ?perm:perm ~mode:Output fname
           >>= fun chan -> Lwt.return (fname, chan))
         (fun exn ->
            match exn with
@@ -1473,8 +1475,8 @@ let open_temp_file:
     in
     fun () -> helper 0
 
-let with_temp_file ?buffer ?perm f =
-  open_temp_file ?buffer:buffer ?perm:perm () >>= fun (fname, chan) ->
+let with_temp_file ?buffer ?flags ?perm f =
+  open_temp_file ?buffer:buffer ?flags:flags ?perm:perm () >>= fun (fname, chan) ->
   Lwt.finalize
     (fun () -> f (fname, chan))
     (fun () -> close chan >>= fun _ ->
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1454,7 +1454,7 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
     (fun () -> f ic)
     (fun () -> close ic)
 
-let open_temp_file_with_filename:
+let open_temp_file:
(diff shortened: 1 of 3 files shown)


src/unix/lwt_io.ml
```diff
@@ -1454,6 +1454,41 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
     (fun () -> f ic)
     (fun () -> close ic)
 
+let open_temp_file_with_filename:
+  ?buffer : Lwt_bytes.t ->
+  ?perm : Unix.file_perm ->
+  unit -> (string * output_channel) Lwt.t
+  = fun ?buffer ?perm ->
+    let f = Some [Unix.O_WRONLY; Unix.O_CREAT; Unix.O_TRUNC;
+                  Unix.O_NONBLOCK; Unix.O_EXCL] in
```
aantron: Nit, but I believe:

- `O_TRUNC` is not necessary, as you are passing `O_CREAT` and `O_EXCL`, so the file will not be an existing one.
- `O_NONBLOCK` has no effect on ordinary files, as being created here.

src/unix/lwt_io.ml
```diff
@@ -1454,6 +1454,41 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
...
+    in
+    fun () -> helper 0
+
+let open_temp_file:
+  ?buffer : Lwt_bytes.t ->
+  ?perm : Unix.file_perm ->
+  unit -> output_channel Lwt.t
+  = fun ?buffer ?perm ->
+    let f = open_temp_file_with_filename ?buffer:buffer ?perm:perm in
+    fun () -> f () >>= (fun (_, chan) -> Lwt.return chan)
```
aantron: Again a nit, but this would be easier to read if it didn't partially apply `open_temp_file_with_filename`, as that makes me worry about side effects if I partially apply `open_temp_file`.

How about

```ocaml
... fun ?buffer ?perm () ->
  open_temp_file_with_filename ?buffer:buffer ?perm:perm () >>= fun (_, chan) ->
  Lwt.return chan
```

or even

```ocaml
  open_temp_file_with_filename ?buffer:buffer ?perm:perm () >|= snd
```

src/unix/lwt_io.ml
```diff
@@ -1454,6 +1454,41 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
...
+
+let with_temp_file ?buffer ?perm f =
+  let otf = open_temp_file_with_filename ?buffer:buffer ?perm:perm in
+  otf () >>= fun (fname, chan) ->
```
aantron: Same issue here with worrying about side effects.

src/unix/lwt_io.ml
```diff
@@ -1454,6 +1454,41 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
...
+  Lwt.finalize
+    (fun () -> f chan)
+    (fun () -> close chan >>= fun _ ->
+      Lwt_unix.unlink fname)
```
aantron: Just a note, if `f` fails, and then `unlink` fails, the exception of `f` will be "swallowed" and replaced by the exception of `unlink`. See #442. I don't know what the right behavior should be. If this behavior is not what we want, it's probably `finalize` that needs to be fixed.

src/unix/lwt_io.ml
```diff
@@ -1454,6 +1454,41 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
...
 
+let open_temp_file_with_filename:
+  ?buffer : Lwt_bytes.t ->
+  ?perm : Unix.file_perm ->
+  unit -> (string * output_channel) Lwt.t
+  = fun ?buffer ?perm ->
+    let f = Some [Unix.O_WRONLY; Unix.O_CREAT; Unix.O_TRUNC;
+                  Unix.O_NONBLOCK; Unix.O_EXCL] in
+    let rec helper n =
+      let fname = "lwt_io_temp_file_" ^ string_of_int n in
```
aantron: In case of a process rapidly opening temporary file names in a loop, and keeping them open for a while, this will cause the first `N` filenames to be already taken on iteration `N + 1`, and this function will perform `N` system calls that will fail. I recommend using a random number generator instead, see [`Random.State`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Random.State.html). There is also an [example](https://github.com/ocaml/ocaml/blob/7a315bdb5e66bffdd7cef536373a2e0b753d1b7f/stdlib/filename.ml#L233) of it being used in the stdlib.

aantron: Also, I believe this will try to create temporary files in the current directory. We should try in system temporary file directories instead:

- [Unix](https://github.com/ocaml/ocaml/blob/7a315bdb5e66bffdd7cef536373a2e0b753d1b7f/stdlib/filename.ml#L86-L87)
- [Win32](https://github.com/ocaml/ocaml/blob/7a315bdb5e66bffdd7cef536373a2e0b753d1b7f/stdlib/filename.ml#L113-L114)

In particular, the current directory is fairly likely not to have permissions for creating files at all. We don't want to touch it for other reasons as well. For example, it may be monitored by a daemon that restarts builds, etc.

The stdlib also added optional arguments for the caller to control which directory is used: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Filename.html#VALtemp_file. We may want those in the interface as well.

You may be able to use that interface in your tests. It would still be nice to test temporary files in the system directory as well, but you may not have such an easy time scanning it.

src/unix/lwt_io.mli
```diff
@@ -404,6 +404,14 @@ val with_file :
       file and passes the channel to [f]. It is ensured that the
       channel is closed when [f ch] terminates (even if it fails). *)
 
+val open_temp_file :
+  ?buffer:Lwt_bytes.t ->
+  ?perm:Unix.file_perm -> unit -> output_channel Lwt.t
```
aantron: At least this function should evaluate to a `(string, output_channel) Lwt.t`. With this interface, I believe there is no way for the caller to delete the file afterwards. I believe this issue is why [Filename.open_temp_file](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Filename.html#VALopen_temp_file) was added. `with_temp_file` should also pass the filename to its callback.

src/unix/lwt_io.ml
```diff
@@ -1454,6 +1454,41 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
...
+  = fun ?buffer ?perm ->
+    let f = open_temp_file_with_filename ?buffer:buffer ?perm:perm in
+    fun () -> f () >>= (fun (_, chan) -> Lwt.return chan)
+
+let with_temp_file ?buffer ?perm f =
+  let otf = open_temp_file_with_filename ?buffer:buffer ?perm:perm in
+  otf () >>= fun (fname, chan) ->
+  Lwt.finalize
+    (fun () -> f chan)
+    (fun () -> close chan >>= fun _ ->
```
aantron: I don't remember off the top of my head what will happen if the file is already closed. Should probably add a test for that.

Closing this temporary file in the callback is going to be a very common use case for one-process use. A process will create a temporary file, write contents to it, close it, pass the filename to someone else, or open it itself again for reading later.

aantron:
Aside from the nits and small interface alterations, I think this is pretty much what's needed. Perhaps @copy can comment as well :)

src/unix/lwt_io.ml
fdopen: It should also be possible to override the default flags. You usually want to add `O_CLOEXEC`, but not always ...

src/unix/lwt_io.ml
copy: I would add that the prefix to the file name shouldn't be hardcoded and should be passed into the function so the user can give it a descriptive name.

copy:
This looks good to me, besides that the file name should be specified by the user.

aantron:
Thanks :)

src/unix/lwt_io.ml
```diff
@@ -1454,6 +1454,49 @@ let with_file ?buffer ?flags ?perm ~mode filename f =
...
+      let fname = temp_file_name dir prefix in
+      Lwt.catch
+        (fun () -> open_file ?buffer:buffer ?flags:flags ?perm:perm ~mode:Output fname
+          >>= fun chan -> Lwt.return (fname, chan))
+        (fun exn ->
+           match exn with
+           | Unix.Unix_error _ when n < 1000 -> helper (n + 1)
+           | _ -> raise exn)
+    in
+    fun () -> helper 0
```
aantron: Ah, this is still an example of doing side effects before full application (in this case, I believe a coeffect, as applying `open_temp_file` to all its labeled arguments but *not* `()` will cause it to read `Filename.get_temp_dir_name ()` if `?temp_dir:None`, which "fixes" the temporary directory, even if it is changed before applying to `()`, or between multiple applications to `()`). Sorry for missing it, I'll just tweak it in `master`.

PR #469, opened 2017-09-06
New reference docs for module Lwt

aantron:
See them [online][lwtmli-preview].

<br/>

[captura de pantalla 2017-09-06 a la s 10 31 02][lwtmli-preview]

[lwtmli-preview]: https://ocsigen.github.io/lwt/manual-draft/Lwt.html

<br/>

These should be considered an early draft. Reviews greatly appreciated, as are improvements. If you are good at design, improving the appearance would also be quite welcome :)

This is the first installment of the [new manual][doc-project].

[doc-project]: https://github.com/ocsigen/lwt/projects/2

<br/>

## Introductory material

Probably the most important for Lwt users are tutorials, cheat sheets, examples, and other introductory material. However, I decided to write those a bit later. For now, I'm working in order of what I personally find most intimidating, to get it out of the way :)

There *are* already (1) a [skeleton of a tutorial][tutorial], (2) a section on the Lwt [execution model][execution-model], and (3) an [overview of the library][overview]. Even though they deserve to be on separate pages, I've parked them module `Lwt` for the time being. I'll revise and factor them out later, as the overall manual starts to take shape.

[tutorial]: https://ocsigen.github.io/lwt/manual-draft/Lwt.html#3_Tutorial
[execution-model]: https://ocsigen.github.io/lwt/manual-draft/Lwt.html#3_Executionmodel
[overview]: https://ocsigen.github.io/lwt/manual-draft/Lwt.html#3_GuidetotherestofLwt

<br/>

## Terminology and structure

I've tried to avoid misleading mental models and terminology, such as *threads*. Instead, everything is explained in terms of *promises* and *callbacks*. I've also changed the order and sectioning completely, and added lots of simple sample programs.

I've switched completely to the JavaScript terminology for resolving promises, i.e. *fulfilling*, *rejecting*, and *resolving*. If/when this PR is merged, I'll update `lwt.ml` to use this terminology as well. Right now, it uses *resolving*, *failing*, and *completing*.

<br/>

## Appearance and composition

I applied a stylesheet and a postprocessor to the docs. Both are temporary, and contain a *lot* of unmaintainable hacks.

The full manual will need a clean new documentation generator or postprocessor. Otherwise, we won't be able to maintain a high level of quality over time.

Since that's not available for now, and the manual is not written anyway, in the immediate future, the text from this PR will be accessible only through the existing Ocsigen manual. Given that we intend to replace that manual, I don't think we should bother too much with the details of the text's appearance in it.

<br/>

## More

- The future manual will need a library-wide search or fast go-to function.
- We will probably need Reason examples and/or syntax.
- I converted ticked type variables to Greek letters. I think it looks nicer, but it may surprise some readers. I suspect many people are not aware that `'a` is actually pronounced *alpha*, etc. Using Greek letters also makes it more difficult to use the browser find function, and makes copy+paste less useful. So, I'm not sure Greek letters are worth it.

### DIFF
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -25,151 +25,55 @@
 
 type +'a t
 
-val return : 'a -> 'a t
-
-val fail : exn -> 'a t
-
-val fail_with : string -> 'a t
-
-val fail_invalid_arg : string -> 'a t
-
-val bind : 'a t -> ('a -> 'b t) -> 'b t
-
-val (>>=) : 'a t -> ('a -> 'b t) -> 'b t
-
-val (=<<) : ('a -> 'b t) -> 'a t -> 'b t
-
-val map : ('a -> 'b) -> 'a t -> 'b t
-
-val (>|=) : 'a t -> ('a -> 'b) -> 'b t
-
-val (=|<) : ('a -> 'b) -> 'a t -> 'b t
-
-val return_unit : unit t
-
-val return_none : 'a option t
-
-val return_some : 'a -> 'a option t
-
-val return_nil : 'a list t
-
-val return_true : bool t
-
-val return_false : bool t
-
-val return_ok : 'a -> ('a, _) Result.result t
-
-val return_error : 'e -> (_, 'e) Result.result t
-
-type 'a key
-
-val new_key : unit -> 'a key
-
-val get : 'a key -> 'a option
-
-val with_value : 'a key -> 'a option -> (unit -> 'b) -> 'b
-
-val catch : (unit -> 'a t) -> (exn -> 'a t) -> 'a t
-
-val try_bind : (unit -> 'a t) -> ('a -> 'b t) -> (exn -> 'b t) -> 'b t
-
-val finalize : (unit -> 'a t) -> (unit -> unit t) -> 'a t
-
-val wrap : (unit -> 'a) -> 'a t
-
-val wrap1 : ('a -> 'b) -> 'a -> 'b t
-
-val wrap2 : ('a -> 'b -> 'c) -> 'a -> 'b -> 'c t
-val wrap3 : ('a -> 'b -> 'c -> 'd) -> 'a -> 'b -> 'c -> 'd t
-val wrap4 : ('a -> 'b -> 'c -> 'd -> 'e) -> 'a -> 'b -> 'c -> 'd -> 'e t
-val wrap5 : ('a -> 'b -> 'c -> 'd -> 'e -> 'f) -> 'a -> 'b -> 'c -> 'd -> 'e -> 'f t
-val wrap6 : ('a -> 'b -> 'c -> 'd -> 'e -> 'f -> 'g) -> 'a -> 'b -> 'c -> 'd -> 'e -> 'f -> 'g t
-val wrap7 : ('a -> 'b -> 'c -> 'd -> 'e -> 'f -> 'g -> 'h) -> 'a -> 'b -> 'c -> 'd -> 'e -> 'f -> 'g -> 'h t
-
-val choose : 'a t list -> 'a t
-
-val nchoose : 'a t list -> 'a list t
-
-val nchoose_split : 'a t list -> ('a list * 'a t list) t
-
-val join : unit t list -> unit t
-
-val ( <?> ) : 'a t -> 'a t -> 'a t
-
-val ( <&> ) : unit t -> unit t -> unit t
-
-val async : (unit -> 'a t) -> unit
-
-val ignore_result : 'a t -> unit
-
-val async_exception_hook : (exn -> unit) ref
-
 type 'a u
 
-val wait : unit -> 'a t * 'a u
-
-val wakeup : 'a u -> 'a -> unit
-
-val wakeup_exn : 'a u -> exn -> unit
+val task : unit -> 'a t * 'a u
 
 val wakeup_later : 'a u -> 'a -> unit
 
 val wakeup_later_exn : 'a u -> exn -> unit
 
-val waiter_of_wakener : 'a u -> 'a t
+val return : 'a -> 'a t
 
-type +'a result = ('a, exn) Result.result
+val fail : exn -> 'a t
 
-val make_value : 'a -> 'a result
-  [@@ocaml.deprecated
-    " Use Result.Ok, which is the same as Ok since OCaml 4.03."]
+val bind : 'a t -> ('a -> 'b t) -> 'b t
 
-val make_error : exn -> 'a result
-  [@@ocaml.deprecated
-    " Use Result.Error, which is the same as Error since OCaml 4.03."]
+val catch : (unit -> 'a t) -> (exn -> 'a t) -> 'a t
 
-val of_result : 'a result -> 'a t
+val finalize : (unit -> 'a t) -> (unit -> unit t) -> 'a t
 
-val wakeup_result : 'a u -> 'a result -> unit
+val try_bind : (unit -> 'a t) -> ('a -> 'b t) -> (exn -> 'b t) -> 'b t
 
-val wakeup_later_result : 'a u -> 'a result -> unit
+val async : (unit -> 'a t) -> unit
 
-type 'a state =
-  | Return of 'a
-  | Fail of exn
-  | Sleep
+val async_exception_hook : (exn -> unit) ref
 
-val state : 'a t -> 'a state
-
-val is_sleeping : 'a t -> bool
-
-exception Canceled
-
-val task : unit -> 'a t * 'a u
-
-val on_cancel : 'a t -> (unit -> unit) -> unit
-
-val add_task_r : 'a u Lwt_sequence.t -> 'a t
-
-val add_task_l : 'a u Lwt_sequence.t -> 'a t
-
-val cancel : 'a t -> unit
+val join : unit t list -> unit t
 
 val pick : 'a t list -> 'a t
 
+val choose : 'a t list -> 'a t
+
 val npick : 'a t list -> 'a list t
+
+val nchoose : 'a t list -> 'a list t
+
+val nchoose_split : 'a t list -> ('a list * 'a t list) t
+
+exception Canceled
+
+val cancel : 'a t -> unit
+
+val on_cancel : 'a t -> (unit -> unit) -> unit
 
 val protected : 'a t -> 'a t
 
 val no_cancel : 'a t -> 'a t
 
-val pause : unit -> unit t
+val wait : unit -> 'a t * 'a u
 
-val wakeup_paused : unit -> unit
-
-val paused_count : unit -> int
-
-val register_pause_notifier : (int -> unit) -> unit
+val map : ('a -> 'b) -> 'a t -> 'b t
 
 val on_success : 'a t -> ('a -> unit) -> unit
 
@@ -183,16 +87,112 @@ module Infix : sig
 
   val (>>=) : 'a t -> ('a -> 'b t) -> 'b t
 
-  val (=<<) : ('a -> 'b t) -> 'a t -> 'b t
-
   val (>|=) : 'a t -> ('a -> 'b) -> 'b t
 
-  val (=|<) : ('a -> 'b) -> 'a t -> 'b t
+  val ( <&> ) : unit t -> unit t -> unit t
 
   val ( <?> ) : 'a t -> 'a t -> 'a t
 
-  val ( <&> ) : unit t -> unit t -> unit t
+  val (=<<) : ('a -> 'b t) -> 'a t -> 'b t
+
+  val (=|<) : ('a -> 'b) -> 'a t -> 'b t
 end
+
+val return_unit : unit t
+
+val return_none : 'a option t
+
+val return_nil : 'a list t
+
+val return_true : bool t
+
+val return_false : bool t
+
+type +'a result = ('a, exn) Result.result
+
+val of_result : 'a result -> 'a t
+
+val wakeup_later_result : 'a u -> 'a result -> unit
+
+type 'a state =
+  | Return of 'a
+  | Fail of exn
+  | Sleep
+
+val state : 'a t -> 'a state
+
+type 'a key
+
+val new_key : unit -> 'a key
+
+val get : 'a key -> 'a option
+
+val with_value : 'a key -> 'a option -> (unit -> 'b) -> 'b
+
+val wakeup : 'a u -> 'a -> unit
+
+val wakeup_exn : 'a u -> exn -> unit
+
+val wakeup_result : 'a u -> 'a result -> unit
+
+val make_value : 'a -> 'a result
+  [@@ocaml.deprecated
+    " Use Result.Ok, which is the same as Ok since OCaml 4.03."]
+
+val make_error : exn -> 'a result
+  [@@ocaml.deprecated
+    " Use Result.Error, which is the same as Error since OCaml 4.03."]
+
+val waiter_of_wakener : 'a u -> 'a t
+
+val add_task_r : 'a u Lwt_sequence.t -> 'a t
+
+val add_task_l : 'a u Lwt_sequence.t -> 'a t
+
+val pause : unit -> unit t
+
+val wakeup_paused : unit -> unit
+
+val paused_count : unit -> int
+
+val register_pause_notifier : (int -> unit) -> unit
+
+val wrap : (unit -> 'a) -> 'a t
+
+val wrap1 : ('a -> 'b) -> 'a -> 'b t
+
+val wrap2 : ('a -> 'b -> 'c) -> 'a -> 'b -> 'c t
+val wrap3 : ('a -> 'b -> 'c -> 'd) -> 'a -> 'b -> 'c -> 'd t
+val wrap4 : ('a -> 'b -> 'c -> 'd -> 'e) -> 'a -> 'b -> 'c -> 'd -> 'e t
+val wrap5 : ('a -> 'b -> 'c -> 'd -> 'e -> 'f) -> 'a -> 'b -> 'c -> 'd -> 'e -> 'f t
+val wrap6 : ('a -> 'b -> 'c -> 'd -> 'e -> 'f -> 'g) -> 'a -> 'b -> 'c -> 'd -> 'e -> 'f -> 'g t
+val wrap7 : ('a -> 'b -> 'c -> 'd -> 'e -> 'f -> 'g -> 'h) -> 'a -> 'b -> 'c -> 'd -> 'e -> 'f -> 'g -> 'h t
+
+val return_some : 'a -> 'a option t
+
+val return_ok : 'a -> ('a, _) Result.result t
+
+val return_error : 'e -> (_, 'e) Result.result t
+
+val fail_with : string -> 'a t
+
+val fail_invalid_arg : string -> 'a t
+
+val (>>=) : 'a t -> ('a -> 'b t) -> 'b t
+
+val (>|=) : 'a t -> ('a -> 'b) -> 'b t
+
+val ( <?> ) : 'a t -> 'a t -> 'a t
+
+val ( <&> ) : unit t -> unit t -> unit t
+
+val (=<<) : ('a -> 'b t) -> 'a t -> 'b t
+
+val (=|<) : ('a -> 'b) -> 'a t -> 'b t
+
+val is_sleeping : 'a t -> bool
+
+val ignore_result : 'a t -> unit
 
 (**/**)
 
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -26,24 +26,25 @@
 
 (** Asynchronous programming with promises.
 
-    {b Promises} are placeholders for values that take a long time to compute.
-    Promises are similar to [ref]s – they can store one value. Here is how they
-    differ from [ref]s:
+    A {b promise} is a placeholder for a single value which might take a long
+    time to compute. Speaking roughly, a promise is a [ref] that can be filled
+    in later. To make that precise, here is how promises differ from [ref]s:
 
-    - A promise might not have a value yet. This is called a {e pending}
-      promise.
-    - Writing a value into a promise is called {e resolving} it.
-    - It's possible to attach {b callbacks} to a pending promise. They will be
-      run when the promise gets resolved.
-    - A promise can be resolved only once. Once a promise has a value, it's
-      immutable.
+    - A promise might not have a value yet. A promise in this state is called a
+      {e pending} promise.
+    - Writing a value into a promise is called {e resolving} it. A promise with
+      a value is called a {e resolved} promise.
+    - Each promise can be resolved only once. After a promise has a value, the
+      promise is immutable.
+    - It's possible to attach {b callbacks} to a promise. They will run when the
+      promise has a value, i.e. is resolved. If the promise is already resolved
+      when a callback is attached, the callback is run (almost) right away. If
+      the promise is pending, the callback is put into a list and waits.
 
-    So, promises are write-once optional references, and when they don't yet
+    So, promises are optional, write-once references, and when they don't yet
     have a value, they store a list of callbacks that are waiting for the value.
 
-    Why is this useful?
-
-    The waiting callbacks make promises a nice data type for asynchronous
+    The waiting callbacks make promises a natural data type for asynchronous
     programming. For example, you can ask Lwt to [read] a file. Lwt immediately
     returns you only a {e promise} for the data.
 
@@ -196,9 +197,9 @@ let () =
     But, as you can see, this is verbose, and the indentation gets a bit crazy.
     So, we will always use [let%lwt].
 
-    The code above reads two lines in sequence, because we waited for [line_1],
-    before calling the second {!Lwt_io.read_line} in the callback, to start the
-    second I/O.
+    The code above reads two lines in sequence, because we ask Lwt to wait for
+    [line_1], before calling the second {!Lwt_io.read_line} in the callback, to
+    start the second I/O.
 
     We could also run I/O {e concurrently}. All we have to do is not start the
     second I/O in a callback of the first. Because it doesn't make sense to read
@@ -242,23 +243,30 @@ let () =
 ]}
 
     And that's it! Concurrency in Lwt is simply a matter of whether you start an
-    operation in the callback of another one or not.
+    operation in the callback of another one or not. As a convenience, Lwt
+    provides a few {{: #2_Concurrency} helpers} for common concurrency patterns.
+
+
 
     {3 Execution model}
 
     It's important to understand that promises are a pure-OCaml data type. They
-    don't do any fancy scheduling or I/O. They are just lists of callbacks if
-    pending, and containers for one value if resolved.
+    don't do any fancy scheduling or I/O. They are just lists of callbacks (if
+    pending), or containers for one value (if resolved).
 
     The interesting function is {!Lwt_main.run}. It's a wrapper around
     {{: http://man7.org/linux/man-pages/man2/select.2.html} [select(2)]},
     {{: http://man7.org/linux/man-pages/man7/epoll.7.html} [epoll(7)]},
     {{: https://www.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2}
-    [kqueue(2)]}, or whatever asynchronous API your system provides. Let's take
-    [select(2)] for the example.
+    [kqueue(2)]}, or whatever asynchronous I/O API your system provides. On
+    browsers, the work of {!Lwt_main.run} is done by the surrouding JavaScript
+    engine, so you don't call {!Lwt_main.run} from inside your program. But the
+    execution model is still the same, and the description below applies!
 
-    On browsers, the [select(2)] call is inside the JavaScript engine, so you
-    don't call {!Lwt_main.run}. But the logic is still the same!
+    To avoid writing out “underlying asynchronous I/O API,” we'll assume, in
+    this section, that the API is [select(2)]. That's just for the sake of
+    abbreviation. It doesn't actually matter, for most purposes, what the
+    underlying I/O API is.
 
     Let's use the program from the tutorial that reads two lines as an example.
     Here it is, again, in its desugared form:
@@ -281,7 +289,7 @@ let () =
 ]}
 
     {!Lwt_main.run} is your program's main I/O loop. You pass it a single
-    top-level promise, and it:
+    promise, and it:
 
     + Uses [select(2)] to put your process to sleep until the next I/O
       completes.
@@ -336,7 +344,8 @@ let () =
       is nothing conceptually special about rejection – it's just that you can
       ask for callbacks to run only on fulfillment, only on rejection, etc.
     - {{: #2_Cancelation} Cancelation}. This is a special case of rejection,
-      with extra helpers in the Lwt API.
+      specifically with exception {!Lwt.Canceled}. It has extra helpers in the
+      Lwt API.
     - {{: #2_Concurrency} Concurrency helpers}. All of these could be
       implemented on top of {!Lwt.bind}. As we saw, Lwt concurrency requires
       only deciding whether to run something inside a callback, or outside it.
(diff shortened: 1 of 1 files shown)


aantron:
Indeed, I often use identifiers in my work. I generated these docs with `ocamldoc`, though, and it forgets all identifiers and renames everything back to `'a`, `'b`, etc. I decided not to postprocess those out, because it would be even more hacks... This will probably be a future improvement.

mjambon:
I opened an issue about type variables here: https://caml.inria.fr/mantis/view.php?id=7623

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+
+    Why is this useful?
+
+    The waiting callbacks make promises a nice data type for asynchronous
+    programming. For example, you can ask Lwt to [read] a file. Lwt immediately
+    returns you only a {e promise} for the data.
+
+    You can neglect this promise for a while. You can do some other computation,
+    request more I/O, etc. At some point, you might decide to attach a callback
+    to the [read] promise, maybe several callbacks.
```
raphael-proust: I don't think this represents a typical usage of Lwt: in most cases, you attach the callbacks immediately. I.e., in typical use, there are more bind operations on unnamed promises than on named ones.

The description makes it look like you need to sort of be careful about when you bind things. Specifically, the combination of this paragraph and “In the meantime, the [read] operation is running in the background” (below) makes it sound like the concurrent operations occur because you waited to attach a callback.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+    differ from [ref]s:
 
-    Lwt also provides the syntax extension {!Pa_lwt} to make code
-    using Lwt more readable.
-*)
+    - A promise might not have a value yet. This is called a {e pending}
+      promise.
+    - Writing a value into a promise is called {e resolving} it.
+    - It's possible to attach {b callbacks} to a pending promise. They will be
+      run when the promise gets resolved.
```
raphael-proust: There should be a mention early on that the callbacks typically resolve another promise. Not necessarily here, but maybe bellow just after as part of the answer in “why it is this useful?”

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+
+
+    {3 Quick start}
+
+    {e All} of Lwt is variations on:
+
+    - {b Promises} of type ['a ]{!Lwt.t} are placeholders for values of type
+      ['a].
+    - {!Lwt.bind} attaches {b callbacks} to promises. When a promise gets a
+      value, its callbacks are called.
```
raphael-proust: (This goes with a previous comment about callbacks.) It's not clear here that bind attaches callbacks that resolve a promise. These callbacks can have side effects, but they also must (per type discipline) resolve a promise.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+        Lwt_io.printf "%s and %s\n" line_1 line_2))
+  in
+
+  Lwt_main.run p
+
+(* ocamlfind opt -linkpkg -package lwt.unix code.ml && ./a.out *)
+]}
+
+    But, as you can see, this is verbose, and the indentation gets a bit crazy.
+    So, we will always use [lwt%lwt].
```
raphael-proust: Maybe show the bind operator (`>>=`) version?

```
+{[
let () =
  let p : unit Lwt.t =
    Lwt_io.(read_line stdin) >>= fun line_1 ->
    Lwt_io.(read_line stdin) >>= fun line_2 ->
    Lwt_io.printf "%s and %s\n" line_1 line_2
  in

  Lwt_main.run p

(* ocamlfind opt -linkpkg -package lwt.unix code.ml && ./a.out *)
]}
```

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+
+    The code above reads two lines in sequence, because we waited for [line_1],
+    before calling the second {!Lwt_io.read_line} in the callback, to start the
+    second I/O.
```
raphael-proust: Nitpick: “we waited” is actually “Lwt waited”. I.e., it's all automatic and “we” don't have to manage the waiting.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+          (* Only 2 seconds were left in the 5-second wait, so
+              this callback runs 2 seconds after the first callback. *)
+          Lwt_io.printl "Only 2 more seconds passed")))
+  end
+
+(* ocamlfind opt -linkpkg -package lwt.unix code.ml && ./a.out *)
+]}
+
+    And that's it! Concurrency in Lwt is simply a matter of whether you start an
+    operation in the callback of another one or not.
```
raphael-proust: Maybe also mention that there are other operators (e.g., `join`) which are shown later.

raphael-proust:
Reviewed the first part. I'll review the rest later.

Richard-Degenne:
Suggestion off the top of my head: I feel like the manual could greatly benefit from diagrams and other graphical elements. I think it would make it easier to mentally represent the way certain functions works, for instance.

How complicated would that be to implement/maintain?

aantron:
I thought about that too, but I think I'd have to either get help with it, or save it for future work. I don't want to block getting the basic content in on designing a visual language right now.

If appealing enough, it could help the docs a lot, though :)

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
-    thread is running, it executes as much as possible, and then
-    returns (a value or an error) or sleeps.
+(** Asynchronous programming with promises.
 
-    Note that inside a Lwt thread, exceptions must be raised with
-    {!fail} instead of [raise]. Also the [try ... with ...]
-    construction will not catch Lwt errors. You must use {!catch}
-    instead. You can also use {!wrap} for functions that may raise
-    normal exception.
+    {b Promises} are placeholders for values that take a long time to compute.
```
ivg: not all promised values take long time to compute. So I would suggest the following wording: "Promises are placeholders for values that are determined asynchronously". Also, your definition here doesn't take into account that a promise might be rejected. 

See also ours definitions of [futures][1], that is another name for what you're calling promises :)

[1]: https://github.com/BinaryAnalysisPlatform/bap/blob/master/lib/bap_future/bap_future.mli#L23

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+]}
+
+    And that's it! Concurrency in Lwt is simply a matter of whether you start an
+    operation in the callback of another one or not.
+
+    {3 Execution model}
+
+    It's important to understand that promises are a pure-OCaml data type. They
+    don't do any fancy scheduling or I/O. They are just lists of callbacks if
+    pending, and containers for one value if resolved.
```
raphael-proust: s/and/or/

I'd even rewrite it so the structure of the sentence is more obvious:

> They are just a list of callbacks (for a pending promise) or a simple value (for a resolved promise).

Or something like that.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+
+    {b (1)} [Lwt.bind] returns a promise [p_3] immediately. [p_3] starts out
+    pending, and is resolved as follows:
+
+    - The first condition to wait for is that [p_1] becomes resolved. It does
+      not matter whether [p_1] is already resolved when [Lwt.bind] is called, or
+      becomes resolved later – the rest of the behavior is the same.
+    - If and when [p_1] becomes resolved, it will, by definition, be either
+      fulfilled or rejected. Fulfilling is the interesting case, but rejection
+      is handled very simply, so we cover rejection of [p_1] first.
```
raphael-proust: This sentence about interesting and simple cases is not needed. (It's singposting but on an hyperlocal scale.)

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+    - If [p_1] is rejected, [p_3] is rejected with the same exception.
+    - If [p_1] is fulfilled, with value [v], [f] is applied to [v].
+    - [f] may finish by returning the promise [p_2], or raising an exception.
+      Again, the exception case is simpler, so we cover that first.
```
raphael-proust: Same as above: no signposting needed.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+    - If [f] raises an exception, [p_3] is rejected with that exception.
+    - Finally, the remaining case is when [f] returns [p_2]. From that point on,
+      [p_3] is effectively made into a reference to [p_2]. This means they have
```
raphael-proust: The notion of references is never discussed before. I don't think it's an issue, but then I'm familiar with the library. How confusing is it for a beginner?

Drup:
Nice work!

I don't really agree with the presentation order. Maybe it's because I'm familiar with the previous one, but I think you should first present Lwt.t/return/bind (map ?), then move on to Lwt.u/task/wakeup_later (we should find a better term for the last one, if that's to become the official basic primitive). I would also move concurency above rejection. In general, I would present things in an order that is coherent in how often you use them.

> I converted ticked type variables to Greek letters. 

Having experience with numerous research paper that do this: **DON'T**. It makes copy pasting a huge PITA and is pretty much useless anyway.

raphael-proust:
Yes, I agree that `t`/`bind`/`return` should be first.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+    {3 Guide to the rest of Lwt}
+
+    This module [Lwt] is the pure-OCaml definition of promises and
+    callback-calling. It has a few extras on top of what's described above:
+
+    - {{: #2_Rejection} Rejection}. Lwt promises can actually be resolved in two
+      ways: {e fulfilled} with a value, or {e rejected} with an exception. There
+      is nothing conceptually special about rejection – it's just that you can
+      ask for callbacks to run only on fulfillment, only on rejection, etc.
+    - {{: #2_Cancelation} Cancelation}. This is a special case of rejection,
```
avsm: perhaps specify that the `Lwt.Canceled` exception is the special case value here.

src/core/lwt.mli
aantron: I see what you're saying, but I actually don't want to describe typical usage here. I think the typical usage obscures what Lwt actually is. It makes Lwt look like straight-line code. I think that's partially responsible for how difficult it is to "get" Lwt (even if it looks nice). My goal here, at the beginning of the docs, is to build up an accurate mental model quickly, in the simplest terms possible. Any suggestions on that are welcome.

I'll think about how to make it clear that waiting is not related to concurrency.

src/core/lwt.mli
aantron: This was deliberate, because I'm effectively using `Lwt.bind` as a proxy for all those functions that attach callbacks, not being literal about `bind` itself. I don't want it to be clear that `bind`'s callback returns another promise, because not all the other functions have that (`Lwt.map`, `Lwt.on_success`), so it's not actually what I want to be relevant here.

I'm basically looking at `Lwt.bind` as a special case of a more basic mechanism (callbacks), and I believe that mechanism is more important to getting an understanding of Lwt, than the type signature of `bind`.

`bind` is, of course, the most common one. But I think the reader can find its usage and type signature by clicking on the link. There, they will see the promise-valued callback, and see why it is that way.

Further discussion, objections, etc., welcome :)

src/core/lwt.mli
aantron: Instead of this?

Just background:

I decided to offer the PPX for everything by default. To save space, and avoid needlessly offering too many options, I mention a `>>=` version of some code in the docs of `Lwt.bind` only.

src/core/lwt.mli
aantron: And to be more specific, the "simplest possible terms" I chose were 1. promises and 2. attaching callbacks to promises, so this affects how the rest of the intro reads.

src/core/lwt.mli
aantron: See later comment for discussion. I think it's better to show that through examples (to be created later), rather than complicate the promise/callback idea. It is mentioned in execution model, that callbacks often resolve more promises.

I chose to focus on a single "step" of execution here, rather than the whole Lwt "loop." I hope it gives a simpler and more local view to the reader.

aantron:
@Drup, @raphael-proust, when I was learning about Lwt, I often heard a focus on `bind`, "bind points," and some other things with `bind` I can't remember immediately. I think this focus gives an inaccurate mental model of how Lwt works, or at least distracts people. Certainly, my mental model of Lwt, when treated as a concurrency library, rather than a monad, does not have any special place for `bind` or `return`. Those are practically accidents, though useful in a huge number of situations, and though they make Lwt (almost?) a monad.

Instead, my mental model of "Lwt the concurrency library" is pretty much based on the idea of promises as fat refs, and essentially arbitrary callbacks being stored in a list in pending promises. Both promises and callbacks are simple-to-imagine heap "objects." They are also good for describing the execution model of Lwt. So, I decided to refocus the entire API docs on them.

Note that you can conceive of Lwt without `bind` and `return`, in fact `task`, `wakeup_later_result`, and `on_any`, I think, are sufficient primitives for simulating the parts of Lwt that are not deprecated (I maybe missing something).

This affects the presentation order. In my experience, if I hear that promises exist, the first thing I want to know is how to create one, specifically one that I think I actually would want to use – a pending one. It's true that `return` creates a promise too, but a degenerate one that isn't relevant to a new reader of the API that might still be looking for the value in Lwt. So, that's why I put `Lwt.task` first.

Next, `bind` is useless if a pending promise doesn't change state, so I put the state changes next. After that, `bind`. `return` snuck in there is a convenient helper function.

As for order of concurrency and rejection, I actually tend to use `Lwt.catch` and `Lwt.finalize` a lot more than the concurrency helpers, so I think it's debatable what the order should be, on the basis of frequency of usage. However, all the concurrency helpers have rejection propagation behavior, so I chose to put rejection first, although this is a bit of a weak argument as I didn't also choose to put cancelation first. I guess another factor was that rejection is one of the major states of a promise.

aantron:
The other advantage of presenting `task` early, I think, is it shows the proper places of resolvers right away.

src/core/lwt.mli
aantron: What happened here is that I gave up hope trying to find a formulation that is both accessible and precise for the first sentence. Of course, what is written there is not fully right. But:

- The word "asynchronous" seems like it will require further definition, and I want a sentence comprehensible immediately here.
- "Promises are placeholders for values that are determined asynchronously" is also imprecise, in a slightly different way, because promises can be created already-determined, for example.

So far, given the set of not-really-correct choices (including some other ones I came up with), I'm trying to choose the incorrect one that gets the most benefit out of being incorrect, by giving as immediate an intuition as possible :) After all, promises are *mostly* to be used as placeholders for values that *might* take some large amount of time to compute :)

As for rejection, I decided to simply leave it out, as a complication to be discussed later. It doesn't seem that different from saying that a value of type `int -> string` in OCaml is a function that evaluates to `string`s, and forgetting to mention that it might raise an exception (or not terminate).

aantron: Also, this reference is suffering the effects of having the tutorial inlined. Once that's factored out, I think it might be more realistic to use your suggestion, be precise about rejection in the first paragraph immediately, etc., and refer readers to the tutorial for a more progressive introduction. However, since I don't have the separate tutorial now to see how that will look, I prefer to introduce things progressively in `lwt.mli` for the time being.

ivg:
Btw, speaking of early misconceptions, when I was learning lwt the first time, I missed the fact that Lwt is a library for cooperative multi-tasking, not the preemptive. So my first programs had a lot of problems with promises that were always determined and never cooperate with another, sorry, threads. So probably it would be a good idea to empathize the cooperative nature of the tasks. That basically a computation of a promise continues until it either blocks or explicitly yields the control to another computation.

aantron:
I tried to use the Execution Model section to describe almost exactly this. However, I want to avoid the idea of "cooperative tasks," because I want to avoid the idea of "tasks" at all. There are in fact no tasks in Lwt. I think it was previously a mistake to introduce them, which mistake is partly responsible for your misconception (and mine, and many others'). Once we start talking about cooperative tasks, people start thinking about preemptive tasks, contrasting them, looking for similarities, etc. There are basically no relevant similarities, we don't want to risk the chance of borrowing ideas from preemptive tasking, and so the whole task terminology is misleading and IMO should be avoided.

aantron:
Also, the Execution Model section aside, I hope that once the reader is thinking only of promises, which are fat refs in the heap, and callbacks being called, the reader won't be tempted to think that there's anything more magical going on with callbacks (such as preemption) than if, say, your function gets called by `List.map`. They both just run until they produce a value or an exception, or forever.

So, hopefully, presenting Lwt in terms of completely ordinary callbacks *preempts* (argh) the possibility of this confusion. If the reader wants to see more details (Lwt starts to seem too magical), they can head to Execution Model (which will later be a separate page, I think).

In general, the goal here is to present Lwt as as un-magical and obvious as possible, to avoid having to read any such long explanations :)

aantron:
...and the same goes for the idea of an Lwt *scheduler*. Lwt doesn't really have a scheduler.

The core library has a functionality to defer calling of callbacks, in order basically to avoid stack deepening. This is the "completion loop" (to be renamed) in `lwt.ml`. It's basically an implementation detail.

`Lwt_main` has a wrapper around `select` or your other I/O API, that knows how to associate certain promises with respective I/O actions, so that the promises get resolved when `select` says those I/O actions completed.

Neither one of these mechanisms is a scheduler in any useful sense, but calling either one a scheduler makes it seem like there is more in Lwt than actually is there. People waste time trying to understand the elusive Lwt scheduler.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+  Lwt_main.run p
+
+(* ocamlfind opt -linkpkg -package lwt.unix code.ml && ./a.out *)
+]}
+
+    {!Lwt_main.run} is your program's main I/O loop. You pass it a single
+    top-level promise, and it:
+
+    + Uses [select(2)] to put your process to sleep until the next I/O
+      completes.
```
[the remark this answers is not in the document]
aantron: `Lwt_main.run` doesn't execute `p`, because `p` is a promise (a boring data "object," not something that can be executed). If one writes, for example,

```ocaml
Lwt_main.run (Lwt_io.(read_line stdin))
```

it's equivalent, down to the syntactic level, to

```ocaml
let p = Lwt_io.(read_line stdin) in
Lwt_main.run p
```

What happens here is that running `Lwt_io.read_line` sets up the `read`, regardless of the existence of any later call to `Lwt_main.run`, and by the time `Lwt_main.run` is called, there is just a pending promise with nothing at all left to execute (except `select`).

So, indeed, `Lwt_main.run` does first put your process to sleep until the next I/O to complete. This is true in all cases. I specifically want to be clear about this, to avoid the idea that `Lwt_main.run` ever executes anything, *except* after calling `select` and seeing least one I/O has completed (modulo any internal optimizations). I also want to avoid the idea that `p` is special in any way, in the sense of `Lwt_main.run` driving the evaluation of the expression that produces it. That expression is evaluated due to the surrounding semantics of OCaml. The only thing special about `p` that concerns `Lwt_main.run`, is that it acts as a flag for stopping the I/O loop and not calling `select` anymore, and that whether `p` is fulfilled or rejected determines which value `Lwt_main.run` evaluates to, respectively what exception it raises.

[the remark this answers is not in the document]
aantron: I guess I should clarify in the docs that `Lwt_main.run` doesn't drive the I/O that is associated with `p` either. That is, it is not because we passed `p` to `Lwt_main.run`, that the I/O "behind" the `read_line` call ran.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+    + That next I/O happens to be the one that reads [line_1]. {!Lwt_main.run}
+      knows that I/O is supposed to resolve [line_1_promise], so it puts
+      [line_1] into the promise and resolves it.
+    + Resolving is an ordinary OCaml operation. It causes all the callbacks of
+      [line_1_promise] to run, one after another. Each callback is also ordinary
+      OCaml code. In our case, there is only one callback, but in general, there
+      might be several, and they might also resolve additional promises. So,
+      promise resolution triggers a “cascade” of callbacks. Eventually, however,
+      we should run out of callbacks, and control will return to
+      {!Lwt_main.run}.
```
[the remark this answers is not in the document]
aantron: > Also, another thing it should say is that this example is a simple promise with a string of callbacks. However, in more complex programs with non-trivial arrangements of callbacks/promises, instead of putting the process to sleep when it reaches an I/O operation, the scheduler will first execute callbacks that are ready. Only when it runs out of those does it select.

Actually, the scheduler (there is none, see [comment](https://github.com/ocsigen/lwt/pull/469#issuecomment-333222702), so I mean `Lwt_main.run`) doesn't execute any callbacks that are ready. They are all executed implicitly by promise resolution, by callbacks triggering each other, not by a scheduler. So this paragraph is that additional paragraph you suggest to write, but I guess it should be clarified, and maybe the `Lwt_main.run` paragraph should be split off and/or reordered.

aantron:
Seeing the comments about Greek letters from @Drup and @mjambon, I'll drop them. The stylesheet and postprocessor aren't part of the PR, and are for preview purposes only, so we'll simply not use Greek letters in a future "true" Lwt manual :)

mjambon:
@aantron it turns out that the problem with type variables getting mangled by ocamldoc no longer exists or it only occurs in a special case and we can't reproduce. See https://caml.inria.fr/mantis/view.php?id=7623

Please check your setup and let us know if you're having the problem with a recent version of ocamldoc.

aantron:
@mjambon it does work indeed. Thanks for letting me know!

src/core/lwt.mli
raphael-proust: I see what you mean by having readers build a correct mental model instead of being shown standard/usual/typical code. I still thinks that this paragraph can be misleading w.r.t. the mental model because readers may infer that binding interferes with promise resolution.

Maybe explicitly saying something along the lines of

> afterwards, two things happen (in any order):
> - you attach a callback to the promise
> - the read operation terminates which resolves the promise
>
> These operations can happen in any order. Whether you bind before or after the read terminates does not affect the program.
>
> When both of these operations have happened, the callback is called.

src/core/lwt.mli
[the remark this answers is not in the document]
raphael-proust: The explanation in this comment thread is good. Maybe it should be included (with minor edits) into the manual!

src/core/lwt.mli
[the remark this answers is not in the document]
raphael-proust: With this conversation, I'm realising how bad I am at explaining Lwt… Which makes me realise: Good work on the whole thing!

src/core/lwt.mli
aantron: That's probably a good idea. I'll probably spend a loooong time figuring out a really brief way to say that :p

src/core/lwt.mli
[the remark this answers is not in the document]
aantron: Yes, thanks for helping to work it out :)

I think a lot of this belongs in the docs of `Lwt_main.run`, but since `Lwt_main` is not up for rewriting yet, maybe I can somehow "park" all of it in Lwt for now. Or just leave a note to revisit the distribution of information between `Lwt` and `Lwt_main` run later.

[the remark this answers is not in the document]
aantron: Also, I just realized I gave myself a thumbs-up, lol. I moved it to your comment, where it belongs :)

src/core/lwt.mli
aantron: Also, I want to note, that my idea of the completed docs includes a set of examples. I hope that will be a good place to see typical usage. There should also be a tutorial, and this mental-model-building should eventually end up there, rather than in the reference docs.

src/core/lwt.mli
aantron: Technically, I think "and" is correct there without parentheses. This is because of the conditional clauses. (Exclusive) "or" is correct with parentheses, because the sentence needs to make sense with the parenthesized portions elided. But you're right, the sentence needs attention anyway, so I'm looking at it...

src/core/lwt.mli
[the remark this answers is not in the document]
aantron: Looking at it more, I decided to save it for addressing later, when more of the docs are written. This is because this page is already trying to be

1. A dry reference for `Lwt`
2. A walkthrough for `Lwt`
3. Examples for `Lwt`
4. A dry reference for `Lwt_main.run`
5. A walkthrough for `Lwt_main.run`
6. Examples for `Lwt_main.run`
7. A few other things
8. Still relatively concise and comprehensible

It's basically not going to satisfy all of these. We will revisit this once there are more pages to tease this apart into.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+
+    - {e fulfilled}, and containing one value of type ['a],
+    - {e rejected}, and containing one exception, or
+    - {e pending}, in which case it may become fulfilled or rejected later.
+
+    A {e resolved} promise is one that is either fulfilled or rejected, i.e. not
+    pending. Once a promise is resolved, its content cannot change. So, promises
+    are {e write-once references}. The only possible state changes are (1) from
+    pending to fulfilled and (2) from pending to rejected.
+
```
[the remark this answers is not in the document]
aantron: Why here and not in the next paragraph?

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+    Promises are typically “read” by attaching {b callbacks} to them. The most
+    basic functions for that are {!Lwt.bind}, which attaches a callback that is
+    called when a promise becomes fulfilled, and {!Lwt.catch}, for rejection.
+
+    Promises variables of this type, ['a Lwt.t], are actually {b read-only} in
+    Lwt. Separate {e resolvers} of type ['a ]{!Lwt.u} are used to write to them.
+    Promises and their resolvers are created together by calling {!Lwt.task}.
+    There is one exception to this: most promises can be {e canceled} by calling
+    {!Lwt.cancel}, without going through a resolver. *)
```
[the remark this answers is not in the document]
aantron: I'm not sure why. `Lwt.u` also aren't writeable promises, they are the write ends of each promise for which you happen to have an exposed `Lwt.u`. All (pending) promises are writeable, it's just the `Lwt.t` reference to a promise doesn't allow writing through it. Anyway, this paragraph's idea structure is:

1. `Lwt.t` read-only
2. The write end is `Lwt.u`, which is usually obtained by `task`
3. Cancelation is an exception to (1), and weakly to (2) (to the implication that you *need* the `Lwt.u` to write to a promise).

It would seem like an important omission to leave (3) out of `Lwt.t` by moving it to `Lwt.u`, as it's very important for reasoning about promises that the only way to write to them through an `Lwt.t` is by cancelation.

How does this read to you?

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+
+    If the promise is not pending, [Lwt.wakeup_later] raises
+    {{: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Pervasives.html#VALinvalid_arg}
+    [Pervasives.Invalid_argument]}, unless the promise is {{: #VALcancel}
+    canceled}. *)
+
+val wakeup_later_exn : _ u -> exn -> unit
+(** [Lwt.wakeup_later_exn r exn] is like {!Lwt.wakeup_later}, except it
+    {e rejects} the associated {e pending} {{: #TYPEt} promise} with [exn]
+    instead. *)
```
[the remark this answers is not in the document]
aantron: The place where it's mentioned earlier is actually a hyperlink, so the user can find it easily from there. Also, I think we should start getting away from `Lwt.result` (not `Lwt_result`!), and towards `Lwt_result` :p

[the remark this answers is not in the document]
aantron: That said, if you or others use `Lwt.result`, `wakeup_later_result`, etc., a lot, we can move the function up. I'd rather not link to it from directly inside the description of `wakeup_later_exn`, because then it seems we'd also have to link to it from `wakeup_later`, and it becomes weird to give it such special status eventually.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+    value instead, you have to wrap it:
+
+{[
+let%lwt line = Lwt_io.(read_line stdin) in
+Lwt.return (line ^ ".")
+]} *)
+
+val fail : exn -> _ t
+(** [Lwt.fail exn] is like {!Lwt.return}, except the new {{: #TYPEt} promise}
+    that is {e already rejected} with [exn]. *)
```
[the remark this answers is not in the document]
aantron: The thing with `Lwt.fail` is that it's not actually as useful as `Lwt.return`.

1. `raise`d exceptions are captured in almost all contexts where Lwt is used, and converted into calls to `Lwt.fail` (or the equivalent) anyway.
2. `raise` trivially satisfies the type system.
3. The rest of the reference recommends `try%lwt`, and the major use of `Lwt.fail`, re-raising exceptions, is taken care of by that.

So I have a hard time introducing a non-misleading usage. That's why I find it easier to refer to `Lwt.return` without new examples :p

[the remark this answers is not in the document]
aantron: I'm happy for people to educate me here, but I use `raise` a lot when programming with Lwt, and `fail` almost never. Also, I'd like to move away somewhat from exceptions, see earlier comment about favoring `Lwt_result`.

src/core/lwt.mli
```diff
@@ -21,502 +22,1785 @@
...
+    {e fulfilled}}.
+
+    The value that [p_1] is fulfilled with is passed to [f]. [f] then returns
+    another promise, [p_2]. This second promise means [f] can spawn more I/O
+    operations – they will create [p_2].
+
+    [Lwt.bind] itself returns immediately. It only attaches the callback [f] to
+    [p_1] – it does not wait for [p_2]. {e What} [Lwt.bind] returns is yet a
+    third promise, [p_3]. Roughly speaking, fulfillment of [p_3] represents both
+    [p_1] and [p_2] becoming fulfilled, one after the other.
```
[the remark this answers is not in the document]
aantron: Thanks for this, it's a lot clearer. I'm going for the second version, with "performing" replaced by "perhaps starting."

src/core/lwt.mli
aantron: I decided to be clearer about attaching callbacks to pending promises/trying to attach them to resolved promises immediately in the bulleted list that lists differences between promises and `ref`s. Let's see how that works..

src/core/lwt.mli
aantron: I'm not sure what to do here, and I'm tempted to push it to a future rewrite (I'll give this page at least one more thorough look as the manual comes together, plus anyone else is welcome to).

The "references" I'm mentioning here are an extra-library concept, as in runtime-level references (the same way `Lwt.t`s and their corresponding `Lwt.u`s are references to the same object in memory). Not `ref` references. I'm hoping that enough readers of these docs will be familiar with this idea. Of course, I may be wrong.

aantron: I guess it's not that beginner friendly. We could try wording like "made to behave exactly the same as `p_2`," but I want to convey a stronger image that `p_3` *is* `p_2`, because it pretty much is, given how Lwt is implemented :) "Behave exactly the same as `p_2`" leaves me with some kind of uncertainty about the extent to which that's true, etc.

aantron:
I took a break, to be able to get a really fresh look at the text. I can't think about text as a reader, after being caught up in thousands of its minor details as a writer for several weeks straight. With the clearer view, let's just say I find much of my own writing cringeworthy :p

I've tried to address the review comments, by changing the text (or debating!). The rendered version of the latest commit is now [online](file:///Users/antron/Coding/lwt-manual/Lwt.html) for your reviewing pleasure.

There will be at least one more point in the future when I will take a hard look at module Lwt: as the rest of the docs take shape, we will need to make sure that `lwt.mli` makes sense in context, and all the docs work well together. We can make many additional changes then, by then, or after then.

Then, or before, we will have to move out a lot of the non-API-reference material to separate manual pages.

Always, everyone is welcome to edit the docs, etc.

For now, the docs will be rendered only on the Ocsigen website. The preview I've been manually creating was only for this PR.

I hope to merge this in a few days or a week, so please sneak in any final comments. However, as mentioned, it's also possible to make commits after the PR is merged.

src/core/lwt.mli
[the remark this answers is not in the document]
raphael-proust: I understand making the different comments about the different function somewhat proportionate to the importance that they have for the programmer. However, here the functions have similarities in terms of functionality/use (which is visible from the common part of their signatures).

When things share functionalities, explaining them with repetition is generally a good idea. Specifically, using *parallel syntax* helps the reader understand what's shared and what's not shared between the functions. See “parallel syntax” in Steven Pinker's The Sense of Style.

Maybe this type of explanations is better suited in a full API reference doc rather than an introductory tutorial. Is it intended more as the former of the latter?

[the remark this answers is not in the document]
aantron: Point well taken.

> Maybe this type of explanations is better suited in a full API reference doc rather than an introductory tutorial. Is it intended more as the former of the latter?

Currently it's an incoherent mix of everything, but the specific part we are commenting on here, when it matures, is supposed to be an entry in a full API reference doc, whereas the introductory material will be moved out.

PR #471, opened 2017-09-15
Deprecate >> in ppx

### LINKED ISSUE #387: Deprecate >> in the PPX
See [discussion](https://discuss.ocaml.org/t/operator-in-lwt/296/11).

We should probably have the PPX emit a warning each time `>>` is used.

cc @Richard-Degenne @dkim @Drup

raphael-proust:
This PR is for issue: #387

It adds a warning to generated code that uses `>>` from ppx.

Note that, in its current form, the PR removes tests for `>>` altogether. These tests fails because of `-w @a` in the flags which turn all warning into errors. Comments and suggestions on this are welcome.

EDIT by @aantron: resolves #387.

### DIFF
--- a/src/ppx/ppx_lwt.ml
+++ b/src/ppx/ppx_lwt.ml
@@ -339,15 +339,29 @@ let mapper =
           let pat = if !strict_seq then [%pat? ()] else [%pat? _] in
           let lhs, rhs = mapper.expr mapper lhs, mapper.expr mapper rhs in
           if !debug then
-            [%expr Lwt.backtrace_bind (fun exn -> try raise exn with exn -> exn)
-                                      [%e lhs]
-                                      (fun [%p pat] -> [%e rhs])
-              [@ocaml.ppwarning "The operator >> is deprecated"]
-            ]
+            Ast_helper.Exp.attr
+              [%expr Lwt.backtrace_bind (fun exn -> try raise exn with exn -> exn)
+                                        [%e lhs]
+                                        (fun [%p pat] -> [%e rhs])
+              ]
+              (Ast_mapper.attribute_of_warning
+                { Location.
+                  loc_start = lhs.pexp_loc.Location.loc_end;
+                  loc_end = rhs.pexp_loc.Location.loc_start;
+                  loc_ghost = false;
+                }
+                "The operator >> is deprecated"
+              )
           else
-            [%expr (Lwt.bind [%e lhs] (fun [%p pat] -> [%e rhs]))
-              [@ocaml.ppwarning "The operator >> is deprecated"]
-            ]
+            Ast_helper.Exp.attr
+              [%expr (Lwt.bind [%e lhs] (fun [%p pat] -> [%e rhs]))]
+              (Ast_mapper.attribute_of_warning
+                { Location.
+                  loc_start = lhs.pexp_loc.Location.loc_end;
+                  loc_end = rhs.pexp_loc.Location.loc_start;
+                  loc_ghost = false;
+                }
+                "The operator >> is deprecated")
         else
           default_mapper.expr mapper expr
       | { pexp_desc = Pexp_apply (fn, args); pexp_attributes; pexp_loc } when !log ->
--- a/src/ppx/ppx_lwt.ml
+++ b/src/ppx/ppx_lwt.ml
@@ -334,7 +334,7 @@ let mapper =
             "Lwt's finally should be used only with the syntax: \"(<expr>)[%%finally ...]\"."
         ))
 
-      | [%expr [%e? lhs] >> [%e? rhs]] ->
+      | [%expr [%e? lhs] >> [%e? rhs]] as e ->
         if !sequence then
           let pat = if !strict_seq then [%pat? ()] else [%pat? _] in
           let lhs, rhs = mapper.expr mapper lhs, mapper.expr mapper rhs in
@@ -344,23 +344,12 @@ let mapper =
                                         [%e lhs]
                                         (fun [%p pat] -> [%e rhs])
               ]
-              (Ast_mapper.attribute_of_warning
-                { Location.
-                  loc_start = lhs.pexp_loc.Location.loc_end;
-                  loc_end = rhs.pexp_loc.Location.loc_start;
-                  loc_ghost = false;
-                }
-                "The operator >> is deprecated"
-              )
+              (Ast_mapper.attribute_of_warning e.pexp_loc
+                "The operator >> is deprecated")
           else
             Ast_helper.Exp.attr
               [%expr (Lwt.bind [%e lhs] (fun [%p pat] -> [%e rhs]))]
-              (Ast_mapper.attribute_of_warning
-                { Location.
-                  loc_start = lhs.pexp_loc.Location.loc_end;
-                  loc_end = rhs.pexp_loc.Location.loc_start;
-                  loc_ghost = false;
-                }
+              (Ast_mapper.attribute_of_warning e.pexp_loc
                 "The operator >> is deprecated")
         else
           default_mapper.expr mapper expr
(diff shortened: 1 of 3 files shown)


aantron:
Thanks for doing this! I haven't tried it out yet, but to offer a suggestion about the warnings: you may be able to disable them by tagging the test with `[@ocaml.warning "-whatever"]`. Since I haven't tried it yet, I don't know what warning number `ppwarning` emits, if any.

raphael-proust:
Good idea! I tried to add some tags to the local jbuild rule to make warning 22 (deprecation) non-fatal but didn't manage to. I was trying to keep the warning but make it just that: a warning. I'll try deactivating it completely.

raphael-proust:
I didn't manage to use the `@` attribute to disable the deprecation warning. Either I don't know how to use that correctly (which is likely) or it's somehow not passed along after the ppx is applied or somewhere in between: it's passed along but it somehow interact with the scoping.

aantron:
Indeed, I wasn't able to get it work using any form of `ocaml.warning` annotation either. Disabling warning 22 for the whole build of that file seems fine.

I think the only thing missing is somewhat better locations. For example, for this file:

```ocaml
let () =
  Lwt_main.run begin
    Lwt_unix.sleep 1. >>
    Lwt_unix.sleep 1. >>
    Lwt_unix.sleep 1.
  end
```

the output is

```
File "foo.ml", line 1, characters 0-107:
Warning 22: The operator >> is deprecated
File "foo.ml", line 1, characters 0-107:
Warning 22: The operator >> is deprecated
```

Speculating, but I think the location is the whole file or so :)

src/ppx/ppx_lwt.ml
```diff
@@ -341,9 +341,13 @@ let mapper =
...
                                       [%e lhs]
-                                      (fun [%p pat] -> [%e rhs])]
+                                      (fun [%p pat] -> [%e rhs])
+              [@ocaml.ppwarning "The operator >> is deprecated"]
+            ]
           else
-            [%expr Lwt.bind [%e lhs] (fun [%p pat] -> [%e rhs])]
+            [%expr (Lwt.bind [%e lhs] (fun [%p pat] -> [%e rhs]))
+              [@ocaml.ppwarning "The operator >> is deprecated"]
+            ]
```
Drup: You don't provide any location for the attribute here. The location is important, since it's used in the error message. You should build up the attribute by hand using `attribute_of_warning`, it makes things simpler.

Drup:
About localized disabling of warning 22 ... I think it should work on trunk. @octachron might be able to tell more.

Octachron:
I can confirm that disabling locally preprocessor warnings works on trunk.

aantron:
That's good, thanks. We'll have to use the warning options, though, because we are testing on versions down to 4.02.

raphael-proust:
So I'm struggling a bit with using the `Ast_mapper` and `ppx_metaquot`. One of the issue is that documentation for `metaquot` seems difficult to find. Maybe partly because search engines don't play nice with quotations and special characters… There's this http://ocaml.io/ppx_tools_versioned/index.html which, according to the `src/ppx/jbuild` is in use.

Where can I find documentation?

I manage to get something using the `[%attribute …]` construct, but that was a shot in the dark (`[%expr …] and `[%pat …]` exist so I just tried `[%attribute …]`. The problem I have now is that when running the test, it complains about some construct in the test *before* the sequence test having too many arguments.

aantron:
My go-to documentation for metaquot was always the comment in its source code:

https://github.com/ocaml-ppx/ppx_tools/blob/669f00cba3e5e3d3e5e1c452fff67cde1ec8876f/ppx_metaquot.ml#L5-L60

Let us know if you're still blocked on the other issue. Maybe you the easiest way to solve it will be to use `Ast_helper` directly to construct the new AST. Of course, it should be the `ppx_tools_versioned` edition of `Ast_helper`.

raphael-proust:
I finally understand (at least I think I do!) the interaction of `ppx_tools` and `Ast_helper`…

With this final push, the location is still not great (it covers the whole expression, including rhs and lhs terms around the deprecated operator), but it's better than the previous attempts I made.

Possible improvements:

- Currently it generates warning is 22 (preprocessor warning). Do we want to report it as warning 3 (deprecation warning)? Note that `Ast_mapper` has no way to specify which warning we want so I'm not sure how easy it is to do.
- Currently, when running the tests, the flag `-w -22` is added which deactivates warning 22. This is necessary because 22 is marked as fatal in the default flags. I don't know if we can mark warnings as non-fatal. tried `-w -22+22` but this had no effect: 22 was still fatal.

raphael-proust:
Fixed the warning 22 issue mentioned above: `-warn-error -22`

src/ppx/ppx_lwt.ml
```diff
@@ -334,16 +334,23 @@ let mapper =
             "Lwt's finally should be used only with the syntax: \"(<expr>)[%%finally ...]\"."
         ))
 
-      | [%expr [%e? lhs] >> [%e? rhs]] ->
+      | [%expr [%e? lhs] >> [%e? rhs]] as e ->
```
aantron: To get the location of the `>>`, I believe you can either:

1. Change this pattern to manually match on `Pexp_apply` of `Pexp_ident ">>"` (I'm skipping the wrapping location and attributes and structures around each `Pexp_*`). You should then be able to get the location from the `Pexp_ident`.
2. Leave this pattern as is, because it's pretty legible. In the case body below, match again on `Pexp_apply` (what `e` is) and extract the location of the function (it should be a `Pexp_ident ">>"` wrapped in location information).

I don't recall immediately how binary operator application is represented, it's likely that there are actually two `Pexp_apply`s, one nested in the other.

aantron:
For the warning number, I think 22 is fine. We could contrive some crazy scheme to get warning 3, where we introduce a dummy local deprecated value in a local module, just so we can refer to it in the generated code and trigger warning 3. But I doubt the complexity is necessary.

raphael-proust:
Thanks for the suggestion. I went for option 2: rematch `e` into `Pexp_apply (op, _)`. I gives accurate position.

I wonder if ppx_tools' metaquot can do that directly. Does it have alias patterns where you give both a literal pattern and a name?

aantron:
Not that I'm aware of. It may be possible to get that effect for some sub-patterns, but off the top of my head, I'm not aware of a nice way to do it for the operation in this particular one.

aantron:
Maybe you can do some kind of `[%e? [%expr ">>"] as op]` pattern, but I would be surprised if it works :p

aantron:
It looks ready for merge to me. Let me know if/when not going to edit the `>>` pattern :)

The warning doesn't seem picked up by Merlin, but that's okay.

Drup:
> The warning doesn't seem picked up by Merlin, but that's okay.

Actually, that's weird, there is no reason that should happen.

aantron:
> Actually, that's weird, there is no reason that should happen.

It's possible my Merlin setup is broken somehow, the editor often doesn't trigger Merlin, etc. However, Merlin was working for other warnings and errors in other expressions in the same session. I'd be interested in the result if you tried it..

raphael-proust:
In `test/ppx/.merlin` there are flags for using the Lwt ppx and for warnings. I'm not entirely sure what these flags do, but it seems that it hard wires the use of the locally built ppx instead of using the `PKG lwt.ppx` option.

Can anyone with more knowledge of how Merlin's config works comment?

aantron:
@raphael-proust That `.merlin` is generated by Jbuilder, which indeed sets it up so that Merlin loads the local PPX. I'm not a Merlin user, but my limited experience does suggest that the generated `.merlin` is pretty much correct.

aantron:
Did you also observe that Merlin doesn't highlight the error?

BTW I don't think that blocks the PR from my point of view, but I'm also fine with waiting if you'd like to look into this :)

raphael-proust:
On my machine, the error is also not highlighted by Merlin.

You can merge the PR (with squashing would be better I think).

aantron:
Thanks @raphael-proust and @Drup!

PR #474, opened 2017-09-27
Add expect tests for ppx

copy:
This adds expect tests for some of the ppx constructs. It uses some code from [ppx_bisect's test helpers](https://github.com/aantron/bisect_ppx/blob/f1b1899609cfa1f77cdd5f0e32d614ed0d6d9db4/test/src/test_helpers.ml). It doesn't work on Windows, due to using `diff`.

cc @Drup @aantron

### DIFF
--- a/test/ppx_expect/main.ml
+++ b/test/ppx_expect/main.ml
@@ -63,9 +63,18 @@ let () =
     |> List.map Filename.chop_extension
   in
   Sys.chdir test_directory;
-  try
-    List.iter run_test test_cases
-  with
-  | Failure f ->
-    prerr_endline f;
-    exit 1
+  let only_if () =
+    Sys.cygwin = false && Sys.win32 = false &&
+    (* 4.02.3 prints file paths differently *)
+    Scanf.sscanf Sys.ocaml_version "%u.%u"
+      (fun major minor -> major >= 4 && minor >= 3)
+  in
+  let suite = Test.suite "ppx_expect" (
+    List.map (fun test_case ->
+      Test.test_direct test_case ~only_if (fun () ->
+        run_test test_case;
+        true
+      )
+    ) test_cases)
+  in
+  Test.run "ppx_expect" [suite]
--- a/test/ppx_expect/main.ml
+++ b/test/ppx_expect/main.ml
@@ -1,0 +1,78 @@
+let test_directory = "cases"
+let test_cases = [
+  "let_1";
+  "let_2";
+  "let_3";
+  "let_4";
+  "match_1";
+  "match_2";
+  "match_3";
+  "match_4";
+  "try_1";
+  "try_2";
+  "try_3";
+]
+
+let _read_file name =
+  let buffer = Buffer.create 4096 in
+  let channel = open_in name in
+
+  try
+    let rec read () =
+      try input_char channel |> Buffer.add_char buffer; read ()
+      with End_of_file -> ()
+    in
+    read ();
+    close_in channel;
+
+    Buffer.contents buffer
+
+  with exn ->
+    close_in_noerr channel;
+    raise exn
+
+let _command_failed ?status command =
+  match status with
+  | None -> Printf.sprintf "'%s' did not exit" command |> failwith
+  | Some v -> Printf.sprintf "'%s' failed with status %i" command v |> failwith
+
+let _run_int command =
+  match Unix.system command with
+  | Unix.WEXITED v -> v
+  | _ -> _command_failed command
+
+let run command =
+  let v = _run_int command in
+  if v <> 0 then _command_failed command ~status:v
+
+let diff reference result =
+  let command = Printf.sprintf "diff -au %s %s" reference result in
+  let status = _run_int (command ^ " > /dev/null") in
+  match status with
+  | 0 -> ()
+  | 1 ->
+    let _ : int =_run_int (command ^ " > delta") in
+    let delta = _read_file "delta" in
+    Printf.sprintf "> %s:\n\n%s" command delta
+    |> failwith
+  | _ -> _command_failed command ~status
+
+let run_test name =
+  let ml_name = name ^ ".ml" in
+  let expect_name = name ^ ".expect" in
+  let fixed_name = name ^ ".fixed" in
+  let command = Printf.sprintf "ocaml %s 2> %s" ml_name fixed_name in
+  let ocaml_return_code = _run_int command in
+  begin if ocaml_return_code = 0 then
+      failwith (Printf.sprintf "Unexpected toplevel return code: %d" ocaml_return_code)
+  end;
+  diff expect_name fixed_name
+
+let () =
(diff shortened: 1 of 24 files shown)


Drup:
The tests look good!

test/ppx_expect/main.ml
```diff
@@ -0,0 +1,78 @@
+let test_directory = "cases"
+let test_cases = [
```
aantron: It's not necessary, but I think it would be nice to get the test list by listing the directory.

aantron:
Oh, and thanks for being so thorough as to make these tests and PR them :)

copy:
> I would expect it to still work on Windows, because we are testing under Cygwin, so diff is available.

Good point. The diffing seems to work, but it looks like there's a problem with the ppx extension on the Windows CI:

```
# The system cannot find the path specified.
# > diff -au let_1.expect let_1.fixed:
# 
# --- let_1.expect	2017-09-27 13:49:32.561606300 +0000
# +++ let_1.fixed	2017-09-27 13:49:39.080199000 +0000
# @@ -1,3 +1,2 @@
# -File "./let_1.ml", line 6, characters 10-12:
# -Error: This pattern matches values of type unit
# -       but a pattern was expected which matches values of type int
# +File ".\let_1.ml", line 6, characters 6-9:
# +Uninterpreted extension 'lwt'.
# 
```

Any idea what this could be about?

aantron:
I haven't looked into it deeply at all, but given that the existing PPX test suite runs in Windows CI ([example](https://ci.appveyor.com/project/aantron/lwt/build/673/job/osi6fffe36ijc1tr#L1589)), I would guess that there is a problem with the details of how the expect tests are invoking the PPX.

aantron:
And looking at it even closer, it looks like it might be problem with packaging of `lwt.ppx`, since the package is found by `topfind`, but that doesn't result in the PPX being loaded. I'm not sure if that's inherent to the environment the toplevel ends up running in. I suppose it isn't, because it works on Unix.

I would have to RDP into the build worker to debug this. Would you be willing to do that, or debug it on a Windows machine you have access to?

copy:
I don't think I have the expertise with Windows to look into this.

There also seems to be a problem on 4.02.3, which is in how the compiler prints paths:

```
-File "./let_1.ml", line 6, characters 10-12:
+File "let_1.ml", line 6, characters 10-12:
```

I'd suggest to skip these tests on 4.02.3, but alternatively we could also sanitise the output.

aantron:
It might be more productive to run ocamlfind explicitly instead of using the toplevel.

I'm fine with disabling the tests anywhere they don't work for now, though. I think we only really need them to work in one build row. There shouldn't be any differences with other build rows anyway, at least any difference caused by Lwt, and having them work *anywhere* is already an improvement over not having them at all.

aantron:
Well, to be more precise, there shouldn't be any difference due to Lwt between systems for one compiler version. For different versions, though, the compiler *could* give us different locations to work with, that we then use for the generated locations. But this isn't that likely. I guess we won't have any testing on 4.02, but that's okay if we can't easily work around the output issue. If you'd like to sanitize the output, it would be welcome. Not testing on Windows is completely fine.

aantron:
Also, for skipping the test on Windows, the best recommendation I can give is to use Lwt's little testing framework, and wrap each `.ml` file in a single generated test case, and combine all those into a test suite. Use the [`~only_if`](https://github.com/ocsigen/lwt/blob/fa4fa370aacdfe0f6f86bb3716c0bac0cbe3f28d/test/test.mli#L30) optional argument to skip each case on Windows. This will make it clear in the output that the tests were skipped, and the output will be consistent with other skipped tests elsewhere in Lwt's test suites.

Here is some similar code:

https://github.com/ocsigen/lwt/blob/d9ed54c1e47a48bcfc0414105cfc9026a4c8c274/test/unix/test_mcast.ml#L51-L55

copy:
Okay, I updated the test to use Lwt's test suite (which makes more sense anyway), and to skip the test on Windows, cywin, and OCaml < 4.03.

Let me know if I should squash these commits together.

test/ppx_expect/main.ml
```diff
@@ -63,9 +63,18 @@ let () =
...
-    List.iter run_test test_cases
-  with
-  | Failure f ->
-    prerr_endline f;
-    exit 1
+  let only_if () =
+    Sys.cygwin = false && Sys.win32 = false &&
+    (* 4.02.3 prints file paths differently *)
+    Scanf.sscanf Sys.ocaml_version "%u.%u"
+      (fun major minor -> major >= 4 && minor >= 3)
```
aantron: I think 5.00 will wrongly fail this test.

copy: Good point. Note that we'll have a similiar problem in https://github.com/ocsigen/lwt/blob/b52e2115c4b42136fa9480939135ababa219631b/src/util/configure.ml#L100

aantron: Good catch, and thanks for noting. I'll push to `master` to fix that.

aantron:
Thanks for doing this @copy, and thanks @Drup for helping move this along and review it here and in #337.

aantron:
The tests can't find `lwt.ppx` when `lwt` isn't installed through opam (which is typical for me during development). In other words, they have been running against an installed Lwt, not against the Lwt in the working tree. I think the "right" way to write the tests is as a bunch of small Jbuilder-ized programs, so that Jbuilder can find `lwt.ppx` for them in the working tree. For comparison, see these little programs here: https://github.com/ocsigen/lwt/tree/master/test/packaging/jbuilder

Output looks like this:

```
(8/11) Running test "match_4" from suite "ppx_expect"
No such package: lwt
No such package: lwt.ppx
Test "match_4" from suite "ppx_expect" failed. It raised: "Failure(\"> diff -au match_4.expect match_4.fixed:\\n\\n--- match_4.expect\\t2017-10-14 07:39:15.000000000 -0500\\n+++ match_4.fixed\\t2017-10-14 07:39:17.000000000 -0500\\n@@ -1,4 +1,2 @@\\n-File \\\"./match_4.ml\\\", line 12, characters 4-16:\\n-Error: This expression has type int Lwt.t\\n-       but an expression was expected of type unit Lwt.t\\n-       Type int is not compatible with type unit \\n+File \\\"./match_4.ml\\\", line 6, characters 8-11:\\n+Error: Uninterpreted extension 'lwt'.\\n\")".
```

We should probably change `travish.sh` so that it doesn't install Lwt before running the tests, if possible, to catch such conditions.

aantron:
...aaaand I'll work on it :)

aantron:
I worked on it in branch [`ppx-test-install-ordering`](https://github.com/ocsigen/lwt/commits/ppx-test-install-ordering) last week, but eventually moved on to other things after reading a ton of Findlib code.

In the meantime, I disabled the PPX tests temporarily. See the commit linked above.

PR #479, opened 2017-10-08
Make Lwt_ssl -safe-string-compatible

### LINKED ISSUE #478: Lwt_ssl is not safe-string compatible
As a result, it doesn't build with 4.06 defaults: https://travis-ci.org/ocsigen/lwt/jobs/284301791#L1389-L1391

aantron:
This makes `Lwt_ssl` 4.06-friendly by changing `Lwt_ssl.read` and `Lwt_ssl.write` so that they operate on buffers of type `bytes` rather than `string`.

This could be a breaking change for some code. It is, however, consistent with `Lwt_unix`. `Lwt_unix` also has a `write_string` function for writing buffers of type `string`. We may want to add that to `Lwt_ssl` in a separate PR.

As I'm not a user of `Lwt_ssl`, I'd appreciate feedback on this change.

EDIT: This fixes #478.

### DIFF
--- a/src/ssl/lwt_ssl.mli
+++ b/src/ssl/lwt_ssl.mli
@@ -58,8 +58,8 @@ val ssl_perform_handshake : uninitialized_socket -> socket Lwt.t
 val ssl_accept_handshake  : uninitialized_socket -> socket Lwt.t
 (** Await a SSL/TLS handshake on the specified socket (used by servers). *)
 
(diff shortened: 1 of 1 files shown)


aantron:
I'm proceeding with this. Searching opam and GitHub, I haven't found any code that would be broken by this, that wouldn't be broken by 4.06's `-safe-string` default anyway.

Issue #338, opened 2017-04-12
Move the PPX to a separate package (?)


aantron:
I'm not sure if it ought to also be ported to `ocaml-migrate-parsetree`. If so, then port it.

IMO the PPX syntax is much easier to teach than `>|= fun blah ->` everywhere, so I've been writing new docs in terms of it. However, we can just tell newcomers to do `opam install lwt lwt_ppx`. Projects that avoid the PPX for whatever reason can do `opam install lwt` and not have to wait for ports of `ppx_tools`/`ocaml-migrate-parsetree` to new versions of OCaml.

On the other hand, maybe we should just make an effort to make sure that those ports don't take a long time. Also, `ppx_tools` and `ocaml-migrate-parsetree` are such basic packages, that installing several OPAM packages is likely to pull them in whether they are dependencies of the base `lwt` or not.

Please let me know any thoughts on this.

Prompted by @avsm in https://github.com/ocaml/opam-repository/pull/8929#issuecomment-293622509.

EDIT: and there is a previous similar (but not identical) request in #266.

Drup:
Splitting the ppx inside its own package and porting it to ocaml-migrate-parsetree are two completely independent tasks (and would not be done for the same reasons).

I have no opinion on the first. Doing the second while we have oasis needs far too many hacks, and it would be easier to just port lwt to something else.

avsm:
Agree with @drup -- just moving the package out would be a good first start, since a significant number of packages do not use the ppx extension at all, and it's a pity when the core Lwt package fails to build when ppx is not available.

My request is specifically motivated by the desire to do more CI testing with development branches of OCaml, and this is the most common reason why ppx is unavailable (due to an AST change).  Lwt blocks a very large number of packages from building, but a significant percentage of those do not have a hard ppx dependency.

Issue #370, opened 2017-05-22
Move Camlp4 syntax to a separate package


aantron:
To eliminate a depopt from `opam`, and as part of deprecating Camlp4. Related #338.

aantron:
I tried doing something like #481 (factoring out the PPX) to Camlp4, but it doesn't seem to work. I think `archive(syntax ...) ...` predicates have to be directly in at least one package that is mentioned on the command line, not only findable through `requires` predicates. So, we can't use a "proxy" package, like for the PPX. This seems like a bug in Findlib, but I don't think we have time to wait for a fix.

I can think of two or three ways to solve this:

1. Just move the Camlp4 syntax out, without going through a proxy package phase. We would do this in 4.0.0, and cause breakage.
2. Duplicate the Camlp4 syntax both `lwt.syntax` and `lwt_camlp4`, until 4.0.0. This seems equally as easy/difficult as (1), but would be "softer." I don't know if there would be any difficulties I don't foresee, though.
3. Maybe it's possible to create a dummy library that has an `archive(syntax, ...)` predicate, and that will satisfy Findlib. I haven't tried it.

@Drup, any suggestions from your experience?

Drup:
I think it's okay, in this case, to do a one time move. Does the camlp4 syntax extension still has many users ?

aantron:
Yeah, I see `aws`, maybe `biocaml`, maybe `bitcoin`, `eliom`, `facebook-sdk`, maybe `hardcaml`, tests of `imaplet`, `iocaml`, `jitsu`, `joolog`, some examples in `js_of_ocaml`, `lambdoc` examples, `mongo`, `ojquery`, `ojwidgets`, `rdf`, `release`, `riak`, `themoviedb`, `usb`, `xe`, `xen-disk`, `xenctrl`, `xentropyd` just in opam. Didn't look into how many of those are presently maintained.

aantron:
The linked commit duplicates Camlp4 in a `lwt_camlp4` package.

aantron:
I should note that this approach results in Findlib warnings about duplicate interfaces in `lwt` and `lwt_camlp4`. These are temporary until 4.0.0.

PR #481, opened 2017-10-13
Begin factoring out the PPX

aantron:
Discussed in #338.

This is pretty much following our "normal" procedure:

1. This commit creates an opam package `lwt_ppx`, containing a Findlib package `lwt_ppx`, which is a proxy for the current `lwt.ppx` (which is in opam package `lwt`). We will release the commit as part of 3.2.0, and announce the upcoming change.
2. In 4.0.0, we will delete `lwt.ppx` and move the PPX fully to `lwt_ppx`.

There is one flaw: even when using `lwt_ppx`, PPX options still have to be passed to `lwt.ppx`, not to `lwt_ppx`. Once 4.0.0 lands, they will suddenly have to be passed to `lwt_ppx`. This applies only to ocamlfind and tools that use it (like Ocamlbuild). Jbuilder, or any other build tool that runs the PPX as a library linked into a driver, will handle options correctly. I looked through opam users of Lwt, and only sqlexpr seems to use `lwt.ppx` options, in its `README` and a test (cc @mfp). So, we can hope that this breakage in 4.0.0 won't be too bad.

### DIFF
--- a/META.lwt
+++ b/META.lwt
@@ -28,7 +28,7 @@ package "log" (
 package "ppx" (
   #directory = "ppx"
   version = "dev"
-  description = "Lwt PPX syntax extension"
+  description = "Lwt PPX syntax extension (deprecated; use lwt_ppx)"
   requires(ppx_driver) = "lwt.omp"
   requires(-ppx_driver) = "bytes lwt result"
   ppx(-ppx_driver,-custom_ppx) = "./ppx.exe --as-ppx"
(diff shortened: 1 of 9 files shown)


aantron:
I guess we could try duplicating the PPX in both `lwt.ppx` and `lwt_ppx` to avoid the option problem, like proposed in https://github.com/ocsigen/lwt/issues/370#issuecomment-336512965. I don't think duplicating PPXs is as harmful as duplicating libraries, though I'm not sure what will happen if there is a PPX driver being used, and one project somehow pulls both `lwt.ppx` and `lwt_ppx` in. I'd imagine that's very rare.

Drup:
Applying a ppx twice shouldn't be an issue, it should be idempotent.

How do you handle ppx.log ?

aantron:
There is no `ppx.log`, I think only `lwt.syntax.log` for the Camlp4 extension. Not yet sure what to do with Camlp4 exactly, see https://github.com/ocsigen/lwt/issues/370#issuecomment-336512965.

PR #483, opened 2017-10-14
Support disposal of Lwt_pool.t elements

### LINKED ISSUE #480: Request: function to clear/destroy Lwt_pool
#183 already asked for this but was closed, so creating a new issue: will we be able to have a function to clear the pool?

My use case: I'm maintaining a pool of [postgresql-ocaml](https://github.com/mmottl/postgresql-ocaml) connections in my app. It's a wrapper of [libpq](https://www.postgresql.org/docs/9.6/static/libpq.html), the C client library to postgres. The best usage practice of this library is that we must call a `finish` function on a connection when it's no longer needed to free the allocated memory. I want to be able to call `finish` on all connections in the pool when my app shuts down to make sure all connections are properly closed.

Through `validate` I am only able to check if a connection is bad which will then be reconnected.

An API that I have in mind is by providing an optional `?destroy:('a -> unit Lwt.t)` argument to `Lwt_pool.create` and the public API of `Lwt_pool.clear pool` which will call destroy on all pooled resource. Is it possible? Would love to give it a shot.

(Btw, Lwt_pool is awesome, seems like so far it's the only public implementation of resource pool in OCaml?)

hcarty:
An initial attempt at solving #480.  Also tried to catch occurrences of `thread` in the doc comments and replace them with `promise`.

EDIT by @aantron: resolves #480.

### DIFF
--- a/src/core/lwt_pool.ml
+++ b/src/core/lwt_pool.ml
@@ -34,6 +34,8 @@ type 'a t = {
   (* Check a member when its use failed. *)
   validate : 'a -> bool Lwt.t;
   (* Validate old pool members. *)
+  dispose : 'a -> unit Lwt.t;
+  (* Dispose of a pool member. *)
   max : int;
   (* Size of the pool. *)
   mutable count : int;
@@ -44,11 +46,12 @@ type 'a t = {
   (* Threads waiting for a member. *)
 }
 
-let create m ?(check = fun _ f -> f true) ?(validate = fun _ -> Lwt.return_true) create =
+let create m ?(check = fun _ f -> f true) ?(validate = fun _ -> Lwt.return_true) ?(dispose = fun _ -> Lwt.return_unit) create =
   { max = m;
     create = create;
     validate = validate;
     check = check;
+    dispose = dispose;
     count = 0;
     list = Queue.create ();
     waiters = Lwt_sequence.create () }
@@ -104,10 +107,12 @@ let check_elt p c =
         | false ->
           (* Remove this member and create a new one. *)
           p.count <- p.count - 1;
+          p.dispose c >>= fun () ->
           create_member p)
       (fun e ->
          (* Validation failed: create a new member if at least one
             thread is waiting. *)
+         p.dispose c >>= fun () ->
          replace_acquired p;
          Lwt.fail e)
 
@@ -127,12 +132,19 @@ let acquire p =
 
 (* Release a member when its use failed. *)
 let checked_release p c =
-  p.check c begin fun ok ->
-    if ok then
-      release p c
-    else
-      replace_acquired p
-  end
+  let ok = ref false in
+  p.check c (fun result -> ok := result);
+  if !ok then (
+    (* Element is ok - release it back to the pool *)
+    release p c;
+    Lwt.return_unit
+  )
+  else (
+    (* Element is not ok - dispose of it and replace with a new one *)
+    p.dispose c >>= fun () ->
+    replace_acquired p;
+    Lwt.return_unit
+  )
 
 let use p f =
   acquire p >>= fun c ->
@@ -143,5 +155,10 @@ let use p f =
        release p c;
        t)
     (fun e ->
-       checked_release p c;
+       checked_release p c >>= fun () ->
        Lwt.fail e)
+
+let clear p =
+  let elements = Queue.fold (fun l element -> element :: l) [] p.list in
+  Queue.clear p.list;
+  Lwt_list.iter_s p.dispose elements
--- a/src/core/lwt_pool.ml
+++ b/src/core/lwt_pool.ml
@@ -36,7 +36,7 @@ type 'a t = {
   (* Validate old pool members. *)
   dispose : 'a -> unit Lwt.t;
   (* Dispose of a pool member. *)
-  mutable cleared : bool ref;
+  cleared : bool ref ref;
   (* Have the current pool elements been cleared out? *)
   max : int;
   (* Size of the pool. *)
@@ -54,7 +54,7 @@ let create m ?(validate = fun _ -> Lwt.return_true) ?(check = fun _ f -> f true)
     validate = validate;
     check = check;
     dispose = dispose;
-    cleared = ref false;
+    cleared = ref (ref false);
     count = 0;
     list = Queue.create ();
     waiters = Lwt_sequence.create () }
@@ -81,13 +81,19 @@ let release p c =
     (* No one is waiting, queue it. *)
     Queue.push c p.list
 
+(* Dispose of a pool member *)
+let dispose p c =
+  p.dispose c >>= fun () ->
+  p.count <- p.count - 1;
+  Lwt.return_unit
+
 (* Create a new member when one is thrown away. *)
 let replace_acquired p =
   match Lwt_sequence.take_opt_l p.waiters with
   | None ->
     (* No one is waiting, do not create a new member to avoid
        loosing an error if creation fails. *)
-    p.count <- p.count - 1
+    ()
   | Some wakener ->
     Lwt.on_any
       (Lwt.apply p.create ())
@@ -95,7 +101,6 @@ let replace_acquired p =
          Lwt.wakeup_later wakener c)
       (fun exn ->
          (* Creation failed, notify the waiter of the failure. *)
-         p.count <- p.count - 1;
          Lwt.wakeup_later_exn wakener exn)
 
 (* Verify a member is still valid before releasing it; try to replace
@@ -109,13 +114,12 @@ let check_elt p c =
           Lwt.return c
         | false ->
           (* Remove this member and create a new one. *)
-          p.count <- p.count - 1;
-          p.dispose c >>= fun () ->
+          dispose p c >>= fun () ->
           create_member p)
       (fun e ->
          (* Validation failed: create a new member if at least one
             thread is waiting. *)
-         p.dispose c >>= fun () ->
+         dispose p c >>= fun () ->
          replace_acquired p;
          Lwt.fail e)
 
@@ -139,7 +143,7 @@ let checked_release p c cleared =
   p.check c (fun result -> ok := result);
   if cleared || not !ok then (
     (* Element is not ok or the pool was cleared - dispose of it *)
-    p.dispose c
+    dispose p c
   )
   else (
     (* Element is ok - release it back to the pool *)
@@ -151,28 +155,30 @@ let use p f =
   acquire p >>= fun c ->
   (* Capture the current cleared state so we can see if it changes while this
      element is in use *)
-  let cleared = p.cleared in
-  Lwt.catch
-    (fun () ->
-       let t = f c in
-       t >>= fun _ ->
-       if !cleared then (
-         (* p was cleared while t was resolving - dispose of this element *)
-         p.dispose c >>= fun () ->
-         t
-       )
-       else (
-         release p c;
-         t
-       )
-    )
-    (fun e ->
-       checked_release p c !cleared >>= fun () ->
-       Lwt.fail e)
+  let cleared = !(p.cleared) in
+  let promise =
+    Lwt.catch
+      (fun () -> f c)
+      (fun e ->
+         checked_release p c !cleared >>= fun () ->
+         Lwt.fail e)
+  in
+  promise >>= fun _ ->
+  if !cleared then (
+    (* p was cleared while promise was resolving - dispose of this element *)
+    dispose p c >>= fun () ->
+    promise
+  )
+  else (
+    release p c;
+    promise
+  )
 
 let clear p =
   let elements = Queue.fold (fun l element -> element :: l) [] p.list in
   Queue.clear p.list;
-  p.cleared := true;
-  p.cleared <- ref false;
-  Lwt_list.iter_s p.dispose elements
+  (* Indicate to any currently in-use elements that we cleared the pool *)
+  let old_cleared = !(p.cleared) in
+  old_cleared := true;
+  p.cleared := ref false;
+  Lwt_list.iter_s (dispose p) elements
--- a/src/core/lwt_pool.ml
+++ b/src/core/lwt_pool.ml
@@ -39,7 +39,7 @@ type 'a t = {
   max : int;
   (* Size of the pool. *)
   mutable count : int;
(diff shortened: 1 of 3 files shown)


hcarty:
This implementation of `clear` does not invalidate pool elements which are currently in use once they have been released. I'm not sure if that's ok. I'd like to have clear invalidate all pool elements. One way to do that would be to wrap each element in a record with a mutable `disposed` field. If the field is set appropriately the element would be considered invalid without any further checks.

I'll hopefully have a chance early next week to update this PR with that improved functionality unless someone disagrees with the approach.

src/core/lwt_pool.mli
```diff
@@ -27,31 +27,45 @@
...
+      and then re-created with [f].
 
-      The optional function [check] is called after a [use] of an
-      element failed. It must call its argument exactly once with
-      [true] if the element is still valid and [false] otherwise. *)
+      If a call to {!use} fails with a pool element that element will be passed
+      to the optional function [check] as [check element callback].  [check]
+      must call [callback] exactly once with [true] if [element] is still valid
+      and [false] otherwise.  If [check] calls [callback false] then [dispose]
+      will be run on [element].
```
bobbypriambodo: I'm a bit confused on when `dispose` is called. At a glance, it seems that both `validate` and `check` will call `dispose`. Will `dispose` be called twice? If not, how is the order of calling done?

Also this:

> If a call to {!use} fails with a pool element

As my comment in the issue, I think it's unclear why a `use` would fail. Does "fail" equals "trying to use an invalid pool element" (due to `validate`)?

aantron: I believe "fail" means raising an exception or rejecting the `'b Lwt.t` promise that `use`'s callback returns. It should definitely be clarified in the comment, because I also am guessing about what it means.

aantron:
> This implementation of `clear` does not invalidate pool elements which are currently in use

@bobbypriambodo Can you comment on this? What should `clear` do for pool elements that are currently in use?

bobbypriambodo:
> What should `clear` do for pool elements that are currently in use?

Intuitively for me `clear` should dispose _all_ elements. However, I'm not sure what the best way is to achieve it.

The way [node-pool](https://github.com/coopernurse/node-pool) achieve this is to first call `drain()` which will make sure that new `acquire/use` calls will be disallowed when the state is "draining" while letting remaining in-use elements to finish, and resolves when all elements are returned to the pool, at which we could call `clear()` safely. It also hits me that when we call `clear`, we might want to prevent new creations of elements, or else there might be some race conditions involved.

Or, another approach that I thought of is that to leave it as it is, and let the user bear the responsibility of making sure that the pool is definitely not used anymore before clearing. In case of a pool of DB connections that I have in mind, `clear` would be called when the app is shutting down, and it makes sense to first make sure that the app is not accessed before shutting it down (e.g. putting it out of load balancers, mark the instance unhealthy, etc.) to prevent end-users receive all kinds of errors.

src/core/lwt_pool.mli
bobbypriambodo: Ah, that actually makes sense. I think it's the whole `validate` and `invalid elements` on `use` paragraph just above it that confused me, I thought they are still talking about the same (or at least related) thing.

aantron:
@bobbypriambodo thanks for your thoughts!

I guess that goes back to the question of whether we want a minimalistic `Lwt_pool` that is "agnostic" to different ways to address this, or whether there is a common behavior (which I don't know) for this aspect that users tend to expect, and we should implement that behavior.

hcarty:
I agree with @bobbypriambodo - I think it's reasonable to expect `clear` to invalidate and dispose of any existing pool elements once they are released.

I don't want to block usage of the pool on outstanding element use after a `clear`- when I get back to this I'll look for a clean way to handle that too.

hcarty:
And suggestions for approaches are of course welcome!

src/core/lwt_pool.mli
hcarty: I tried to tweak the wording at bit - does this make the use of each argument any clearer?  Suggestions are welcome - the wording has been tricky to get right.

bobbypriambodo:
@hcarty really love the new doc! It significantly clear things up. Now I understand the differences between `check` and `validate`.

Another minor note, there's an inconsistency, you defined `[check element is_ok]` but later `If [check] calls [callback false]`... If I understand correctly `callback` should be `is_ok`.

src/core/lwt_pool.mli
```diff
@@ -27,31 +27,45 @@
     The pool also provides a limit on the number of connections that can
     simultaneously be open.
 
-    Note that pools are not for keeping Lwt threads. Lwt threads are very cheap
-    to create and are pure. It is neither desirable nor possible to reuse them.
-    If you want to have a pool of {e system} threads, consider module
+    Note that pools are not for keeping Lwt promises. Lwt promises are very
```
aantron: I think this paragraph is a fortunate casualty of the change in terminology from threads to promises – as in, I doubt anyone would think of keeping a "promise pool" anyway :) I suggest deleting most of this paragraph, and leaving some modified version of the last sentence about using `Lwt_preemptive` for a system thread pool.

src/core/lwt_pool.mli
```diff
@@ -27,31 +27,45 @@
...
     [Lwt_preemptive]. *)
 
 type 'a t
-  (** Pools containing values of type ['a]. *)
+  (** Pools containing elements of type ['a]. *)
 
 val create :
   int ->
   ?check : ('a -> (bool -> unit) -> unit) ->
   ?validate : ('a -> bool Lwt.t) ->
```
aantron: Optional, but maybe we should also change the order so that `?validate` comes before `?check`? On each call to `use`, `?validate` is called first, *then* `?check` if the promise returned by `?use` is rejected.

src/core/lwt_pool.mli
aantron: The new docs are pretty great!

hcarty:
Thanks for the additional comments. I should get to this later this week, along with a potentially better approach to `clear`.

hcarty:
When `check` returns false the current approach is to drop the existing pool element (and with this PR, `dispose` it) and then create a new one to take its place.  Would it be ok to change that functionality to *not* replace the element in the pool immediately after `check` returns `false`?  If a new element is required to take its place it would still be created during a subsequent `use`.  My reasoning behind this change is:

1. Creating a new pool element may be costly.  It seems better to save that cost until the element is required rather that paying it in the promise where it failed.
2. The current remake-on-`check`-failure logic makes `clear`'s semantics more difficult to understand as `clear` could be called on a pool but an outstanding `use` promise could end up creating a new pool element if it resolves to an exception.

hcarty:
I think the feedback on the doc comments have been addressed in the latest commit.

I've also updated `clear` so that any in-use pool elements will be disposed of once they are released.  The logic there uses some ugly state tracking but hopefully the implementation is understandable.  Please ask questions if it isn't.

Finally, I changed the logic around `check` so that a failed `check` does NOT create a new pool element.  I think that this is better behaviour in general - the only place we should create a new pool member is when we try to acquire one.  This also makes the `clear` semantics and implementation much simpler.

aantron:
> Would it be ok to change that functionality to *not* replace the element in the pool immediately after `check` returns `false`?

As maintainer, changing semantics makes me feel extremely paranoid, but I think this is ok. It may affect element creationg timing for some users, but given Lwt is already a concurrency library, I would imagine nobody is able to count on the timing anyway.

> in the promise where it failed

Sorry to nitpick language, but I have to do this as part of putting the thread terminology to rest. It's misleading to say that anything happens *in* a promise (compare "*in* a thread," which makes sense, but Lwt is no longer called a threading library exactly because that *doesn't* make sense for Lwt). The only things that can be said to be *in* a promise are either a list of callbacks (when pending) or a value or exception (when resolved). A more right thing to say here is something like "rather than paying it right after a promise returned from `use` is rejected," or anything else like that.

src/core/lwt_pool.ml
```diff
@@ -36,6 +36,8 @@ type 'a t = {
   (* Validate old pool members. *)
   dispose : 'a -> unit Lwt.t;
   (* Dispose of a pool member. *)
+  mutable cleared : bool ref;
```
aantron: The amount of indirection here is a bit sneaky, and the below syntax

```ocaml
p.cleared := true;
p.cleared <- ref false;
```

was mysterious to me when I first saw it. I'd suggest making this an "immutable" field of type `bool ref ref`.

aantron:
The docs look pretty good. For me, I'll wait on reviewing the code in detail, as the build is currently failing. I think it's just the order of the labeled arguments that needs updating in the implementation.

hcarty:
> A more right thing to say here is something like "rather than paying it right after a promise returned by the callback of use is rejected," or anything else like that.

I'm uncertain here - from a user's perspective it's `use` which pays the price of creating the new pool element.  `'a Lwt.t` may not be a thread but it still has a context it's responsible for, including its lifetime.  I need to read your updated introduction again to make sure I have my terminology correct!

hcarty:
> I'll wait on reviewing the code in detail, as the build is currently failing. I think it's just the order of the labeled arguments that needs updating in the implementation.

Oh no!  Sorry about that, I pushed too soon.  I'll get a fix in.

aantron:
> from a user's perspective it's use which pays the price of creating the new pool element.

Quite true :)

> `'a Lwt.t` may not be a thread but it still has a context it's responsible for, including its lifetime.

In a very loose sense there is a context, but it's a quite different kind of context than what is associated with a thread. That context is exactly a list of callbacks only (for a pending promise), and the promise is not "responsible" for it – it just has a list. I want to exactly avoid the notion that Lwt promises are similar to threads, including having contexts similar in any way one would ordinarily expect, so I'd rather avoid referring to both as "contexts" as well, as it's a loose use of terminology that could be confusing. The point is, promises are not "containers" for sequences of instructions, not units of execution, not something that runs, not something that can be interrupted, nothing "happens" in them. They are just smart references, that can hold lists of callbacks. Something *else* (a real thread, the main thread of the Lwt program) is responsible for running those callbacks at the right time, that main thread has a genuine thread context, etc. Promises are just inert data objects that are useful for synchronization and thinking about concurrency, and for getting a real thread to run things in a nice order.

aantron:
Or, put another way, promises are similar to streams in FRP, except they are one-shot. They are some kind of objects used for factoring your concurrent program, and you can register functions on them to create some kind of promise "network." Of course, I can't write that in the Lwt introduction :p

hcarty:
Working on that test failure now

hcarty:
Sorry for the noise from that failing test!

Latest changes:

1. `cleared` is now a `bool ref ref` now as suggested by @aantron.
2. Add an internal `dispose` helper function to help ensure `count` stays synced up
3. Make the tests pass and add to the `clear` test to make sure even currently-in-use pool elements are disposed of once they are returned.
4. Lift the disposal logic out of `Lwt.catch` in `use` to avoid potential double application of `dispose` if `dispose` fails.

aantron:
Thanks @hcarty, @bobbypriambodo!

Me and @hcarty decided to merge this, skipping a super-thorough nitpicky review for now, but pushing that review to an overhaul (as recently done with `lwt.ml`/`lwt.mli`). It is to be done before 3.2.0. This is because much of the mental work is the same between reviewing this and overhauling `Lwt_pool`, we have to untangle some issues in the pre-existing code, and there is no point in doing this work twice, and forgetting the results after the first time. Since `dispose`/`clear` will not have been released yet, we can freely address any further issues in `dispose`/`clear` as part of the overhaul.

Any comments or reviews in this PR are still welcome, though!

aantron:
> Finally, I changed the logic around check so that a failed check does NOT create a new pool element. I think that this is better behavior in general - the only place we should create a new pool member is when we try to acquire one. This also makes the clear semantics and implementation much simpler.

@hcarty, is it correct to say that this logic is still *in* the code after this PR? I do think I see it there..

aantron:
Never mind, sorry. This old comment

https://github.com/ocsigen/lwt/blob/2b8e53115fd40cf69563bc4f48f02fa80016ca3a/src/core/lwt_pool.ml#L106-L108

is just 100% misleading. `check_elt` is called around `validate` (before `use`), not `check` (release after failed `use`).

aantron:
@hcarty, @bobbypriambodo what do you think about creating elements on an as-needed basis only, i.e. never re-creating inside either `validate` or `check`? We would create them only as a direct consequence of calling `use`.

I guess this would change the semantics even more, as the pool wouldn't try at all to maintain the number of elements it grew to. If there are five concurrent users, the pool will tend to stay around five elements in size. If the number of concurrent users decreases to two, and elements sometimes die in `validate` or `check`, the pool will tend to decrease to two elements automatically.

I don't know if there is some wisdom behind the current semantics. I was not able to find any documented anywhere. @bobbypriambodo, do you know what is done in popular pool implementations in other languages?

aantron:
This would also mean that if a pool element fails `validate`, we would start consuming pool elements in a loop until we find one that passes, or create a new one. The current behavior on failed `validate` is to leave the tail of the queue alone and create a new element right away. I'm not sure if one behavior is inherently more legitimate than the other.

hcarty:
@aantron My preference is to only create on `use`.  Creating a new element can fail.  From a pool user's point of view that failure would make the most sense when there is a need for a new element.

aantron:
Agreed. `validate` is still "on use," but I think only `use` itself should trigger creation.

As an example of weird behavior triggered by the current effort to maintain the element count, In the current code, if `validate` raises an exception, the pool checks for a *waiting other promise*, then tries to create an element, and fulfills/rejects that *other* promise depending on whether creation succeeded or also raised an exception. This was not documented, I wouldn't like to document, and it suggests to me that even if the right thing to do is to create elements eagerly, it shouldn't be done/factored how it is now.

aantron:
In practice, the above means that if `validate` raises an exception, `Lwt_pool` only tries to create a new element if the pool is already still at the max limit (since that's why there is a waiting promise).

bobbypriambodo:
> do you know what is done in popular pool implementations in other languages?

Sorry but at the moment I have no idea, I was more of a user than implementer and most of the libs I used just worked so I haven't got the chance to dig into the impls. I can look it up though, and perhaps post it into the Lwt_pool overhaul issue.

Issue #468, opened 2017-09-05
Lwt_log: how can we use it?


dannywillems:
Are there simple examples to use `Lwt_log` to write logs? Do we need to initialize something before using `Lwt_log.info "Hello, world!"`?

Is it recommended to use it to log lots of information (to `stdout` and in files)?

aantron:
@dannywillems, try the code in this post: https://discuss.ocaml.org/t/about-lwt-concurrency-library/150/6:

```ocaml
let () =
  Lwt_log_core.default :=
    Lwt_log.channel
      ~template:"$(date).$(milliseconds) [$(level)] $(message)"
      ~close_mode:`Keep
      ~channel:Lwt_io.stdout
      ();

  Lwt_log_core.add_rule "*" Lwt_log_core.Info;

  Lwt_main.run begin
    Lwt_log_core.info "blah"
  end

(* ocamlfind opt -linkpkg -package lwt.unix foo.ml && ./a.out *)
```

As per that post, you would normally use module `Lwt_log`, not `Lwt_log_core` everywhere. I was just being explicit about what actual underlying module everything was coming from.

aantron:
Also, even if this answers your question, let's leave the issue open to remind about improving the docs.

rgrinberg:
@dannywillems Have a look at https://github.com/dbuenzli/logs for a logging solution. It has a better API for logging and isn't Lwt specific. Lwt_log should be deprecated IMO.

dannywillems:
Thanks @rgrinberg and @aantron. Sorry to answer now.

aantron:
No worries, most open source takes place at the rate at which people happen to by luck be available :)

aantron:
I'm in favor of deprecating `Lwt_log` in principle, as we evidently don't have the labor resources to properly maintain it. We would suggest people use `logs`, in particular Logs_lwt`.

I found the docs of `Logs_lwt` also a bit nebulous last December, so maybe they ought to be improved as well. Not sure what their current status is, and I don't remember the specific difficulties I was having. I did end up using it successfully, though.

@rgrinberg, any thoughts on what to do about `Lwt_log` in terms of packaging? Perhaps we could factor it out into its own package, and then constrain that package with an upper bound when `Lwt_log` becomes incompatible with Lwt?

aantron:
[...]

PR #484, opened 2017-10-14
attempt to move lwt.log into separate lwt_log

hannesm:
this is a first attempt to move lwt.log into its own package, so for a 4.0 release it can really be split.  please let me know what you think.

### DIFF
--- a/lwt_log.opam
+++ b/lwt_log.opam
@@ -1,0 +1,24 @@
+opam-version: "1.2"
+version: "dev"
+maintainer: [
+  "Anton Bachin <antonbachin@yahoo.com>"
+]
+authors: [
+  "Gabriel Radanne"
+]
+homepage: "https://github.com/ocsigen/lwt"
+doc: "https://ocsigen.org/lwt/api/Lwt_log"
+dev-repo: "https://github.com/ocsigen/lwt.git"
+bug-reports: "https://github.com/ocsigen/lwt/issues"
+license: "LGPL"
+
+depends: [
+  "jbuilder" {build}
+  # This will be constrained with an upper bound once the log is fully factored
+  # out of package lwt.
+  "lwt"
+]
+
+build: [
+  ["jbuilder" "build" "-p" name "-j" jobs]
+]
--- a/lwt_log.opam
+++ b/lwt_log.opam
@@ -4,7 +4,8 @@ maintainer: [
   "Anton Bachin <antonbachin@yahoo.com>"
 ]
 authors: [
-  "Gabriel Radanne"
+  "Shawn Wagner"
+  "Jérémie Dimino"
 ]
 homepage: "https://github.com/ocsigen/lwt"
 doc: "https://ocsigen.org/lwt/api/Lwt_log"
@@ -14,8 +15,8 @@ license: "LGPL"
 
 depends: [
   "jbuilder" {build}
-  # This will be constrained with an upper bound once the log is fully factored
-  # out of package lwt.
+  # This will be constrained with an upper bound once the log library is fully
+  # factored out of package lwt.
   "lwt"
 ]
 
--- a/META.lwt
+++ b/META.lwt
@@ -16,7 +16,7 @@ plugin(native) = "lwt.cmxs"
 package "log" (
   #directory = "log"
   version = "dev"
-  description = "Logger for Lwt"
+  description = "Logger for Lwt (deprecated; use lwt_log)"
   requires = "bytes lwt result"
   archive(byte) = "lwt_log.cma"
   archive(native) = "lwt_log.cmxa"
--- a/META.lwt_log
+++ b/META.lwt_log
@@ -3,7 +3,7 @@
 # lwt.log.
 
 version = "dev"
-description = "Logger for lwt"
+description = "Logger for lwt (deprecated; use logs library)"
 requires = "lwt_log.unix"
 
 package "core" (
--- a/META.lwt_log
+++ b/META.lwt_log
@@ -4,4 +4,16 @@
 
 version = "dev"
 description = "Logger for lwt"
-requires = "lwt.log"
+requires = "lwt_log.unix"
+
+package "core" (
+  version = "dev"
+  description = "Logger for Lwt (pure-OCaml core)"
+  requires = "lwt.log"
+)
+
+package "unix" (
+  version = "dev"
+  description = "Logger for Lwt (Unix logging destinations)"
+  requires = "lwt.log lwt.unix"
+)
--- a/META.lwt_log
+++ b/META.lwt_log
@@ -1,0 +1,7 @@
+# The Lwt log will, in the future, be moved out to this package lwt_log. For
+# now, this package is a compatibility wrapper for the log's current location in
+# lwt.log.
+
+version = "dev"
+description = "Logger for lwt"
+requires = "lwt.log"
--- a/META.lwt_log
+++ b/META.lwt_log
@@ -4,16 +4,10 @@
 
 version = "dev"
 description = "Logger for lwt (deprecated; use logs library)"
-requires = "lwt_log.unix"
+requires = "lwt.log lwt.unix"
 
 package "core" (
   version = "dev"
   description = "Logger for Lwt (pure-OCaml core)"
   requires = "lwt.log"
-)
-
-package "unix" (
-  version = "dev"
-  description = "Logger for Lwt (Unix logging destinations)"
-  requires = "lwt.log lwt.unix"
 )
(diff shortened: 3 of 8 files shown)


hannesm:
I'm not sure in how many packages lwt should be split -- the log sublibrary is used by both `preemptive` and `unix` -- thus only once these pieces are moved into separate packages, there is an actual effect.  before that it is dangerous (if split away it introduces circular dependencies).  OTOH `unix` and `preemptive` can be rewritten using the `Logs` library (or to not care about initialising loggers, but use Logs, and rely on applicaions to setup reporters as appropriate), but this will likely break the API.  Any opinions?  I always failed to use Lwt_log, and am pretty happy with the `Logs` library myself.

aantron:
There seem to be two kinds of interactions between `Lwt_log` and the rest of `lwt.unix`/`lwt.preemptive`:

1. `lwt.unix` uses `Lwt_log` for actual logging. I think all of these should be removed. Lwt shouldn't do any logging.
2. `Lwt_daemon` actually has a non-trivial interaction in the API. I don't know immediately what to do about this. I guess it would prevent factoring out `lwt.log`.

For (1), I see only two trivial logging expressions (in `lwt_throttle.ml` and `lwt_timeout.ml`). AFAICT, `lwt.preemptive` doesn't actually ever log anything, even though the ability to set a logging function is provided in its API, and it references `Lwt_log` in its default logging function (which, again, AFAICT is never called).

aantron:
In summary, I'd be happy to get the process that this PR would start, actually started, assuming we can figure out what to do about `Lwt_daemon` :) Otherwise, I think we will have to delay it for now, by default.

hcarty:
`Lwt_daemon` may be worth spinning off into its own package as well. It works for basic uses but could use some extra attention from what I recall.

aantron:
Yeah, factoring out `Lwt_daemon` might be the only way to shed `Lwt_log`.

I really don't like the idea of creating new packages just to deprecate them, or factoring out a package with one module that has one function, but I can't think of anything better.. and it will do the trick.

aantron:
Is there maybe some naming convention or `opam` field we could use or create, for marking such packages? Like `lwt_camlp4_deprecated`, or `deprecated: true`?

aantron:
What's slowing me down right now is that I think `Lwt_daemon` is really helpful for Lwt. Like, it is not obvious how to write an Lwt-friendly `daemonize` from scratch. So, we will need to address this either with excellent docs, or with a new `daemonize` (in `Lwt_unix`?) that doesn't touch `Lwt_log`. If we go the latter route, maybe we should include `Lwt_daemon` in the `lwt_log` package, even though it will seem kind of weird.

aantron:
I looked through opam packages depending on Lwt, and found:

- `arakoon`, `charrua-*`, `message-switch` use `Lwt_daemon`, but in a way that does not interact with `Lwt_log` at all.
- `usbmux`, `xentropyd` depend on the `?syslog` parameter of `Lwt_daemon.daemonize`.
- `yurt` depends on the ability to redirect to loggers.

So cc @toolslive, @domsj, @mseri, @gaborigloi, @fxfactorial, @djs55, @zshipko for any opinions on gently moving [`Lwt_daemon`](https://ocsigen.org/lwt/dev/api/Lwt_daemon) to a new package `lwt_log` (or `lwt_log_unix`, see below).

A large number of packages depend on `Lwt_log` itself, but I'm pretty confident we should factor that out. It's going to be a bit of a pain, as there are a non-Unix and a Unix layers, so perhaps we will need two packages. We should consider that now, so we can create the right package names. @Drup, any opinions on this based on Ocsigen experience with `Lwt_log` and Unix/"pure" packages?

The plan would be:

- Create a "proxy" `lwt_log` package with `Lwt_log`, `Lwt_log_core`, `Lwt_daemon` so users can update their deps starting with Lwt 3.2.0 (this PR).
- Remove all dependencies of Lwt on `Lwt_log` besides `Lwt_daemon`.
- Make `lwt_log` a "real" package in 4.0.0, and move the above modules to it. For users that update their deps, the transition should be smooth. We will notify during release of 3.2.0 the maintainers of all opam packages that depend on any of the code to be moved.
- On demand, if someone wants to have a `daemonize` function that does not come from `lwt_log`, we will add `Lwt_unix.daemonize` that is not aware of loggers.

There is one more complication, which is that the PPX and the Camlp4 syntax have some support for `Lwt_log`. However, I think it is indirect. A user of those features can depend on the new `lwt_log`/`lwt_log_unix` to make sure the generated code compiles and links.

lwt_log.opam
```diff
@@ -0,0 +1,24 @@
+opam-version: "1.2"
+version: "dev"
+maintainer: [
+  "Anton Bachin <antonbachin@yahoo.com>"
+]
+authors: [
+  "Gabriel Radanne"
```
Drup: Uh, no. :D

Drup:
What's the purpose of spliting lwt_log out ? I doesn't really bring extra dependencies or anything, and it's not like it's difficult to build.

aantron:
@Drup, to deprecate it and encourage use of `logs` or other libraries. See https://github.com/ocsigen/lwt/issues/468#issuecomment-336575454 and further. Also to make the mapping between opam and Findlib packages more direct.

aantron:
The "broader" picture is I'd be happy to see work on `Lwt_log`, but I don't think there is enough labor available for working on Lwt in general, and there are much higher priority Lwt issues and projects besides `Lwt_log`. As a result, a higher quality can be more easily achieved in libraries outside Lwt. So I prefer to focus on the core of what Lwt does (promises and I/O for now), and shed anything else if there are good alternatives.

toolslive:
@aantron [Alba](https://github.com/openvstorage/alba) is a dependency too. Anyway, the above plan looks fine, and we'll adapt when we upgrade our dependency chain.

rgrinberg:
Another point to move daemonization to its own package is that it really shouldn't be encouraged either. It doesn't really serve a purpose when you have a modern init system http://blog.benoitblanchon.fr/linux-daemon-considered-harmful/

aantron:
It seems to also be deprecated on macOS, but I'm not sure about "obscure" systems.

avsm:
> The "broader" picture is I'd be happy to see work on Lwt_log, but I don't think there is enough labor available for working on Lwt in general, and there are much higher priority Lwt issues and projects besides Lwt_log. As a result, a higher quality can be more easily achieved in libraries outside Lwt. So I prefer to focus on the core of what Lwt does (promises and I/O for now), and shed anything else if there are good alternatives.

This sounds like an excellent approach to me. When `Lwt_logs` was first created, there were very few alternatives, but that is no longer the case.  There are also quite a few libraries available for OS-specific launching now (e.g. ocaml-launchd for osx) as an alternative to daemonising, as @rgrinberg notes.

mseri:
Your plan makes perfect sense and I welcome the split and the cleanup.

We use the daemon and the logs in few places but somewhat inconsistenly. We have one more package that is using `Lwt_daemonize` and `Lwt_log`: https://github.com/xapi-project/wsproxy/blob/master/cli/wsproxy.ml#L116 but I don't think that would be massively affected either

mseri:
@avsm @rgrinberg What libraries would you recommend as daemonising alternatives on linux? We are using `ocaml-systemd` in some places, but we couldn't use it everywhere due to undebuggable segfaults in some cases

rgrinberg:
You don't need any libraries, or daemonization for that matter. Just write
a systemd service file that uses your binary. This process isn't in any way
unique to ocaml so there's not much to even document for us.

aantron:
Ok, I rebased this PR (there was a merge conflict). I added more commits, such that:

- @Drup is no longer listed as the author of `Lwt_log`, per his objection.
- There are two proxy packages created, instead of one. They are `lwt_log.core` for modules `Lwt_log_core`, `Lwt_log_rules` (the pure-OCaml code), and `lwt_log.unix` for modules `Lwt_log`, `Lwt_daemon` (the Unix stuff). This reflects a pre-existing split in the log library in Lwt.
- `lwt_log` is an alias for `lwt_log.unix`. This could be surprising, because `lwt.log` corresponds to `lwt_log.core`, and users that wanted what will be `lwt_log.unix` currently have to depend on `lwt.log` *and* `lwt.unix` (extra credit if you can actually follow that sentence and distinguish underscores from periods very easily :p). However, I believe the *current* setup is the more surprising one, I would expect depending on `lwt.log` to give me module `Lwt_log`, but it doesn't.
- I added a usage test to make sure things remain working in the future.

If there are no objections, I will squash-merge this (so @hannesm will be the author, I will be the committer, and there will be a link from the commit message to this PR for the details).

That will be the first step of factoring out `lwt_log`, and will be released in 3.2.0.

In a previous commit, 4a7cf8c, now in `master`, I removed all other dependencies on `Lwt_log` in `lwt.unix`, besides the one in `Lwt_daemon`.

hannesm:
@aantron thanks for taking care of that!  highly appreciated!

Issue #485, opened 2017-10-20
"Overhaul" Lwt_pool


aantron:
While keeping the API the same, we should clear up the implementation, and write somewhat better API docs for the interface. Follow-on to #483. cc @hcarty

bobbypriambodo:
Would you think that this is a proper time to get feedback on Lwt_pool's semantics? I'm thinking of digging into implementations in other libs/languages, but perhaps we also can ask at discuss.ocaml.org (or at least direct them to this issue) for people who use Lwt_pool if they have concerns on the underlying implementation (or missing features). Identify if the current API and behavior makes sense at all, and we can overhaul all the way.

I can't help but feel that the current API of Lwt_pool is kinda strange, but that might just be me. In any case, most of usages of Lwt_pool that I found on GitHub don't specify the optional arguments, and I might have missed it but I found none that uses `?check`.

But yeah, tweaking the API or behavior might then introduce major semver update, and I'm not sure if that's acceptable.

aantron:
Getting feedback on Lwt_pool semantics is indeed good.

Regarding the optional arguments: we probably shouldn't remove them anytime soon anyway. However, since they are optional, we can heavily de-emphasize them, essentially deprecate them, add new optional arguments, or add a new pool creation API that doesn't have them. So there will be a little bit of a legacy annoyance for us to implement for these arguments, but it shouldn't be too bad.

bobbypriambodo:
Point taken @aantron, thanks. If no one minds I would love to give this a go and see what we can do with it.

vouillon:
The type of the `check` argument should probably have been `'a -> bool Lwt.t`.

The idea was that if we get an exception while using an element of the pool, the element might not be in a stable state, so we should have the opportunity to put it back in a stable state (for instance, if its a database connection, we may want to abort any transaction that might be in progress) or to dispose it, rather than adding it back to the pool right away.

bobbypriambodo:
> The idea was that if we get an exception while using an element of the pool, the element might not be in a stable state

Yes, if that clarifies it. We can use that kind of info in the docs I believe.

bobbypriambodo:
Okay, so here are my thoughts so far. This is mostly going to be my notes from reading stuff and seeing other's implementations.

A basic, typical resource pool has two public API: `acquire` and `release`. Those names should readily imply what they do, so no confusion there. These two functions are the low-level building blocks of a pool.

[...]

PR #490, opened 2017-10-27
Lwt_pool: add a function returning the length of the wait queue

vouillon:
This provides a way to known whether the pool is contended.

### DIFF
--- a/src/core/lwt_pool.ml
+++ b/src/core/lwt_pool.ml
@@ -182,3 +182,5 @@ let clear p =
   old_cleared := true;
   p.cleared := ref false;
   Lwt_list.iter_s (dispose p) elements
+
+let wait_queue_length p = Lwt_sequence.length p.waiters
--- a/src/core/lwt_pool.mli
+++ b/src/core/lwt_pool.mli
@@ -77,3 +77,7 @@ val clear : 'a t -> unit Lwt.t
       pool element.
 
       Disposals are performed sequentially in an undefined order. *)
+
+val wait_queue_length : _ t -> int
+  (** [wait_queue_length p] returns the number of threads currently
+      waiting for an element of the pool [p] to become available. *)


hcarty:
Seems like a reasonable addition.

bobbypriambodo:
Hi @vouillon! Thanks for taking the time to create this PR. May I know what your use case is for this function?

Although I agree with @hcarty that the change seems reasonable, we are actually planning to do a revamp on Lwt_pool (#485), mostly to clarify the semantics and docs. I was thinking this could be a good chance to get another insight on the direction users (and owners!) want Lwt_pool to go 🙂

hcarty:
I lean toward merging this as-is. It can be included in the Grand Lwt_pool Overhaul of 2017. We can go through any bikeshedding over naming and so on as part of that process.

bobbypriambodo:
Oh, sure, I'm not against the addition, the implementation itself is also straightforward. If you and @aantron are also okay with it then please do. Perhaps Jerome can join us later in the discussion on the issue if he has inputs.

vouillon:
We are using `Lwt_pool` to manage a pool of connections to a database for a Web server. If the Web server is overloaded, we don't want threads piling up waiting for a connection to the database, while the Web server slows down to a crawl. We should rather reject some HTTP requests. To implement this, we need a way to observe that the pool is contended. A function `wait_queue_length` is a minimal way to achieve this : one can call it every so often and start rejecting some HTTP requests when the queue becomes too long.
The function is O(n) in the length on the queue, but that should not be a problem: if the queue is long, we are in trouble already.
One could also add some hooks to gather more precise statistics or avoid polling, but that seems overkill.

aantron:
I vaguely remember some other pool library (not in OCaml) supporting a limit on the number of blocked requests for an element. Not sure how that would look in `Lwt_pool` – maybe `use` would have to return a promise for an element option. I guess that would be a breaking change, if done that way.

Besides being potentially incompatible with the current API, I'm not sure if this is otherwise in any way "better" or "worse" than what's being proposed in this PR. I just want to note it.

bobbypriambodo:
> @aantron maybe use would have to return a promise for an element option.

The way Haskell's [Data.Pool](https://hackage.haskell.org/package/resource-pool-0.2.3.2/docs/Data-Pool.html) does it is in addition to `withResource` (the analog to `use`) it also provides `tryWithResource`, where the former will block, but the latter will return immediately with a Maybe Pool. Some other libs also don't implement "waiting" and just throw exceptions on acquiring exhausted pool, others may specify how long they will wait before throwing.

vouillon:
Note that you want to reject HTTP requests as soon as possible, consuming as little of resources as possible. If you need 10 database queries to handle an HTTP request (10 call to `use`), and if each query has a small chance of failing, you have a much larger chance of failing to process a request, after spending a significant amount of resources processing it. It is much better if you reject a fraction of the requests right away and fully process all the other requests.
So, I don't think that fixing a hard limit on the number of blocked requests for an element is the right approach, at least in our case.

bobbypriambodo:
Seeing as there is no apparent opposition, the use case seems valid, and the implementation is straightforward, I'm going to go ahead and merge this so that it can be included in the overhaul plan.

bobbypriambodo:
Thanks @vouillon!

PR #501, opened 2017-11-16
Clarifies Lwt_pool impl function names and comments

bobbypriambodo:
This is a part of Lwt_pool overhaul (#485). Several things this PR is aimed to address:

1. Replace _threads_ with _promise resolvers_ (or just _resolvers_). @aantron need you to clarify this usage as I don't think replacing it with only _promise_ work in this context.
2. Further clarifies implementation comments about `validate` and `check`.
3. Rename the misleading `check_elt` function into `validate_and_return`.
4. Rename `checked_release` function into `check_and_release`.

Number 3 and 4 were meant to give an association with the respective `validate` and `check` functions used in each. I realize that using "and" in function names is bad design, but that might show how the implementation can be improved 🙂 there's probably a better way to actually write the whole thing in a cleaner way, but I'm thinking of getting this off first.

Suggestions welcome!

### DIFF
--- a/src/core/lwt_pool.ml
+++ b/src/core/lwt_pool.ml
@@ -20,7 +20,7 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  *)
 
-(* [Lwt_sequnece] is deprecated to prevent users from using it, but it is used
+(* [Lwt_sequence] is deprecated to prevent users from using it, but it is used
    internally by Lwt. *)
 [@@@ocaml.warning "-3"]
 module Lwt_sequence = Lwt_sequence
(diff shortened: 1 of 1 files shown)


src/core/lwt_pool.ml
```diff
@@ -20,26 +20,21 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
  *)
 
-(* [Lwt_sequnece] is deprecated to prevent users from using it, but it is used
+(* [Lwt_sequence] is deprecated to prevent users from using it, but it is used
```
aantron: Argh, thanks for catching this. I copy/pasted this error into a whole bunch of files.

aantron:
It looks good. Replacing *threads* with *resolvers* when the objects are `Lwt.u`s is indeed correct :) I don't object to `and` in function names, if it's the clearest known descriptive name :) I think we can merge this when all the builds are green. Thanks!

bobbypriambodo:
Great! This PR is a bit gentle, and I think we could refactor and improve things in many places and be more "daring" on next iterations. Really glad there are tests to make sure I don't break things (I think I can attribute this to @jsthomas? Thank you 🙂).

aantron:
Many thanks!

hcarty:
Looks great!

Issue #502, opened 2017-11-26
Memory leak in Lwt_unix.getnameinfo


adrien-n:
Hi,

The code below leaks memory. Using `Unix.getnameinfo` instead of `Lwt_unix.getnameinfo` avoids the leak. This reproducer defines two functions: `leak` and `no_leak`; the first one uses the `getnameinfo` from `Lwt_unix` while the second one uses the one from `Unix`. 

The reproducer calls `Gc.compact`, prints memory stats, then calls `getnameinfo` on `"1.2.3.4"` a thousand times, calls `Gc.compact` again, print memory stats again, and repeats this whole procedure 10 times. Memory usage ends up over 100MB while the code that uses `Unix` only uses a few MBs (3 or so iirc: it's low enough that I had trouble spotting it in `top`).

This happens on lwt 3.1.0, installed through opam on 4.04.2.

    let sockaddr =
      let ip = "1.2.3.4" in
      let inet_addr = Unix.inet_addr_of_string ip in
      Unix.ADDR_INET (inet_addr, 0)

    let leak () =
      match%lwt Lwt_unix.getnameinfo sockaddr [] with
      | _ -> Lwt.return ()
      | exception _ -> Lwt.return ()

    let no_leak () =
      match Unix.getnameinfo sockaddr [] with
      | _ -> Lwt.return ()
      | exception _ -> Lwt.return ()

    let gc () =
      print_endline "=> Compacting";
      Gc.compact ();
      print_endline "=> Memory stats";
      Gc.print_stat stdout;
      print_newline ();
      flush stdout

    let loop_iteration l () =
      gc ();
      let%lwt x = Lwt_list.iter_p leak l in
      gc ();
      Lwt.return x

    let () =
      let l1 = Array.to_list (Array.make 1000 ()) in
      let l2 = Array.to_list (Array.make 10 ()) in
      Lwt_main.run (Lwt_list.iter_s (loop_iteration l1) l2)

aantron:
Thanks for the report. I think we should look into this soon, before releasing 3.2.0 (in the coming days).

hannesm:
please see whether #503 fixes the issue for you

aantron:
Apparently fixed by #503. Thanks @adrien-n for reporting and @hannesm for fixing!

PR #503, opened 2017-11-26
lwt_unix_free_job if getnameinfo was successfull

hannesm:
tentative fix for #502 
I checked all the `result_*` in `lwt_unix_unix.h`, and this was the only one I could spot which missed the `lwt_unix_free_job` in one branch.


aantron:
Thanks for checking that @hannesm, I'll merge once the AppVeyor build passes.

adrien-n:
Thanks for the fix. I can confirm it fixes my reproducer. :)

Issue #463, opened 2017-08-04
[easy] Test order in which cancelation and resolution happen in pick, npick


aantron:
When Lwt resolves several promises "simultaneously," it actually resolves them in sequence, for example by iterating over them, updating their state ([example][iter]). Callbacks are called *during* this iteration, i.e. if iterating over `[promise1; promise2]`, first all of `promise1`'s callbacks are executed completely, then `promise2`'s. This means that if `promise1`'s callbacks examine the state of `promise2`, they will see that it is still pending (`Lwt.Sleep`). Meanwhile, the other way, if `promise2`'s callbacks examine the state of `promise1`, they will see it already resolved, in this case rejected (`Lwt.Fail Lwt.Canceled`):

```ocaml
let () =
  let p1, _ = Lwt.task () in
  let p2, _ = Lwt.task () in

  (* This is just to assert that the callbacks actually ran. *)
  let callback_calls = ref 0 in

  Lwt.on_termination p1 (fun () ->
    Pervasives.incr callback_calls;
    assert (Lwt.state p2 = Lwt.Sleep));

  Lwt.on_termination p2 (fun () ->
    Pervasives.incr callback_calls;
    assert (Lwt.state p1 = Lwt.Fail Lwt.Canceled));

  List.iter Lwt.cancel [p1; p2];

  assert (!callback_calls = 2)

(* ocamlfind opt -linkpkg -package lwt order.ml && ./a.out *)
```

This is probably fine (but maybe not?), because the order of promises in the list is arbitrary. A more interesting case is when [`Lwt.pick`][pick] and [`Lwt.npick`][npick] both (1) do the above iteration, canceling their argument promises, and also (2) resolve their final promise (the one that they return). Should the argument promises be canceled first, or should the final promise be resolved first?

[`Lwt.pick`][pick] cancels the argument promises first. There might be some kind of wisdom implicitly encoded in this choice – I am not sure. However, to ease refactoring for users, the order should be the same in `pick` and `npick`.

When [fixing `Lwt.npick`][npick-pr], this came up, because the existing code in `npick` encouraged the opposite choice, resolving the final promise first (see [discussion](https://github.com/ocsigen/lwt/pull/447#discussion_r128904573) and [exercising code with explanation](https://gist.github.com/aantron/4db12b8e5ca92f57f7a223e594c833cc)). It had to be refactored a bit to get the same order as `pick`.

[...]

PR #507, opened 2017-12-17
Issue 463: Test the order in which cancellation/resolution occur in pick, npick

jsthomas:
This change addresses #463 by introducing tests verifying `pick` and `npick` have consistent behavior, specifically that they cancel argument promises before the final promise resolves. The tests are a modification of [this example](https://gist.github.com/aantron/4db12b8e5ca92f57f7a223e594c833cc), by @aantron.


aantron:
Thanks for that!

Issue #407, opened 2017-06-14
[easy-ish] Reorganize the C files


aantron:
Lwt's C files are currently quite messy. For example, [`lwt_unix_unix.h`](https://github.com/ocsigen/lwt/blob/6649ef6f74fdff417e705e131e5ec31487ea9032/src/unix/lwt_unix_unix.h) is currently about 3300 lines of stress-inducing C code, which is enough to give anyone a headache. Also, it shouldn't be a header file – everything in it is an implementation. However, it is [currently included from `lwt_unix_stubs.c` as a janky form of conditional compilation](https://github.com/ocsigen/lwt/blob/master/src/unix/lwt_unix_stubs.c#L90-L94).

Fixing that conditional compilation scheme might be somewhat involved, but the first thing we should do is split up `lwt_unix_unix.h` and `lwt_unix_windows.h` into lots of little files, close to one per C call or small C call category. For comparison, see the files of [`Unix` from the standard library](https://github.com/ocaml/ocaml/tree/50b2c095e6772e4a058434d328d59b7094a4dbff/otherlibs/unix), and the [Windows equivalent](https://github.com/ocaml/ocaml/tree/50b2c095e6772e4a058434d328d59b7094a4dbff/otherlibs/win32unix).

We can create all those little files in two new subdirectories of [`src/unix/`](https://github.com/ocsigen/lwt/tree/6649ef6f74fdff417e705e131e5ec31487ea9032/src/unix) (one for actual Unix and one for Windows), and, as a first cut, `#include` each one of them into either `lwt_unix_unix.h` and `lwt_unix_windows.h`. This should be relatively simple, as C preprocessor inclusion is textual substitution. Header guards (`#pragma once` or `#ifndef`) are not necessary, because we know each of these files is only going to be included once, and also because the ultimate plan is to turn them into `.c` files eventually.

It would be great if we could figure out how to make git trace the history of each one of these files back to `lwt_unix_unix.h` or `lwt_unix_windows.h`, but I doubt git is actually able to do that.

**If you'd like to work on this, please leave a comment below. We want to make sure that nobody else is likely modifying these files at the same time, to avoid a potentially brutal merge conflict.**

andrewray:
I am sure this is possible, however, a word of caution -- I changed the lwt_XXX_stubs.c file to a .h file to ensure jbuilder copied it when building (in fact it actually installs them as well).

I believe the more flexible way to deal with C code is to define it in a separate library.

[...]

PR #508, opened 2017-12-17
Start split lwt unix C bindings to multiple C files

cedlemo:
#407


aantron:
The PR overall looks good. I can't wait until the entire C bindings looks like this. It will be much more pleasant to work with.

The AppVeyor failures look to be likely due to some missing `caml/*.h` includes in various files.

Compiling on macOS, the Windows C files are empty, which results in a bunch of warnings like this:

```
src/unix/liblwt_unix_stubs.a(windows_get_page_size.o) has no symbols
src/unix/liblwt_unix_stubs.a(windows_read.o) has no symbols
# etc
```

This is not a big deal. I guess the "right" way to solve it is to not compile the files when not on Win32 in the first place, but I understood that it's not so easy to arrange that, so we can deal with it later/never.

For comparison, this is how I solved a no symbols warning in the past: c1021129ad433ff0d2ffb01f56c6c760ea26deb2. However, I don't think that's practical or wise to duplicate in every Windows `.c` file.

aantron:
Thanks! Had to restart like four stuck builds in AppVeyor, but it's good now :)

Issue #452, opened 2017-07-18
[easy] Test interaction of Lwt.finalize and Lwt.cancel


aantron:
See https://discuss.ocaml.org/t/lwt-finalize-and-failure-in-the-cleanup-code/525/4.

Basically, when a promise is canceled, after Lwt [figures out which initial promise](https://github.com/ocsigen/lwt/blob/fa9eb6781a813122b94b56b31361c29492782e51/src/core/lwt.ml#L234-L271) to actually cancel, cancelation becomes just an ordinary exception (`Lwt.Canceled`) propagating normally forwards through the promise chains.

When this exception is handled by [`Lwt.finalize`](https://ocsigen.org/lwt/dev/api/Lwt#VALfinalize), if the finalizer (`g`) *also* raises an exception, that exception "replaces" the `Lwt.Canceled`. From the docs (clarified recently by @Richard-Degenne in #444):

> `finalize f g` runs `f ()`, and then `g ()`, whether `f ()` fails or completes normally. `Lwt.finalize` returns the result of `f ()`, unless `g ()` fails, in which case it returns the exception raised by `g ()`.

We already [have tests](https://github.com/ocsigen/lwt/blob/fa9eb6781a813122b94b56b31361c29492782e51/test/core/test_lwt.ml#L933-L950) for "replacement" of an ordinary exception by another ordinary exception by `g` in `finalize`, but [not a test](https://github.com/ocsigen/lwt/blob/fa9eb6781a813122b94b56b31361c29492782e51/test/core/test_lwt.ml#L2711) when the first exception was triggered by `Lwt.cancel`. This issue is about adding the latter kind of test.

<br/>

#### Some more discussion

[...]

PR #509, opened 2017-12-18
Issue 452: Add a test for the interaction of Lwt.finalize and Lwt.cancel.

jsthomas:
This change addresses #452 by introducing a test verifying the following property:

When running `Lwt.finalize f g`, if `f` is terminated by `Lwt.cancel` (a particular type of exception), and running `g` produces a second exception `e`, then the return value of `Lwt.finalize f g ` is `e`.


aantron:
Thanks again!

Issue #395, opened 2017-06-11
[easy] Involved tests for Lwt_list


aantron:
For each (pair of) function(s) in [`Lwt_list`](https://ocsigen.org/lwt/dev/api/Lwt_list), (say `Lwt_list.map_s` and `Lwt_list.map_p`) we should probably test:

- [x] ~~How they respond to exceptions raised by `f`, i.e. `Lwt_list.map_s (fun _ -> raise Exit) ...`~~ (@jsthomas in #512). The `_p` functions might be particularly interesting here. We should find out what they actually do, and document it (or document `Lwt_list` at all, for that matter).
- [x] ~~That the `_s` functions really run `f` in series, and the `_p` functions really run it in parallel. The parallel tests should be easier here, as one could `Lwt.wakeup_later` a promise that another invocation of `f` is waiting on, and verify that the other invocation does some work first, i.e. two or more invocations "talk" to each other. Not immediately sure what to do for `_s`.~~ (@jsthomas in #535).
- [x] After #347 is addressed, we should test that all functions are stack-safe, by running them on big lists.

See #385 for the testing workflow.

jsthomas:
I'm interested in working on this issue.

aantron:
For the `_p` functions, it looks like the ones that depend on `Lwt.join` (`iter_p` and `iteri_p`), if multiple calls to `f` fail, the failure of the first one to fail in time will be reported. For all other `_p` functions, it looks like the left-most failure will be reported, regardless of whether that was the first failure in time.

That may be worth fixing.

It's probably best to wait until we are testing for tail-recursive implementations, as we may find some stack bugs that we have to fix, and that will affect how we may want to fix exception reporting.

aantron:
I think there is also a difference in *when* the promises returned by the `_p` functions are resolved. For the ones that internally use `join`, the final promises are resolved only when all the individual promises created for the whole list resolve. For the rest, if one of the individual promises is rejected, the final promise is rejected right away, without waiting for any promises further down the list to be resolved.

We should probably choose what's right, and test for these behaviors. I suspect that `join`-like behavior is the most desirable. It most closely mimics the behavior of the `_p` functions in the case where none of the promises are rejected.

aantron:
[...]

PR #512, opened 2017-12-22
Issue 395: Add tests for exception handling behavior of Lwt_list functions.

jsthomas:
This change addresses the first of three parts of #395, testing how functions in `Lwt_list` respond to exceptions. 

The functions `map_p`, `map_s`, `iter_p`, and `iter_i` were already tested by a base testing function `test_exception`. The tests introduced here make use of that same function, which verifies that if an exception `E` occurs when running something like `Lwt_list.map_p f l`, then the promise evaluates to `Lwt.Fail E` (and the exception doesn't propagate any farther).

In the course of writing my tests, I found that this exception handling behavior is **not** consistent across all `Lwt_list` functions in their current implementation. For example, an exception arising in the course of running `Lwt_list.mapi_p f l` gets propagated up to the caller; you would need to surround it with a `try ... catch` to handle it. In my (inexpert) opinion, the behavior tested by `test_exception` seems like the right design.

The first commit in this PR sets up the tests to verify every function in `Lwt_list` follows the convention defined in `test_exception` and the second commit modifies `Lwt_list` to make the tests pass.


aantron:
Thanks! This fixes a ton of bugs in `Lwt_list`. I'm squash-merging the commits, to avoid adding a point in history where the build would fail (this harms bisecting).

PR #515, opened 2017-12-23
[4.0.0] Complete factoring out the PPX

### LINKED ISSUE #338: Move the PPX to a separate package (?)

aantron:
The PPX was already fully packaged as `lwt_ppx`. This commit deletes package `lwt.ppx`.

Resolves #338.

I'll fix the CI builds later, as the CI needs a general cleanup anyway.

### DIFF
--- a/doc/CONTRIBUTING.md
+++ b/doc/CONTRIBUTING.md
@@ -72,15 +72,13 @@ git clone https://github.com/your-user-name/lwt.git
 cd lwt/
 ```
 
-We'll use Lwt's own [`opam`][opam-depends] file to install Lwt's dependencies
-automatically. Before doing that, you may want to switch to a special OPAM
-switch just for Lwt:
+Now, we need to install Lwt's development dependencies. Before doing that, you
+may want to switch to a special OPAM switch for working on Lwt:
 
 ```
-opam switch 4.04.02-lwt --alias-of 4.04.2   # optional
-eval `opam config env`                      # optional
-opam pin add --no-action lwt .
-opam install --deps-only lwt
+opam switch 4.06.1-lwt --alias-of 4.06.1   # optional
+eval `opam config env`                     # optional
+make dev-deps
 ```
 
 [opam-depends]: https://github.com/ocsigen/lwt/blob/8bff603ae6d976e69698fa08e8ce08fe9615489d/opam/opam#L35-L44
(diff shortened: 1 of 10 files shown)


doc/CONTRIBUTING.md
```diff
@@ -129,14 +128,7 @@ modified code when you run these commands.
...
 
-Coverage analysis is only enabled for `src/core/*` for now. To generate coverage
-reports, first pin Bisect_ppx to its development version:
-
-```
-opam pin add --dev-repo bisect_ppx
-```
-
-Then run
+To generate coverage reports, run
```
avsm: minor nit, missing `:` at end of sentence

aantron: Ultra grammar mode! :p

Modulo mistakes made here and there, I put the colon (sometimes preceded by "with" or some other preposition) when the sentence introducing the quote/list/other block can stand on its own, and the quote, etc., can be viewed as a separate illustration. It should be possible to replace the colon with a period, maybe deleting a preposition or other filler word.

I tend to omit the colon when the introduction sentence doesn't make sense on its own. In this case, it has a *different* sense:

> "To generate coverage reports, run."
>
> whenever-i-go-running-running-meme 1

The command, in this case, cannot be taken out of that sentence (or phrase), as it's the direct object. It should actually be set inline, but is a block for easier copy/pasting when someone just wants to quickly find all the commands in sequence, and avoid reading text.

I believe this is the recommended usage, at least in the U.S.

PR #519, opened 2017-12-23
[4.0.0] Switch to new callback deferral semantics

### LINKED ISSUE #329: Bind quirk

aantron:
This is the last bit of #500, and is discussed in the most detail in that PR.

Resolves #329.

### DIFF
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -1661,8 +1661,7 @@ struct
              [Proxy _] links, and it's not necessary to worry about whether the
              inductive reasoning at [may_now_be_proxy] applies. *)
 
-        let State_may_have_changed p' =
-          resolve ~allow_deferring:false p' p_result in
+        let State_may_have_changed p' = resolve p' p_result in
         ignore p'
       in
 
@@ -1693,8 +1692,7 @@ struct
            this is as in [protected] and [may_now_be_proxy], but even simpler,
            because [p'] is not cancelable. *)
 
-        let State_may_have_changed p' =
-          resolve ~allow_deferring:false p' p_result in
+        let State_may_have_changed p' = resolve p' p_result in
         ignore p'
       in
       add_implicitly_removed_callback p_callbacks callback;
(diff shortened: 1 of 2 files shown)


aantron:
Per https://discuss.ocaml.org/t/1337/13, Lwt is abandoning or deferring this change, so I am closing the PR.

Issue #511, opened 2017-12-18
Retiring lwt_camlp4 and lwt_log from the repo


aantron:
These are deprecated, in favor of the PPX and [`Logs_lwt`](http://erratique.ch/software/logs/doc/Logs_lwt.html), respectively. Once these are factored out into their own opam packages (#370, #484) in 4.0.0, we don't intend to ever maintain them again. At that point, the code will be only an eyesore in the Lwt repo.

So, I want to delete it from the repo. I propose to git-subtree each of these modules out into their own repos, say `aantron/lwt_camlp4` and `aantron/lwt_log`, where they can remain archived forever. If somebody wants to take over them and do some more work, we can then easily hand them over, with a nice and complete history already there in each repo.

aantron:
Actually, I'd eventually like to also move at least `lwt_ssl` and `lwt_glib` from the main repo, without deprecating them. The same goes for `lwt_react`, but the desire there is a bit weaker. @rgrinberg, I remember you had an objection to this because there would be lots of separate issue trackers. However, these packages really have independent life cycles at this point, I think, and they see practically no issue traffic.

Other than the separate issue trackers and code living in separate repos, there shouldn't be any effect on users. We would change the `dev-repo` fields in `opam` for the latest releases. I don't think we would have to even re-release the packages.

Any thoughts?

rgrinberg:
For repos that you plan on deleting I think this is fine. But for the other repos my opinion is that while I see little risk in doing this, I also see absolutely no gain. All of these libraries still have you as the maintainer, so if there's more maintenance overhead with more repositories. You're only creating overhead for yourself by adding tasks like syncing breaking changes, managing constraints, more CI setups, etc. If you'd like to do this extra work, feel free to do it. Maybe I'm just not putting as much value on separate release cycles as you are.

Drup:
Yeah, I'm not convinced either. Also, forced synchronised releases is a problem when the big one at the source of the dependency chain moves much slower than the small ones. That's the other way around here. :)

aantron:
It actually seems like less of a headache to me, most especially for the libraries with extra system dependencies (`lwt_ssl`, `lwt_glib`). Splitting off their CI would simplify the CI of both `lwt` and those libraries. It is currently done conditionally, and it's annoying to deal with and understand. It grinds me down every time I maintain it.

[...]

PR #524, opened 2017-12-24
Move lwt_glib to its own repository

aantron:
It can now be found at [aantron/lwt_glib](https://github.com/aantron/lwt_glib), where it has its own build system and CI setup. I've `filter-branch`ed the entire history of `lwt_glib` out into that repo.

This breaks the "unified" docs build of Lwt, but that has been broken since mid-2017 at this point anyway. The Ocsigen infrastructure seems to be down.

I don't know if we need a new opam release of `lwt_glib`. The only significant change is the `dev-repo` field should now point to the other repo.

I built `lwt_glib` from that repo, and built `0install`, its only opam dep, against it. It seems to work fine.

After this point, we can treat `lwt_glib` as another ordinary user of Lwt. This simplifies CI and avoids wasting the attention of contributors.

Part of #511.

cc @talex5


hannesm:
@aantron you mentioned that the only change for a release would be a `dev-repo` update -- you can do this directly in the opam file in opam-repository.  though the `url` will still point to the latest `lwt` tarball, but i guess that's fine..

aantron:
Thanks @hannesm. That shouldn't cause people to re-download `lwt_glib` and trigger re-building of deps, right? I'm assuming that would only be triggered by changing the `url` file, specifically the checksum (which indeed doesn't have to change).

Drup:
Do you want me to move the repository inside ocsigen's organization ?

aantron:
> Do you want me to move the repository inside ocsigen's organization ?

I believe @Drup has continued this discussion here: https://github.com/ocsigen/lwt/issues/511#issuecomment-354983413

Issue #523, opened 2017-12-24
Remove the Unix stubs generator


aantron:
The [stubs generator](https://github.com/ocsigen/lwt/blob/caa642e8ecf043dae522bb7bedc90b1ca471c9e3/src/unix/stubs/gen_stubs.ml) (in [`src/unix/stubs`](https://github.com/ocsigen/lwt/tree/caa642e8ecf043dae522bb7bedc90b1ca471c9e3/src/unix/stubs)) generates [several of the C files](https://github.com/ocsigen/lwt/blob/caa642e8ecf043dae522bb7bedc90b1ca471c9e3/src/unix/jbuild#L51-L78) that make up the Unix binding, and one `.ml` file:

```
    lwt_unix_job_access.c
    lwt_unix_job_chdir.c
    lwt_unix_job_chmod.c
    lwt_unix_job_chown.c
    lwt_unix_job_chroot.c
    lwt_unix_job_close.c
    lwt_unix_job_fchmod.c
    lwt_unix_job_fchown.c
    lwt_unix_job_fdatasync.c
    lwt_unix_job_fsync.c
    lwt_unix_job_ftruncate.c
    lwt_unix_job_link.c
    lwt_unix_job_lseek.c
    lwt_unix_job_mkdir.c
    lwt_unix_job_mkfifo.c
    lwt_unix_job_rename.c
    lwt_unix_job_rmdir.c
    lwt_unix_job_symlink.c
    lwt_unix_job_tcdrain.c
    lwt_unix_job_tcflow.c
    lwt_unix_job_tcflush.c
    lwt_unix_job_tcsendbreak.c
    lwt_unix_job_truncate.c
    lwt_unix_job_unlink.c
    lwt_unix_jobs_generated.ml
```

It is meant to cut down on repetitive code. However, I personally have found it to be slightly opaque, and potentially doing more harm to legibility and maintainability than good.

This issue is about removing the stubs generator, and replacing it by ordinary code. Alternatively, if the stubs generator is really very useful, then this issue is about explaining why that is :)

Probably the easiest way to remove the stubs generator is to dump the code it generates, and commit it. It should be found in `_build/default/src/unix/jobs-unix`, and we want to move it from there to `src/unix/(unix|windows)_c`. After that, the `jobs-unix` directory can be deleted from the source tree.

[...]

PR #526, opened 2017-12-25
Remove stubs generator

cedlemo:
#523


aantron:
Thanks for that. The Unix binding is getting pretty clear, even if still quite large.

PR #531, opened 2017-12-31
Issue 395: Add tests that Lwt_list functions run in parallel/series

jsthomas:
This change proposes tests to address the second of the three parts of #395. These tests check that the `Lwt_list` functions ending in `_p` run in parallel and the `_s` tests run in series.

There are a couple of points where I think these tests are potentially dubious.
- As they're currently written, a test that fails will hang. I think this isn't ideal from a CI perspective (it would be preferable to immediately get a pass/fail).
- Based on comments on #395, it seems we'd prefer to have more self-contained tests rather than relying on many "helper functions". I've tried to avoid introducing too many helpers, but it's clear there's a lot of overlap/similarities between the new tests.

### DIFF
--- a/test/core/test_lwt_list.ml
+++ b/test/core/test_lwt_list.ml
@@ -132,19 +132,16 @@ let test_map_parallelism ?(rev=false) map =
   let expected = if rev then [2; 1] else [1; 2] in
   p >>= (fun l -> Lwt.return (l = expected))
 
-let test_map_serialism ?(rev=false) map =
-  let m = Lwt_mutex.create () in
-  let g x =
-    Lwt_mutex.unlock m;
-    Lwt.return (x + 1) in
+let test_map_serialism map =
+  let l = ref [] in
   let f x =
-    Lwt_mutex.lock m >>= fun _ ->
-    if x = 0 then g x
-    else Lwt.return (x + 1)
+    let p = Lwt.pause () >>= fun _ -> Lwt.return x in
+    assert (List.length !l = 0 || not @@ Lwt.is_sleeping @@ List.hd !l);
+    l := p :: !l;
+    p
   in
-  let p = map f [0; 1] in
-  let expected = if rev then [2; 1] else [1; 2] in
-  p >>= (fun u -> Lwt.return (u = expected))
+  let p' = map f [0; 1; 2] in
+  p' >>= (fun _ -> Lwt.return_true)
 
 let test_iter_parallelism iter =
   let t, w = Lwt.wait () in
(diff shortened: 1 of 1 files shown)


test/core/test_lwt_list.ml
```diff
@@ -119,6 +119,96 @@ let test_map f test_list =
...
+    Lwt_mutex.unlock m;
+    Lwt.return (x + 1) in
+  let f x =
+    Lwt_mutex.lock m >>= fun _ ->
+    if x = 0 then g x
+    else Lwt.return (x + 1)
+  in
+  let p = map f [0; 1] in
+  let expected = if rev then [2; 1] else [1; 2] in
+  p >>= (fun u -> Lwt.return (u = expected))
```
aantron: This test doesn't seem to be testing serialization of calls to `f` to me. `map_s` is supposed to guarantee that `f 1` isn't called until the promise returned by `f 0` is fulfilled.

For this test to check that, it would have to have some way of failing if `f 1` starts running before the promise `f 0` is fulfilled.

I'm actually not sure immediately how to do that. Perhaps you would have to mutate some state in a callback of `f 0`, and `f 1` would be fulfilled with that state. I don't think that's legitimate, though, because it would depend on Lwt's internal ordering of callback calls to ensure that the `f 0` callback that mutates the state runs before the call to `f 1` scheduled by `map_s`.

aantron: `Lwt.pause () >>= f` might help here. It will force everything that is ready to run immediately to run before `f`.

jsthomas: I've posted this new approach:
```
let test_map_serialism map =
  let l = ref [] in
  let f x =
    let p = Lwt.pause () >>= fun _ -> Lwt.return x in
    assert (List.length !l = 0 || not @@ Lwt.is_sleeping @@ List.hd !l);
    l := p :: !l;
    p
  in
  let p' = map f [0; 1; 2] in
  p' >>= (fun _ -> Lwt.return_true)
```
This test passes if I give it `Lwt_list.map_s`, but fails if I give it `Lwt_list.map_p`. I'm not sure if this totally captures what you were asking for above, though.

aantron: I was thinking something much simpler. I just tried this:

```ocaml
let test_map_serialism map =
  let f2_ran = ref false in

  let f x =
    if x = 1 then
      Lwt.pause () >>= fun () ->
      assert (not !f2_ran);
      Lwt.return ()
    else begin
      f2_ran := true;
      Lwt.return ()
    end
  in

  map f [1; 2] >>= fun _result ->
  Lwt.return true
```

The state is just a boolean. `f 1` starts and puts itself to sleep until the next Lwt "tick." This should give `f 2` a chance to run if it can, and if it does, it will flip the state. So when `f 1` is resumed on the next tick, we just check that the state still hasn't been touched. At the end, for good measure, we could check the result is correct, but that overlaps with other tests already written.

This also passes with `map_s`, but not `map_p`. I didn't add parameters for working with `_rev` variants.

jsthomas:
Closing this and opening a new PR (#535). I've reorganized my code enough that I think it's getting confusing to follow the changes/discussion above.

aantron:
Ok, thank you. It's also possible to force-push into a PR to replace the entire history, and leave a comment, but either way works fine :)

Issue #528, opened 2017-12-26
PPX: remove -no-debug option


aantron:
The `-no-debug` option disables PPX backtrace support (which is currently broken anyway).

In opam, `-no-debug` is used only by `sqlexpr`.

I've never heard of any complaints that backtrace support is too slow. On the contrary, everyone wants backtrace support to be restored and improved in Lwt.

I propose we make `-no-debug` do nothing, simplify the PPX code, and remove the option entirely in some far future version of Lwt.

kandu:
`-no-debug` ease my life when hacking lwt.ppx, the generated code is mush shorter so that the generated parsetree is acceptable to be parsed by my own eyes. 😸

Issue #527, opened 2017-12-26
[easy-ish] PPX: replace [%lwt ...] by [%lwt.catch ...] or remove it


aantron:
The catch-all `[%lwt ...]` syntax has been a source of problems in #307 and #525, due to how PPX works.

In particular, `let%lwt x = e in e'` is represented as `[%lwt let x = e in e']` in the AST, and all the other special `...%lwt` keywords (`try%lwt`, etc.) are represented similarly, i.e. as ordinary, unadorned expressions nested inside an `[%lwt ...]` extension point.

Besides keywords, the Lwt PPX provides that if you have `[%lwt e]`, then it will be expanded to `Lwt.catch (fun () -> e) (fun exn -> Lwt.fail exn)`. However, because this conflicts with interpreting the keywords (above), this works for all expressions except `let`, `try`, etc., and only for them.

This has two problems:

1. It is a bit difficult to predict and reason about. It's probably a bit scary to refactor, as an exception handler could suddenly disappear if the top level becomes a keyword, though it doesn't have many users.
2. It slows us down adding interactions between `%lwt` and additional keywords, such as `begin%lwt` (#525) and `;%lwt` (#307), because giving those keywords a new special interpretation under `%lwt` can break the interpretation of already-written user code.

This issue proposes to delete the catch-all interpretation of `[%lwt ...]` from the PPX. If we need a replacement, I suggest `[%lwt.catch ...]`.

I see `[%lwt ...]` being used in `ocsigen` and `sqlexpr`, perhaps as a result of conversion from Camlp4.

Given how few users there are, I suggest to delete this interpretation entirely.

I think we should do this regardless of whether we merge #525 or #307, because it will simplify Lwt, and give it more options for evolving in the future.

cc @Drup @hcarty @kandu @raphael-proust

Issue #520, opened 2017-12-23
Delete logging support from the PPX


aantron:
`Lwt_log` is deprecated and is being factored out (#484). The PPX has [special support for it](https://ocsigen.org/lwt/api/Ppx_lwt#2_Logging), but I think we should drop that. It doesn't seem to be used by anything in opam (nobody passes `-logs` to the PPX). My guess is that it's not used outside opam, either.

Issue #495, opened 2017-11-02
Remove >> syntax from the PPX


Sventimir:
I have recently spent half an afternoon wondering, why an operator `>>` I defined in a module does not work properly in my code, only to find out that the reason was `lwt.ppx` used in the project. Why is `>>` defined as a syntax extension, when it's just a simple function? It could (and certainly should!) be defined in `Lwt.Infix` as simple as this:

```
let (>>) a b = a >>= fun _ -> b
```

Having it as a syntax extension instead denies users the possibility to define their own `>>` operator, which is idiomatic for any monad, not only Lwt. Besides it's just confusing when somebody (like myself) does not know about it. I would gladly see it changed. In fact I could probably do it myself unless there is a good reason not to, which is why I post here instead of submitting a pull-request directly.

hcarty:
If you do submit a PR to add `>>` please define it as `fun () -> ...` rather than `fun _ -> ...` to help catch bugs!

`>>` from lwt.ppx is considered deprecated, in case that helps, in part because of the issues you mention.

fdopen:
It's defined as a syntax extension because otherwise the evaluation order would be unspecified - and surprising with the current implementation and impure code.
Just compare the output of `Lwt_io.printl "1" >> Lwt_io.printl "2"` with your definition and the syntax extension.

Sventimir:
How is it possible that `bind` does not have that problem, but `>>` defined in terms of `bind` does? It doesn't make much sense to me. Also if it's important to keep `>>` a syntax extension, then why it's already deprecated?

aantron:
@Sventimir The difference is that with `bind`, you have:

```ocaml
Lwt.bind (e1) (fun () -> e2)
```

whereas with `>>`, you have

```
e1 >> e2
```

Because `e2` is not inside a function abstraction (thunk, closure, whatever) in the `>>` case, Lwt can't control its order of evaluation. OCaml decides that, and leaves it unspecified.

And it's not important to keep `>>` as a syntax extension. Everyone, including the authors of it, want to remove it. That's why we deprecated it, and intend to delete it.

I will link to issues when I return to the computer.

Sventimir:
Thank you for the explanation. Indeed, replacing the syntax extension with a function does break a test for sequencing Lwt operations. I suppose I know understand why.

aantron:
[...]

PR #534, opened 2018-01-02
Begin cleaning up the PPX

aantron:
This commit deprecates a lot of little-used PPX features, which are probably also somewhat ill-conceived. This includes all PPX options and several bits of syntax:

- `-no-debug`: There is no point in supporting this (#528). In opam, it is used only in `sqlexpr` (cc @mfp).
- `-no-sequence`, `-no-strict-sequence`: We are removing `>>` anyway (#495), so we should remove these options. They have no users in opam.
- `-log`, `-no-log`: We are deprecating Lwt_log, so it's probably worthwhile to remove logging support from the PPX (#520). These options have no users in opam.
- `[%lwt ...]` syntax expanding to `Lwt.catch ...`. This is causing problems (#527), so we don't want to support it. It is used in `eliom` and `sqlexpr` (cc @mfp @balat @jrochel @vouillon @vasilisp).
- `[%finally ...]` should be replaced by `[%lwt.finally ...]`. The bare `[%finally ...]` is used in `devkit`, `gdb`, `gdbprofiler`, `sqlexpr` (cc @cyberhuman @ygrek @copy @mfp).

Some more notes:

- While editing the PPX, I noticed an undocumented feature: a `let%lwt ...` structure item becomes a call to `Lwt_main.run`. It might be impossible to support this on Node.js in a reasonable way. I'm not sure if we want to deprecate it yet, but it's good that it is not documented.
- `-no-sequence` is the most legitimate of the PPX options. Users that want `>>` as a definable operator may be using it to suppress Lwt's `>>`. I'm not sure if the warning will be too noisy for them.
- This PR doesn't check for `[%lwt let ...]`, just `[%lwt ...]` when it would expand to `Lwt.catch ...`. We should probably disable `[%lwt let ...]` and `[%lwt ...]` around all other constructs, by checking the location of the `%lwt` extension point, as is being done in #525 for `;`. This is still being discussed there, so left for future work.
- @kandu, this PR doesn't include your change about inserting spaces into command line argument descriptions. Whichever order this PR and #525 get merged in, I'll be sure to port/cherry-pick that commit and you will get credit for it :)

Some time after deprecating all this, maybe in `lwt_ppx` 2.0.0, we can delete support for all these features from the PPX, making it much simpler than it is now. A side effect of factoring the PPX out into `lwt_ppx` is that it can have its own release cycle, so we could do this without a major release of the main package `lwt`.

### DIFF
--- a/src/ppx/ppx_lwt.ml
+++ b/src/ppx/ppx_lwt.ml
@@ -48,6 +48,32 @@ let log        = ref false
 let sequence   = ref true
 let strict_seq = ref true
 
+let used_no_debug_option = ref false
+let used_log_option = ref false
+let used_no_log_option = ref false
+let used_no_sequence_option = ref false
+let used_no_strict_sequence_option = ref false
+
+let no_debug_option () =
+  debug := false;
+  used_no_debug_option := true
+
+let log_option () =
+  log := true;
+  used_log_option := true
+
+let no_log_option () =
+  log := false;
+  used_no_log_option := true
+
+let no_sequence_option () =
+  sequence := false;
+  used_no_sequence_option := true
+
+let no_strict_sequence_option () =
+  strict_seq := false;
+  used_no_strict_sequence_option := true
+
 (** let%lwt related functions *)
 
 let gen_name i = lwt_prefix ^ string_of_int i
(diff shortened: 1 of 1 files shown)


aantron:
Also, I believe most of this stuff was included in the PPX to make it more obvious how to convert from the Camlp4 syntax. However, these particular features turned out to be not needed by most users.

copy:
> [%finally ...] should be replaced by [%lwt.finally ...]

May I ask what the rationale is behind this change? I don't mind updating my software, I just prefer the having a less noisy version. A common pattern in ppxs to avoid name clashes is to provide both a prefixed and a non-prefixed version. For example, from ppx_deriving:

> The [%derive.x:] syntax can be shortened to [%x:], given that the deriver x exists and the payload is a type. If these conditions are not satisfied, the extension node will be left uninterpreted to minimize potential conflicts with other rewriters.

aantron:
Just to avoid crowding the namespace. I got the impression that the main reason it exists in unprefixed form is that the Lwt Camlp4 syntax had a `finally` keyword.

I wasn't aware of this being a common PPX pattern, if that's so we can remove that commit. It's also the one I'm least convinced about deprecating.

aantron:
I checked the reverse dependences for `lwt.finally` as well. It's used in `ocurl` and `otetris`. Interestingly, it's also used in `devkit`, and it appears the commit that introduced `lwt.finally` is more recent than the one that introduced `finally`. Anyway, it's not clear that the community prefers either form, largely because the community doesn't seem to use this syntax.

Actually, in all cases except `otetris`, either syntax seems to have been the result of conversion from Camlp4. I was hoping to see if there is a preference either way for writing new code, but the sample is pretty small.

Of course, that only covers code in opam, so I don't really know.

ygrek:
Here is some quick data point : 
```
$ ag --stats-only %lwt.finally
80 matches
41 files contained matches
35910 files searched
$ ag --stats-only %finally
83 matches
31 files contained matches
35910 files searched
```

ygrek:
PS I hope `-no-sequence` stays as long as `>>` is supported so that we can ensure no new `>>` is introduced in the code accidentally.

Drup:
I think `--no-debug` should stay around, mostly because it's very useful when you are doing dev on the ppx itself.

I agree with @copy on `%finally`. Having both a qualified and unqualified version is a common idiom, and I think it's a useful one.

I mostly agree with the rest.

aantron:
@Drup, can you say how `-no-debug` helps with developing the PPX? I don't see what the benefit is.

I find the existence of `-no-debug` to be a hindrance to developing the PPX, as conditional code is needed because of it.

aantron:
Also I believe @kandu mentioned that it's easier to work on the PPX with `-no-debug`.

aantron:
Okay, I think I will merge this without the commit about deprecating `[%finally]`.

@Drup: `-no-debug` will be deprecated, but it will remain for now. Hopefully, we will (have time to) figure out what's happening with stack traces soon, then we can decide what exactly to do with it. If we can't get backtraces to work using the PPX, we won't be emitting `backtrace_` functions anyway, and the output will be simple too.

@ygrek: Indeed, we won't remove `-no-sequence` until `>>` is removed, and it will stick around for a while after that, too, as a no-effect option, to avoid breaking builds.

Drup:
Well, `-no-debug` makes the code produced by `-dsource`/`-dparsetree` readable. :]

Is the conditional code such an issue ? It's fairly well localized and not extremely complicated.

aantron:
I don't think either the conditional code nor complex `-dsource` output is a big deal, no. I'd like to remove `-no-debug` from the docs and discourage anyone from passing it "in production" for now.

Drup:
> I'd like to remove -no-debug from the docs and discourage anyone from passing it "in production" for now.

That's totally fair. :)

PR #535, opened 2018-01-06
Issue 395: Verify Lwt_list functions run in parallel/series.

jsthomas:
This change proposes tests to address the second of the three parts of #395. These tests check that the `Lwt_list` functions ending in `_p` run in parallel and the `_s` tests run in series.


aantron:
Great, these work very well! I tried messing with them by flipping the `_p` and `_s`.

PR #538, opened 2018-01-15
Issue 395: Add tests for stack safety in Lwt_list.

jsthomas:
This change proposes new tests to verify that functions in `Lwt_list` are stack safe, which addresses the third and final part of #395.

The first commit adds new tests, applying each function in `Lwt_list` to a list of one million elements. (I also updated the labels on tests in #535, to make the terminology a little more consistent; now we use the term "serialization" everywhere). These tests do not pass for several of the `Lwt_list` functions.

The second commit updates `Lwt_list` to make the tests pass. I've tried to introduce the `[@ocaml.tailcall]` where appropriate to make our use of tail recursion more explicit.

### DIFF
--- a/test/core/test_lwt_list.ml
+++ b/test/core/test_lwt_list.ml
@@ -147,11 +147,6 @@ let test_serialization ?(rev=false) map =
   let p = map f [0; 1] in
   p >>= (fun _ -> Lwt.return_true)
 
-let test_big_list m =
-  let make_list n = Array.to_list @@ Array.init n (fun x -> x) in
-  let f _ = Lwt.return () in
-  m f (make_list 1_000_000) >>= (fun _ -> Lwt.return_true)
-
 let test_for_all_true f =
   let l = [true; true] in
   f (fun x -> Lwt.return (x = true)) l
@@ -204,7 +199,7 @@ let test_rev_map f =
   f (fun n -> Lwt.return (n * 2)) l >>= fun after ->
   Lwt.return (after = [6; 4; 2])
 
-let suite = suite "lwt_list" [
+let suite_primary = suite "lwt_list" [
   test "iter_p" begin fun () ->
     test_iter Lwt_list.iter_p [1; 0; 1];
     test_exception Lwt_list.iter_p;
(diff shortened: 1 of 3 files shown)


test/core/test_lwt_list.ml
```diff
@@ -575,9 +580,132 @@ let suite = suite "lwt_list" [
...
+    let m f l = Lwt_list.fold_right_s (fun x _ -> f x >>= fun _ -> Lwt.return ()) l () in
+    test_big_list m
+  end;
+
+  test "filter_map_p big list" begin fun () ->
+    let m f = Lwt_list.filter_map_p (fun x -> f x >>= fun u -> Lwt.return (Some u)) in
+    test_big_list m
+  end;
+
+  test "filter_map_s serlialism" begin fun () ->
```
aantron: I think the title is wrong here :)

test/core/test_lwt_list.ml
```diff
@@ -575,9 +580,132 @@ let suite = suite "lwt_list" [
...
+    let m f = Lwt_list.filter_p (fun x -> f x >>= fun _ -> Lwt.return_true) in
+    test_big_list m
+  end;
+
+  test "filter_s big list" begin fun () ->
+    let m f = Lwt_list.filter_s (fun x -> f x >>= fun _ -> Lwt.return_true) in
+    test_big_list m
+  end;
+
+  test "filter_map_p big list" begin fun () ->
```
aantron: I think this test and the next one are duplicates of two tests above.

aantron:
If you do want to make the tail-recursion suite conditional though, the tester's `~only_if` argument should be helpful:

https://github.com/ocsigen/lwt/blob/4f0dabca5228b36a65ea7e49375a15d2070161fa/test/test.mli#L40-L41

test/core/test_lwt_list.ml
jsthomas: Good catch, thanks.

test/core/test_lwt_list.ml
```diff
@@ -585,7 +580,15 @@ let suite = suite "lwt_list" [
...
   end;
+]
 
+let test_big_list m =
+  let make_list n = Array.to_list @@ Array.init n (fun x -> x) in
+  let f _ = Lwt.return () in
+  m f (make_list 10_000_000) >>= (fun _ -> Lwt.return_true)
+
+let suite_intensive = suite "lwt_list big lists"
+    ~only_if:(fun () -> Sys.getenv_opt "LWT_STRESS_TEST" <> None) [
```
aantron: `getenv_opt` requires 4.05, so this won't compile on most versions of OCaml Lwt still supports.

aantron:
Thanks! I modified `.travis.yml` in a commit right after the PR, to set the environment variable.

Issue #475, opened 2017-09-28
Jenkins builds appearing as stuck


aantron:
Looking at [recent PRs](https://github.com/ocsigen/lwt/pulls), the Jenkins status is pending on the last three (#474, #471, #469), and has been so for weeks. @vasilisp could this be related to the recent infrastructure changes that the Ocsigen website also suffered from?

aantron:
In addition, it looks like Jenkins is no longer running on `master`, and therefore not updating the ocsigen.org manual. For example, compare this commit https://github.com/ocsigen/lwt/commit/af91fad3c07223bf11e917daa2bbd3b283acdcad, and the dev manual at https://ocsigen.org/lwt/dev/api/Lwt_log. Looking at ocisgen.org-data, I see the [last commit](https://github.com/ocsigen/ocsigen.org-data/commits/master) was on July 27.

aantron:
I disabled the `ci.inria.fr` webhooks for now. Not sure if that is sufficient to stop the stuck Jenkins status from appearing, but I hope so.

They webhooks are still there, so they can be easily re-enabled whenever the Jenkins setup is getting fixed/debugged.

aantron:
Closing this in favor of tracking elsewhere, e.g. https://github.com/ocsigen/ocsigen.org-data/issues/13.

PR #542, opened 2018-01-16
Lwt_sequence: remove check for active node in length function

cedlemo:
@aantron,

while I was removing the check, I was asking myself if when a sequence is shared/used in two threads, how can I be sure that the `Lwt_sequence.length` is "atomic". For example, it is hard in the test to : 

* create a sequence
* create a thread that use `Lwt_sequence.length` on this sequence
* create another thread that remove one or more node in this sequence
* use Lwt.join on the 2 threads

but we can imagine that it could happen? 

or is there some kind of Lwt "magic" that sets some "mutex" or something like that in order to garantee that operation like `Lwt_sequence.length` is atomic.

### DIFF
--- a/src/core/lwt_sequence.ml
+++ b/src/core/lwt_sequence.ml
@@ -69,11 +69,7 @@ let length seq =
     if curr == seq then
       len
     else
-      let node = node_of_seq curr in
-      if node.node_active then
-        loop node.node_next (len + 1)
-      else
-        loop node.node_next len
+      let node = node_of_seq curr in loop node.node_next (len + 1)
   in
   loop seq.next 0
 


aantron:
@cedlemo, Lwt has promises instead of threads, and all it ever does is run callbacks that are attached to those promises. Callbacks are always called in the main thread. Once Lwt calls a callback, the callback runs to completion, like any other OCaml function call. So, there is no way for Lwt to "interrupt" `Lwt_sequence.length`. You may find the introductory material here useful: https://ocsigen.github.io/lwt/manual-draft/Lwt.html.

Of course, this kind of race condition is possible with real preemptive threads, but we don't promise that `Lwt_sequence` is thread-safe. Indeed, even the existing `if`-check isn't safe with preemptive threads. Lwt itself uses `Lwt_sequence` only from the main thread.

cedlemo:
> You may find the introductory material here useful: https://ocsigen.github.io/lwt/manual-draft/Lwt.html.

yes totally usefull, thanks you. Maybe it was just me, but I did not find a link to this manual in https://ocsigen.org/lwt/3.1.0/manual/.

aantron:
Yes, the Ocsigen docs aren't being updated (#475). We have to create a new docs generating and hosting process.

cedlemo:
FYI in the introduction, there is a bad link:

> After that is the reference proper, which goes into painful levels of detail on every single type 

reference proper points to https://ocsigen.github.io/lwt/manual-draft/2_Fundamentals instead of  
https://ocsigen.github.io/lwt/manual-draft/Lwt.html#2_Fundamentals

aantron:
Thanks, I've fixed it in `master`, though I probably won't re-upload the rendered draft.

Issue #455, opened 2017-07-19
[easy] Document that Lwt_io.printf is more like sprintf than printf


aantron:
From https://discuss.ocaml.org/t/lwt-execution-order-and-parallelism/564/8:

`Lwt_io.printf` calls `ksprintf` internally. This has some surprising consequences:

- ~~`%!` doesn't really flush anything, since no channel is associated with a `ksprintf`.~~ *Addressed by @smithjessk in #544*.
- The [type is weird](https://ocsigen.org/lwt/dev/api/Lwt_io#VALprintf):

    ```ocaml
    val printf : ('a, unit, string, unit Lwt.t) Pervasives.format4 -> 'a
    ```

    Compare [`printf`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Printf.html#VALprintf) and [`sprintf`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Printf.html#VALsprintf):

    ```ocaml
    val printf : ('a, out_channel, unit) format -> 'a
    val sprintf : ('a, unit, string) format -> 'a
    ```

    It should look like `printf,` but it looks more like `sprintf`.

    I don't remember the details now, but I vaguely recall someone getting bitten, or at least nibbled, by this.

These are especially annoying, if one tries to refactor non-Lwt code to Lwt. The different semantics make that code transformation slightly dangerous. These things also came up when discussing #310 on IRC, but are not explicitly listed in that issue.

I don't know if this can/should be fixed somehow – it's probably a hard issue. But we should at least [document in the `.mli`](https://github.com/ocsigen/lwt/blob/540ac3d86db4c06d62793993c3372151a9ef221a/src/unix/lwt_io.mli#L363) file that `%!` does nothing, and you have to use `Lwt_io.(flush channel)` if you want to flush it. This part is relatively easy. It's also worthwhile to look at the other `printf`-style functions nearby, and see if they actually correspond to the ones in module `Printf` the way users expect, which `Lwt_io.printf` doesn't.

aantron:
Also cc @Drup, due to considerable experience with `Format`, etc. :)

Drup:
So, yes, this is pretty much what I meant in #310 (except, knowing Format/Printf too much, I didn't detailed the actual consequences of my various points).

I believe my alternative implementation posted in #310 solves most problems.

aantron:
[...]

PR #544, opened 2018-02-08
Document Lwt_io.printf flushing behavior

smithjessk:
As described in #455.

### DIFF
--- a/src/unix/lwt_io.mli
+++ b/src/unix/lwt_io.mli
@@ -362,10 +362,10 @@ val fprint : output_channel -> string -> unit Lwt.t
 val fprintl : output_channel -> string -> unit Lwt.t
 
 val fprintf : output_channel -> ('a, unit, string, unit Lwt.t) format4 -> 'a
-(** [%!] does nothing here. To flush the channel, use {!flush} [ channel]. *)
+(** [%!] does nothing here. To flush the channel, use [Lwt_io.flush channel]. *)
 
 val fprintlf : output_channel -> ('a, unit, string, unit Lwt.t) format4 -> 'a
-(** [%!] does nothing here. To flush the channel, use {!flush} [ channel]. *)
+(** [%!] does nothing here. To flush the channel, use [Lwt_io.flush channel]. *)
 
 val print : string -> unit Lwt.t
 val printl : string -> unit Lwt.t
--- a/src/unix/lwt_io.mli
+++ b/src/unix/lwt_io.mli
@@ -371,19 +371,20 @@ val print : string -> unit Lwt.t
 val printl : string -> unit Lwt.t
 
 val printf : ('a, unit, string, unit Lwt.t) format4 -> 'a
-(** [%!] does nothing here. To flush the channel, use [Lwt_io.(flush Lwt_io.stdout)]. *)
+(** [%!] does nothing here. To flush the channel, use [Lwt_io.(flush stdout)].
+*)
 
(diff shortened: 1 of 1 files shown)


src/unix/lwt_io.mli
```diff
@@ -360,16 +360,30 @@ val write_value : output_channel -> ?flags : Marshal.extern_flags list -> 'a ->
 
 val fprint : output_channel -> string -> unit Lwt.t
 val fprintl : output_channel -> string -> unit Lwt.t
+
 val fprintf : output_channel -> ('a, unit, string, unit Lwt.t) format4 -> 'a
+(* %! does nothing here. To flush the channel, use Lwt_io.(flush channel) *)
```
aantron: Hey, just in case: to make this an [ocamldoc comment](https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html#sec349), it should be:

```
(** [%!] does nothing here. To flush the channel, use [Lwt_io.flush channel]. *)
```

The `(**` makes ocamldoc pick it up (so it is included in the generated manual), and the `[...]` is the ocamldoc source code style. I also took the liberty of not opening `Lwt_io` in `channel`, because that is likely to be a user's identifier, rather than something from `Lwt_io` when someone is using `fprintf`. But for `print*f` and `eprint*f`, it should probably be with `Lwt_io` as you have it, and `Lwt_io.stdout` and `Lwt_io.stderr`, respectively, rather than a user's identifier "`channel`".

smithjessk: Sorry, but I'm not sure what you mean by "not opening `Lwt_io` in `channel`".

aantron: Oh, the `Lwt_io.(foo bar)` syntax is a "local open," it brings the identifiers of `Lwt_io` in scope for both `foo` and `bar`.

So, for stdout, we probably want something like `Lwt_io.(flush stdout)`, because that means "take both `flush` and `stdout` potentially from `Lwt_io` (or anything else that is in scope)." 

But for a user's custom channel, `Lwt_io.flush channel`, because only `flush` here would come from `Lwt_io`.

smithjessk: Ah, sorry; I just saw this comment. Thanks for the explanation.

I addressed this in 5cd987d5b99db37091e64b5c65b111311569813b. Does it look okay?

In f103d781161b9975215809e3f84bf090a0a5c7e5, I linked to `Lwt_io.flush` for the `fprint*f` docs. I tried to do the same for `print*f` and `eprint*f`, but I couldn't find a way to link to the identifiers `flush` and `stdout` in docs of the form `Lwt_io.(flush stdout)`. Am I missing something?

aantron: Thank you, it looks good now :) For the cross-reference links, I'd say just not to use them, because, as you can see, the support for including cross-references inside expressions is lacking, and you have to manually try to guess where to insert spaces around and inside `[...]`, etc. It's difficult for most people to predict how that will be displayed, and it makes the `.mli` file hard to read.

The rule I use for myself is:

- If it's a single identifier on its own, try to get it cross-referenced with `{!...}`.
- If it's a non-trivial expression like `Lwt_io.flush foo`, don't try to get anything inside it cross-referenced, and just write it verbatim.

So, my suggestion is to omit f103d78.

smithjessk: cc9328e reverts f103d78.

aantron:
Thank you!

smithjessk:
My pleasure! Thanks for the comments.

PR #547, opened 2018-02-13
Documentation improvements: Lwt_main.run and Lwt_io.wait_read/wait_write

dmbaturin:
Myself I've found the documentation for them confusingly incomplete and attempted to not use the former when necessary and use the latter when not necessary.

### DIFF
--- a/src/unix/lwt_main.mli
+++ b/src/unix/lwt_main.mli
@@ -28,8 +28,9 @@ val run : 'a Lwt.t -> 'a
       and returns the value of [p] if it is fulfilled. If [p] is rejected with
       an exception, that exception is raised.
 
-      Every native or bytecode program that uses Lwt should always use this function
-      for evaluating a promise at the top level (such as its main function or main loop),
+      Every native or bytecode program that uses Lwt should always use
+      this function for evaluating a promise at the top level
+      (such as its main function or main loop),
       otherwise promises that depend on I/O operations will not be resolved.
 
       Example:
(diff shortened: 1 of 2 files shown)


src/unix/lwt_main.mli
```diff
@@ -28,6 +28,18 @@ val run : 'a Lwt.t -> 'a
...
       an exception, this exception is raised.
 
+      Every program that uses Lwt should always use this function to launch
+      their threads, otherwise correct scheduling is not guaranteed.
+
+      Example:
+      {[
+        let main () = Lwt_io.write_line Lwt_io.stdout "hello world"
+
+        let _ =
```
hcarty: It may be better to use `let () =` here to encourage more a more precise definition of what's being returned.

src/unix/lwt_main.mli
```diff
@@ -28,6 +28,18 @@ val run : 'a Lwt.t -> 'a
       then returns the value returned by the thread. If [t] fails with
       an exception, this exception is raised.
 
+      Every program that uses Lwt should always use this function to launch
+      their threads, otherwise correct scheduling is not guaranteed.
```
hcarty: If I understand correctly, this is not true when targeting Javascript via js_of_ocaml.  It is true for native and bytecode compilation.

aantron: Indeed, JS environments have the equivalent of `Lwt_main.run` built into them, conversely `Lwt_main.run` isn't even available when targeting JS.

aantron:
Thanks :)

Issue #310, opened 2017-01-09
Lwt_io, Printf and Format


Drup:
There are two issues with `Lwt_io`:

- It needs unix. but the `printf` part should not.
- All the `printf` functions are basically broken. They rely on `Printf` instead of `Format`, which has become the standard in OCaml pretty printing, and they do not handle the `%a` very well.

My current idea is to:

- Split Lwt_io in two parts, one unix-independent (`Lwt_format` ?)
- Deprecate `Lwt_io.print*`
- Reimplement all the printing functions in term of formats. There are some concerns with blocking, but I have some ideas to solve them.

I think this process would respect #293 and allows us to improve the state of lwt-based printing.

aantron:
I always had the impression that a lot of the functions in `Lwt_io` should be in a module that corresponds to the I/O functions of `Pervasives` and does not depend on `Lwt_unix` – not just the formatting functions.

Drup:
@aantron maybe we should create new functions in `Lwt_unix` for the stuff that should be in `Lwt_io`, deprecate them, and move `Lwt_io` in `lwt` after the deprecation ?

aantron:
Something to that effect. I am not sure if `Lwt_unix` is necessarily the right destination, but we can use a deprecation to warn about removal of functions from `Lwt_io`.

hcarty:
@Drup `Lwt_fmt` maybe?  For the `Format`-focused code?

Drup:
I was wondering if my idea would work, so here is an implementation: https://gist.github.com/Drup/2698c69fd70cf526ba3164b135177bcd

aantron:
The biggest part of this was done by @Drup in #548.

PR #548, opened 2018-02-14
Lwt_fmt

Drup:
This is a prelude to solving #310. This adds a new module `Lwt_fmt` which exposes some lwt-powered printf function in the style of Format. The functions return promises and are compatible with all the Format/Fmt combinators. The implementation is not complicated, but uses some fairly subtle Format tricks. I added some tests to ensure that the behavior of flushing and composition was correct.

For now, the module is in `lwt_unix`. I believe this should not be the case, but I expect @aantron to have better ideas on how and what to split where.

cc @octachron @c-cube (@dbuenzli) as fellow Format fetishists.

### DIFF
--- a/src/unix/Lwt_fmt.mli
+++ b/src/unix/Lwt_fmt.mli
@@ -65,18 +65,24 @@ val make_formatter :
     functions to update the underlying channel.
 *)
 
+val get_formatter : formatter -> Format.formatter
+(** [get_formatter fmt] returns the underlying {!Format.formatter}. 
+    To access the underlying formatter during printing, it is
+    recommended to use [%t] and [%a].
+*)
+
 (** {2 Printing} *)
 
 val fprintf : formatter -> ('a, Format.formatter, unit, unit Lwt.t) format4 -> 'a
 
 val kfprintf :
-  (Format.formatter -> unit Lwt.t -> 'a) ->
+  (formatter -> unit Lwt.t -> 'a) ->
   formatter -> ('b, Format.formatter, unit, 'a) format4 -> 'b
 
 val ifprintf : formatter -> ('a, Format.formatter, unit, unit Lwt.t) format4 -> 'a
 
 val ikfprintf :
-  (Format.formatter -> unit Lwt.t -> 'a) ->
+  (formatter -> unit Lwt.t -> 'a) ->
   formatter -> ('b, Format.formatter, unit, 'a) format4 -> 'b
 
 val flush : formatter -> unit Lwt.t
--- a/src/unix/Lwt_fmt.mli
+++ b/src/unix/Lwt_fmt.mli
@@ -1,0 +1,98 @@
+(* OCaml promise library
+ * http://www.ocsigen.org/lwt
+ * Copyright (C) 2018 Gabriel Radanne
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, with linking exceptions;
+ * either version 2.1 of the License, or (at your option) any later
+ * version. See COPYING file for details.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA
+ * 02111-1307, USA.
+ *)
+
+(** Format API for Lwt-powered IOs *)
+
+(** This module bridges the gap between [Format] and [Lwt].
+    Although it is not required, it is recommended to use this module with the
+    [Fmt] library.
+
+    Compared to regular formatting function, the main difference is that
+    printing statements will now return promises instead of blocking.
+*)
+
+val printf : ('a, Format.formatter, unit, unit Lwt.t) format4 -> 'a
+(** Returns a promise that prints on the standard output. 
+    Similar to [Format.printf]. *)
+
+val eprintf : ('a, Format.formatter, unit, unit Lwt.t) format4 -> 'a
+(** Returns a promise that prints on the standard error. 
+    Similar to [Format.eprintf]. *)
+
+(** {1 Formatters} *)
+
+type formatter
+(** Lwt enabled formatters *)
+
+type order =
+  | String of string * int * int (** [String (s, off, len)] indicate the output of [s] at offset [off] and length [len]. *)
+  | Flush (** Flush operation *)
+
+val make_stream : unit -> order Lwt_stream.t * formatter
+(** [make_stream ()] returns a formatter and a stream of all the writing
+    order given on that stream.
+*)
+
+
+val of_channel : Lwt_io.output_channel -> formatter
+(** [of_channel oc] creates a formatter that writes to the channel [oc]. *)
+
+val stdout : formatter (** Formatter printing on {!Lwt_io.stdout}. *)
+val stderr : formatter (** Formatter printing on {!Lwt_io.stdout}. *)
+
+val make_formatter :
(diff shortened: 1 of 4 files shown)


src/unix/Lwt_fmt.mli
```diff
@@ -0,0 +1,98 @@
...
+    Although it is not required, it is recommended to use this module with the
+    [Fmt] library.
+
+    Compared to regular formatting function, the main difference is that
+    printing statements will now return promises instead of blocking.
+*)
+
+val printf : ('a, Format.formatter, unit, unit Lwt.t) format4 -> 'a
+(** Returns a promise that prints on the standard output. 
+    Similar to [Format.printf]. *)
```
dbuenzli: `{!Format.printf}`

src/unix/Lwt_fmt.mli
```diff
@@ -0,0 +1,98 @@
...
+
+val eprintf : ('a, Format.formatter, unit, unit Lwt.t) format4 -> 'a
+(** Returns a promise that prints on the standard error. 
+    Similar to [Format.eprintf]. *)
```
dbuenzli: `{!Format.eprintf}`

src/unix/Lwt_fmt.mli
```diff
@@ -0,0 +1,98 @@
...
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA
+ * 02111-1307, USA.
+ *)
+
+(** Format API for Lwt-powered IOs *)
+
+(** This module bridges the gap between [Format] and [Lwt].
+    Although it is not required, it is recommended to use this module with the
+    [Fmt] library.
```
dbuenzli: These toplevel module names could be `{!links}`.

src/unix/Lwt_fmt.mli
```diff
@@ -0,0 +1,98 @@
...
+    {!Format.formatter} [fmt]. The [commit] function will be called by the printing
+    functions to update the underlying channel.
+*)
+
+(** {2 Printing} *)
+
+val fprintf : formatter -> ('a, Format.formatter, unit, unit Lwt.t) format4 -> 'a
+
+val kfprintf :
+  (Format.formatter -> unit Lwt.t -> 'a) ->
```
Octachron: I think it would make sense to have `(formatter -> 'a)` rather than `(Format.formatter -> unit Lwt.t -> 'a)` for the continuation argument:  with `write_pending` exposed, it is still possible to write `fprintf` from the `formatter -> 'a` version and it avoids exposing the underlying `Format.formatter` of the `formatter` argument.

hakuch:
For what reason would I want the formatting functions to return promises?

Drup:
@hakuch The underlying channel of

hakuch:
Thanks, @Drup. That makes sense!

aantron:
This seems pretty nice. My only real concern is where it should go. I don't mind ultimately adding a new module to Lwt, but what are people's opinions on that, as opposed to:

- A separate repo,
- Somewhere inside `Lwt_io`,
- elsewhere?

src/unix/Lwt_fmt.mli
```diff
@@ -0,0 +1,98 @@
...
+
+
+val of_channel : Lwt_io.output_channel -> formatter
+(** [of_channel oc] creates a formatter that writes to the channel [oc]. *)
+
+val stdout : formatter (** Formatter printing on {!Lwt_io.stdout}. *)
+val stderr : formatter (** Formatter printing on {!Lwt_io.stdout}. *)
+
+val make_formatter :
+  commit:(unit -> unit Lwt.t) -> fmt:Format.formatter -> formatter
```
c-cube: maybe add a `()` argument at the end, so future versions can add optional arguments?

c-cube:
This looks cool. I think it should go in the core lwt library, provided it's only accessed through module aliases (and thus is of minimal overhead for people who don't use it).

hcarty:
I like it as a top-level module slightly more than in `Lwt_io`. There are no extra requirements so it would be nice to keep it in the main Lwt repository.

src/unix/Lwt_fmt.mli
Drup: Hum, that is right, but I'm pretty sure people will forget to call `write_pending` if you don't do it for them, which will lead to very weird bugs.

I'm trying to think if exposing the `Format.formatter` is a bad thing or not. It's not *really* unsafe, but I'm not sure if it's useful or not.

src/unix/Lwt_fmt.mli
Drup: Good idea.

src/unix/Lwt_fmt.mli
Octachron: In term of coherency of the API, I think that in any case it would be better to expose a `formatter` rather than `Format.formatter` otherwise `kfprintf`  provides a strange entry point to features only available for `Format.formatter` (e.g. modifying margin, tags).

I agree that people will probably forget the call to `write_pending`, even if this could be mitigated by adding `Do not forget to call {!write_pending}` to the documentation and the limited amount of people using kprintf without reading the documentation.

Drup: Or maybe we can keep the `unit Lwt.t` and just expose a `formatter`.

Do you think I need to expose a function to retrieve the `Format.formatter` too ?

Octachron: Yep, I think that `formatter -> unit Lwt.t -> `a` is a good compromise.

> Do you think I need to expose a function to retrieve the Format.formatter too ?

I think that It would be nice to avoid it. I believe that letting `make_formatter` handle the configuration of the formatter is fine since *safely* modifying the underlying formatter is as hard as *safely* creating a second formatter with the same output channel. Nevertheless, in this case, it might be interesting to add the part of the `Format` API not exposed to `kfprintf`, i.e. the tabulation boxes and `pp_print_if_newline`.

Drup: Isn't that point sort of moot, given the existence of `%t` and `%a`? `Lwt_fmt.fprintf fmt "%t" (fun fmt -> ....)` will always work.

Octachron: True,`%t` and `%a` should be enough for tabulation boxes and friends.

aantron:
Very well. Perhaps this is already proposed, but I suggest we merge this as is, putting it into `lwt.unix`. We can move it to the core `lwt` if/when `Lwt_io` is untangled. That change should be backwards-compatible for any users (if we have a release): anyone who uses `Lwt_fmt` will have to depend on `lwt.unix`, which depends on `lwt`, so either way people will be pulling in a package with the `Lwt_fmt` module.

Drup:
@aantron That seems like a good direction for me. Let me fix what I discussed with @Octachron above and we will be good to go! :)

Drup:
I just added a new commit that make the API a bit more regular, as discussed with @Octachron . If CI is happy, this is ready to be merged as far as I'm concerned.

aantron:
Ok, thank you!

rgrinberg:
Something that's unrelated but still important:

Is there a plan for Lwt to move away from having multiple top level modules per library? It would be really nice if Lwt switched to using `(wrapped true)` or made all modules apart from `Lwt` internal (the way dbuenzli does it).

The reason why I'm raising this here is that this PR adds yet another toplevel module which is a bit unfortunate.

Drup:
@rgrinberg lwt has multiple distinct libraries. If you were to wrap everything, `Lwt_fmt` would become `Lwt_unix.Fmt`, not `Lwt.Fmt`. In the end, it would be strictly more inconsistent than the current situation, and I don't see which gain you would have (all the modules are already prefixed).

rgrinberg:
> If you were to wrap everything, Lwt_fmt would become Lwt_unix.Fmt, not Lwt.Fmt

To me this is a good thing. Yes, this can look a little less aesthetically pleasing, but it streamlines understanding new code for people as you no longer have to think about which library introduces a particular module.

aantron:
I want to reorganize the module structure of Lwt eventually, but I have no concrete plan to do it at the moment.

I think putting modules now in `lwt.unix` under `Lwt.` can be done with module aliases, but, we may have redesigned the Unix binding by then anyway (#328).

PR #552, opened 2018-02-28
Make (Lwt.nchoose_split []) immediately return

### LINKED ISSUE #551: Lwt.nchoose_split hangs indefinitely with an empty list
The following will hang indefinitely:

```ocaml
let l : unit Lwt.t list = []

let _, _ = Lwt_main.run (Lwt.nchoose_split [])
```

Tested with Lwt 3.2.1 and Lwt's `master` branch as of b27bea7c4c57de0ff729fc07f6862852b573a45d.

Ideally `nchoose_split` would immediately return `[], []` or, at worst, fail.

hcarty:
Fixes #551 

This is a change in (undocumented) behavior.  For the better in my opinion but I'm not sure how others feel.

### DIFF
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -2921,6 +2921,10 @@ struct
 
     let p = check_for_already_resolved_promises [] ps in
     p
+
+  let nchoose_split = function
+    | [] -> return ([], [])
+    | l -> nchoose_split l
 end
 include Concurrent_composition
 
--- a/test/core/test_lwt.ml
+++ b/test/core/test_lwt.ml
@@ -2037,7 +2037,7 @@ let suites = suites @ [nchoose_tests]
 let nchoose_split_tests = suite "nchoose_split" [
   test "empty" begin fun () ->
     let p = Lwt.nchoose_split [] in
-    Lwt.return (Lwt.state p = Lwt.Sleep)
+    Lwt.return (Lwt.state p = Lwt.Return ([], []))
   end;
 
   test "some fulfilled" begin fun () ->


aantron:
@hcarty What motivates this change?

Issues:

- If we do make this change, we should probably also make it for `nchoose`, `npick`, `pick`, and `choose`, i.e. all the other variants of `race` in Lwt.
- Starting reasoning with `choose`, it seems "correct" that waiting for the first promise to resolve in an empty list of promises would result in a promise that is pending forever. In the case of `pick` and `npick`, it is still possible to resolve these promises by canceling them, so it is possible that there are scenarios where waiting on an empty list of promises is not a programming mistake. Cancelation doesn't apply to `nchoose_split`, but it would be weird to have a different behavior here. I think the next-most reasonable thing is to reject the promise with some exception, however.
- I can't actually think of a good example where the new behavior is bad, but I'm imagining some server that blindly consumes the result lists of `nchoose_split` in a loop. If we make this change, we will turn a thread of execution that became idle and was hanging forever into a busy-loop.
- Despite the above, you may well be one of very few people using `nchoose_split`, so maybe we can just change it.

aantron:
One more thing to note is that the current behavior is consistent with JS [`Promise.race`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race):

> If the iterable passed is empty, the promise returned will be forever pending.

And there is some kind of fuzzy algebraic duality between `join` on an empty list resolving immediately, and `choose` resolving never.

hcarty:
@aantron All good points.  I ran into this (thankfully during early testing) with a program which was handling a capped number of concurrent promises pulled from a queue.  The program would hang forever once it ran out of things to do, even once new tasks showed up.  I was using `nchoose_split` to track the number of completed vs outstanding tasks promises.

As a simplified example, imagine `fetch_jobs_to_do` in the following code may not return promises for some period of time due to a starved queue, then start returning promises again after all pending promises have been resolved:
```ocaml
(* Loop forever, pulling from some queue of promises *)
let rec loop promises max_concurrency =
  (* Run until something finishes *)
  let%lwt finished, pending = Lwt.nchoose_split promises in
  (* Get more work to do.  Note that new_promises could be an empty list. *)
  let new_promises = fetch_jobs_to_do ~number_of_jobs:(max_concurrency - List.length pending) in
  (* Repeat with the new promises and leftovers from the nchoose_split call *)
  loop (List.concat [new_promises, pending]) max_concurrency
```
If the `promises` argument ends up empty this recursive loop will hang forever on the `nchoose_split` line.

hcarty:
The example above and the code which highlighted this issue use `nchoose_split` precisely because it lets the user know _which_ promises are unresolved without lots of extra bookkeeping on the user's part.  `Promise.race` does not allow that if I am reading the documentation correctly.

aantron:
> The example above and the code which highlighted this issue use nchoose_split precisely because it lets the user know which promises are unresolved without lots of extra bookkeeping on the user's part. Promise.race does not allow that if I am reading the documentation correctly.

Yes, indeed :)

It seems like in order for the program above to call `nchoose_split` with no promises, both `new_promises` and `pending` must have been empty on the previous iteration of the loop. That means `fetch_jobs_to_do` is capable of evaluating to the empty list when it is asked to retrieve a nonzero number of jobs (since `max_concurrency` is presumably greater than zero). Doesn't that seem like a bug?

If the above reasoning is correct, then it seems that `fetch_jobs_to_do` is a polling function rather than a waiting function (also suggested by not using `let%lwt`). Won't changing `nchoose_split` to return empty lists can actually send this `fetch_jobs_to_do` into a polling loop of some kind? Since the only place where an Lwt callback is attached is the `let%lwt` in the first line of the body, but it is going to be on an already-resolved promise containing two empty lists, the immediate binding semantics of current Lwt will prevent anything else from interleaving with this polling loop, until the next time `fetch_jobs_to_do` returns a pending promise.

Also @Drup, others, could you, if you have time, comment on the API issues of changing `nchoose_split`, with regard to its similarity to `pick`, `choose`, etc.?

aantron:
Another note is that users already have to watch out for the corner case of calling any of the `pick`/`choose` functions on an empty list, although I suppose `pick`/`choose` are more likely to be called on lists of hardcoded length visible in the syntax, than on arbitrary lists returned by functions, relative to how `nchoose_split` would probably be used. In any case, it's a weak argument in favor of the user already being familiar with the danger :)

aantron:
....and scratch my comment about being able to cancel `Lwt.pick []`. None of these functions produce cancelable promises when given empty lists, because the cancelation depends on being able to cancel at least one of the promises in the list.

hcarty:
My apologies - I oversimplified the example code.  The real codebase's equivalent of `fetch_jobs_to_do` does return a promise.  It's reading from an external queue which may be empty.  If that queue is empty the promise wraps/evaluates to (whatever the proper terminology is) an empty list.

hcarty:
In all of these cases I think hanging forever seems dangerous.  It's an easy way to end up with a dead path in a program with no simple way to identify why the hang occurs.

aantron:
@hcarty Still, if that external queue is empty, won't returning empty lists result in a busy-loop? I suppose that's easier to debug, since at least you might get a semi-useful backtrace if you interrupt it. It seems that rejecting the promise with some exception like `Lwt.Empty_promise_list` might be the best thing to do. It seems somewhat reasonable to say passing an empty list is always a programming error.

aantron:
Some discussions that may be relevant:

- https://github.com/domenic/promises-unwrapping/issues/75
- https://github.com/reactphp/promise/issues/57
- http://bluebirdjs.com/docs/api/promise.race.html

aantron:
After thinking about it more, I think the best thing to do is to reject with `Invalid_argument`. Calling `pick`/`choose`/`race` on an empty list does seem like a programming error, and it's best to catch this and fail as early as possible.

**The alternatives**:

- Returning a forever-pending promise, as now, seems pretty useless, and difficult to debug, as you pointed out. I doubt anyone actually wants that behavior in practice.
- Returning empty lists seems kind of arbitrary, and carries another danger, the danger of busy-loops.

**Concerning algebra**, I doubt anyone relies on the algebraic properties of `pick`/`choose`/`race`. But if there are such users, they are the ones who are already thinking about corner cases, identity elements, associativity, etc. They will quickly find out that `pick`/`choose`/`race` doesn't do what is "algebraically correct" on the empty list, especially since we will document it, but they can get "algebraic" versions of these functions by simply wrapping them and checking for the empty list.

On the other hand, the much more common "casual" user of `pick`/`choose`/`race` is less likely to be thinking about the corner case, so we shouldn't make it difficult to discover or debug.

<br/>

The main remaining concern is this is technically a breaking change :p I expect that nobody would actually be affected if we made it, but I don't know that, but we might have to put it in the 5.0.0 queue or something, but we can schedule 5.0.0 to come pretty soon after 4.0.0 (instead of the normal minor releases Lwt has typically done).

hcarty:
I agree that raising `Invalid_argument` seems like a good approach here.  It also has the benefit of being something that can be applied across all of `pick`/`choose`/`nchoose`/`nchoose_split`.

A busy loop case isn't nearly as bad as a permanent hang in the `nchoose`/`nchoose_split` case since "why does my program make external requests/burn lots of CPU doing nothing?" is, as you said, a lot easier to debug than "my program just isn't doing anything."  But raising a documented exception certainly seems even better than either of those.

hcarty:
@aantron I propose closing this PR and either continuing the discussion in #551 or close this and #551 in favor of a new PR/issue.

aantron:
@hcarty that sounds good. If you'd like to make the PR, it would be welcome, otherwise I can do it. However, we will have to leave it unmerged for several months while we go through the usual warnings/request for feedback about a breaking change, so it would be merged in late spring or so, unless we can come up with some way of explaining to ourselves that it's okay to just do it :)

aantron:
And I suppose we should `raise` the exception rather than reject (`Lwt.fail`) the result promise. Of course, when this is used in a bunch of Lwt code, that exception is likely to be captured and used to reject some other promise, but that's an accident.

aantron:
@hcarty, as we discussed, I'm closing this in favor of #557. Let me know in case of any objections.

talex5:
> Returning a forever-pending promise, as now, seems pretty useless, and difficult to debug, as you pointed out. I doubt anyone actually wants that behavior in practice.

FWIW, one of my services just crashed because of this new behaviour: https://github.com/ocurrent/build-scheduler/issues/9

It's a server, so I want it to run forever, unless some reason to exit occurs. If there are no reasons to quit, then I want it to stay running.

raphael-proust:
```
let choose_or_never = function
  | [] -> fst @@ Lwt.task ()
  | _ :: _ as l -> Lwt.choose l
```

Of possible interest to you: in the Tezos project, we use a wrapper around `Lwt_main.run` to handle clean shutdown and signals. In one case, we used it around `fst @@ Lwt.task ()` and it only exits when signaled or when some part of the program calls an exit function. Interface here: https://gitlab.com/nomadic-labs/tezos/-/blob/raphael-p@improve-lwt-exit/src/lib_stdlib_unix/lwt_exit.mli

We will probably release it in a month or so.

PR #553, opened 2018-03-02
Replace notify=>signal in Lwt_condition documentation

iitalics:
The documentation refers to a function `notify` which does not exist, the actual function is called `signal`.


aantron:
Thanks!

Trivia: I looked in the history. It looks like `notify` was part of an old module `Lwt_monitor` that `Lwt_condition` replaced.

PR #555, opened 2018-03-04
Change install_filter.ml to simplify cross-compilation

whitequark:
This parallels the provision of package name to jbuilder.

Right now I have to carry a patch.

### DIFF
--- a/lwt.opam
+++ b/lwt.opam
@@ -19,7 +19,7 @@ build: [
   [ "ocaml" "src/util/configure.ml" "-use-libev" "%{conf-libev:installed}%"
                                     "-use-camlp4" "%{camlp4:installed}%" ]
   [ "jbuilder" "build" "-p" name "-j" jobs ]
-  [ "ocaml" "src/util/install_filter.ml" ]
+  [ "ocaml" "src/util/install_filter.ml" name ]
 ]
 build-test: [ [ "jbuilder" "runtest" "-p" name ] ]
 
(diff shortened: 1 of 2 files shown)


whitequark:
Nevermind, it's not as simple as that.

aantron:
@whitequark It may help that we will remove `install_filter.ml` as part of the Lwt 4.0.0 release, which should be in April. Hopefully, that will ease maintenance for you (and us).

whitequark:
Not really; since you're splitting lwt into so many opam packages, it's five times the pain for me.

aantron:
Well, this weird mechanism will disappear completely, so that should be helpful.

I looked in `opam-cross-windows` and `opam-cross-android`, and found this patch so far: https://github.com/ocaml-cross/opam-cross-windows/commit/61617990079cb0d55d7e75ebc1373cafaf57c3e1#diff-ee6402e863d36c85601b629aa33cbe1b. I'm not sure how/if this should be upstreamed into Lwt, but I'd be happy to make that patch unnecessary by doing so. I was going to release 3.2.2 very soon, so if you would like to get patches into Lwt itself, it may be an especially convenient time.

I'm assuming you have further patches you are working on.

whitequark:
> I'm not sure how/if this should be upstreamed into Lwt, but I'd be happy to make that patch unnecessary by doing so.

This shouldn't be upstreamed into Lwt. Because of the way we run the cross-compiler (run the bytecode executable using the host ocamlrun), our Windows-targeting ocamlc reports `os_type: Unix` in `ocamlc -config`. You went the extra mile to do everything correctly but sadly that doesn't help, at least not until this is fixed.

Issue #557, opened 2018-03-06
Make Lwt.pick [] raise Invalid_argument


aantron:
This issue is about adding an `if` to [`Lwt.pick`][pick], [`Lwt.choose`][choose], [`Lwt.npick`][npick], [`Lwt.nchoose`][nchoose], and [`Lwt.nchoose_split`][nchoose_split], so that they raise `Invalid_argument` when given the empty list.

[pick]: https://github.com/ocsigen/lwt/blob/020600c2304c3830a6cf71e143c2c3eaeb77155b/src/core/lwt.ml#L2678-L2700
[choose]: https://github.com/ocsigen/lwt/blob/020600c2304c3830a6cf71e143c2c3eaeb77155b/src/core/lwt.ml#L2656-L2676
[npick]: https://github.com/ocsigen/lwt/blob/020600c2304c3830a6cf71e143c2c3eaeb77155b/src/core/lwt.ml#L2791-L2845
[nchoose]: https://github.com/ocsigen/lwt/blob/020600c2304c3830a6cf71e143c2c3eaeb77155b/src/core/lwt.ml#L2732-L2787
[nchoose_split]: https://github.com/ocsigen/lwt/blob/020600c2304c3830a6cf71e143c2c3eaeb77155b/src/core/lwt.ml#L2850-L2925

Right now, each one of them looks like

```ocaml
let pick ps =
  the_code
```

and we need to change that to

```ocaml
let pick ps =
  if ps = [] then
    raise (Pervasives.Invalid_argument
      "Lwt.pick [] would return a promise that is pending forever")
  else
    the_code
```

<br/>

### Background

This was suggested by @hcarty in #551 and #552, and we had a discussion there. You're welcome to add more points to the discussion :) Perhaps we have missed something.

<br/>

### Testing

After making the change, run `make test`. Some of the tests will, naturally, fail ([[1]][pick-test] [[2]][choose-test] [[3]][npick-test] [[4]][nchoose-test] [[5]][nchoose_split-test]). To fix them, they just need to be changed to expect the exception. For example:

```ocaml
test "empty" begin fun () ->
  try
    ignore (Lwt.pick []);
    Lwt.return false
  with Pervasives.Invalid_argument "your chosen text" ->
    Lwt.return true
end;
```

[pick-test]: https://github.com/ocsigen/lwt/blob/020600c2304c3830a6cf71e143c2c3eaeb77155b/test/core/test_lwt.ml#L2516-L2519
[...]

PR #562, opened 2018-03-14
[5.0.0] Lwt.pick, etc raise Invalid_argument on empty list

timreinke:
As described in #557, this PR adjusts the behavior of `Lwt.pick`, `Lwt.choose`, `Lwt.npick`, `Lwt.nchoose`, and `Lwt.nchoose_split` raise `Invalid_argument` when passed an invalid list.  The test and docs are updated to reflect the new behavior.

### DIFF
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -2654,6 +2654,9 @@ struct
   let prng = lazy (Random.State.make [||])
 
   let choose ps =
+    if ps = [] then
+      invalid_arg
+        "Lwt.choose [] would return a promise that is pending forever";
     match count_resolved_promises_in ps with
     | 0 ->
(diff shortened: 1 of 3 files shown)


aantron:
Nice, thank you!

This resolves #557, so I'll close that issue. As described in #557, the PR will have to remain open (for at least three months), but I/others working on Lwt will merge it at the right time :)

Optional: it seems the email address in the commit metadata is not one of the ones associated with your GitHub account, so that's why GitHub is unable to link the commits to your user. Perhaps you're already aware of this, but if not, you may want to amend these commits to change the email address, or resolve this otherwise.

cc @hcarty

hcarty:
Thank you @timreinke!

timreinke:
You're very welcome, thanks for letting me take a crack at it.

Thanks for pointing out the email issue @aantron - whoops. I went ahead and fixed it.

PR #567, opened 2018-03-16
Lwt_io.file_length: raise EISDIR when used on directory

### LINKED ISSUE #563: Lwt_io.file_length of a directory returns 0x7FFFFFFFFFFFFFFF
Not sure it is a bug at all, but I'd have expected an exception (Unix_error EISDIR or Failure) from that user friendly helper module.
What do you think

aantron:
See #563. I don't think `Lwt_io.file_length` can return a meaningful result on a directory, even when that result is not max_int.

Two concerns:

- Maybe we should also fail on other kinds of [non-regular files](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html#TYPEfile_kind). However, I am not sure which [error code](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html#TYPEerror) to use and which file kinds to fail for. We should probably fix this now, when it is a bugfix, because a second time fixing it could mean a breaking change.
- I implemented the fix by calling `stat` before trying to get the file length, which adds one round trip to the Lwt worker thread pool (i.e., it is slow). If this becomes a performance bottleneck, we can speed up the implementation by using a custom job that does both `stat` and gets the file length in one call out to the thread pool. We can also probably use `Lwt_preemptive` for that, as it is getting merged into `Lwt_unix` very soon.

cc @rixed

Resolves #563.

### DIFF
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1508,7 +1508,12 @@ let with_temp_file ?buffer ?flags ?perm ?temp_dir ?prefix f =
       close chan >>= fun () ->
       Lwt_unix.unlink fname)
 
-let file_length filename = with_file ~mode:input filename length
+let file_length filename =
+  Lwt_unix.stat filename >>= fun stat ->
+  if stat.Unix.st_kind = Unix.S_DIR then
+    Lwt.fail (Unix.(Unix_error (EISDIR, "file_length", filename)))
+  else
+    with_file ~mode:input filename length
 
 let close_socket fd =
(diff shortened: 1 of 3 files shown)


rixed:
On non seekable files returning the seek error would make sense. What other file types are you worried about ?

aantron:
Ok, I think we will have `file_length` fail with `EISDIR` for directories, and let the underlying system calls fail with whatever they happen to fail with on each system in all other cases (at least for now). If anyone as better ideas, please let me know.

PR #577, opened 2018-04-15
Add Lwt.return to Lwt_pool example

bobbypriambodo:
An attempt to address [this comment on previous PR](https://github.com/ocsigen/lwt/pull/575#issuecomment-381374646). Is this good enough?


aantron:
Yes, thanks again :)

PR #579, opened 2018-04-19
Fix link without versions numbers with symlink

CelieArnaud:


aantron:
@CelieArnaud Is this the long-term fix for https://github.com/ocsigen/ocsigen.org/issues/18? It seems like it would involve maintenance on each release.

CelieArnaud:
It's only a quick fix, I'm working on a better way to do this automatically.

aantron:
Ok, thank you :)

PR #581, opened 2018-04-20
Use max_concurrency rather than max_threads

### LINKED ISSUE #578: Lwt_stream.iter_n ?max_threads should be ?max_concurrency
Or some other non-"threads" name.

hcarty:
Lwt.iter_n's optional max_threads argument incorrectly used the old
terminology.  This new argument name (max_concurrency) hopefully makes
the intent more clear for users.

Fixes #578

### DIFF
--- a/src/core/lwt_stream.ml
+++ b/src/core/lwt_stream.ml
@@ -910,12 +910,13 @@ let rec iter_p_rec node f s =
 
 let iter_p f s = iter_p_rec s.node f s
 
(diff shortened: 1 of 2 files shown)


hcarty:
This also adds a bit more prose to the `Lwt_stream.iter_n` doc comment.  Feedback on that is welcome.

hcarty:
Thank you too!

PR #582, opened 2018-04-24
Update readme to MIT

tychota:
Done from my phone at airport but should be simple.


aantron:
Thanks, I forgot the `README`.

Looking at the preview https://github.com/tychota/lwt/tree/patch-1#lwt------, the badge isn't rendering correctly, so I'll look into it. But definitely thanks for catching this :)

aantron:
I ended up [simplifying](https://github.com/ocsigen/lwt/commit/364d1474838e2b923b170971aef8cc2d1ae79b13) the license so that GitHub could detect it and display it. This made the badge redundant, and we also didn't need the README section "License" anymore, because the license is now visible and straightforward.

Thanks again :)

PR #585, opened 2018-04-27
Lwt_pool: some additional functionality

jrochel:


aantron:
Thanks, @bobbypriambodo. @jrochel, when you have a chance, could you also rebase this over `master`? It seems to be based on Lwt 3.3.0. We can also do this on the Lwt end, if it is troublesome.

bobbypriambodo:
Thank you for the reply. I'll put some questions here related to the implementations:

I'm a bit uncomfortable with the new `add` function. It works by giving an existing element to it; where do you get the element from? Is it by calling the creation function directly? I'm also concerned about this bit:

```
This function does perform any verification w.r.t. [p.max]. The maximum
number of resources might therefore be exceeded and more than [p.max]
elements will be available to the user.
```

(I assume you meant "does _not_ perform any verification" there.)

I can't help but think that it can lead to confusing things as `max` is no longer the "actual" maximum number of pool elements.

The closest thing to this that I could find is from Apache's [GenericObjectPool#addObject()](https://commons.apache.org/proper/commons-pool/apidocs/org/apache/commons/pool2/impl/GenericObjectPool.html#addObject()) for Java, but it's delegating the actual element creation to the given function, and do a no-op if the max number of elements is already reached. I believe it gives a more predictable behavior.

For the second feature, the `Resource_invalid` exception, we already have a mechanism to mark an element to be disposed of via the `check` function. Does it not fit your use case?

The problem I see from this implementation is that we "leak" the detail that we're using a pool to the function (by raising `Lwt_pool.Resource_invalid` exception). I personally think that the underlying function given to `use` should not be aware of whether they are receiving an element from a pool or by other way (e.g. what if hypothetically we want to utilize a different pooling library, or no pools at all?).

Would be happy to hear your thoughts. Thanks!

jrochel:
> I'm a bit uncomfortable with the new add function. It works by giving an existing element to it; where do you get the element from?

From the outside world. My use case is a cluster of physical/virtual web servers that are launched when the load on the cluster surpasses a given threshold. So I really don't see a sensible way to make this scenario fit the present approach of on-demand creation of resources.

> I'm also concerned about this bit: This function does [not] perform any verification w.r.t. [p.max]. The maximum number of resources might therefore be exceeded and more than [p.max] elements will be available to the user.

It is very understandable that you're concerned. I'm not sure what exactly to do about this yet, but I won't leave it like this. The problem is once more that the scenario doesn't fit the on-demand creation approach. In fact I don't want `Lwt_pool` to ever create resources for me. To achieve this, I create the pool by `let create () = failwith "..." in Lwt_pool.create 0 create`. Because of the maximum number of resources is`0`, `create` isn't ever called. I then use `add` to add resources manually.

> I can't help but think that it can lead to confusing things as `max` is no longer the "actual" maximum number of pool elements.

I agree. I'm not enthused about my current solution either. I'm hoping to find a better approach soon.

> For the second feature, the Resource_invalid exception, we already have a mechanism to mark an element to be disposed of via the check function. Does it not fit your use case?

It doesn't. At least it would be too inefficient as it requires an explicit ping to the resource (DB server/connection) before each use of the resource just to know
whether it is usable. Instead I'd like to use exceptions originating from the actual queries to eliminate connections/servers from the pool.

> The problem I see from this implementation is that we "leak" the detail that we're using a pool to the function (by raising `Lwt_pool.Resource_invalid` I personally think that the underlying function given to use should not be aware of whether they are receiving an element from a pool or by other way (e.g. what if hypothetically we want to utilize a different
> pooling library, or no pools at all?).

Actually I only use a thin layer to transform certain kinds of exceptions (as connection failures) to `Lwt_pool.Resource_invalid`, so there is no need to use `Lwt_pool.Resource_invalid` in the depths of the supplied function. So I'd say that this is the least problematic bit of this PR.

jrochel:
Have a look at [server_pool.ml] (only added for illustration) if you want to get an idea how I'm using `Lwt_pool`.

jrochel:
From my point-of-view this is now a mergable PR. Needless to say I am still open to suggestions for improvement.

aantron:
@jrochel, looking at this PR in detail, I get these impressions:

1. The use case is very specific to an application, and `Lwt_pool`'s existing design isn't fit for it.
2. To make `Lwt_pool` fit, this PR asks for some surprising/unnatural usage (which may be natural for a different module, but not for `Lwt_pool`).

In this case, I would suggest taking `Lwt_pool` out of Lwt, renaming it, and modifying it inside the application. Alternatively, an `Lwt_pool`-like module that *does* suit your application can be maintained as a separate library. I think `Lwt_pool` is meant to be a simple module for relatively simple situations.

What do you think about that?

aantron:
In case you choose to maintain a pool suitable for your need as a separate library, we would be happy to link to it from the `Lwt_pool` docs, and/or include it in the related libraries list in the README.

jrochel:
In the meantime I have used and tested the above functionality as described. It seems to work well, so now I'd like to go about finding a version of this PR that everybody is fine with. To recapitulate, there are the following functionalities that I have implemented:
- `Resource_invalid`: an exception that allows (during resource creation and resource usage) for signalling that a resource is no longer valid. This is an alternative to (or complements) a `validate` function. In my use case it is more efficient since the alternative would be to ping a server before each DB query.
- attempts: `use` has now two optional parameters `creation_attempts` and `usage_attempts` (both `1` by default) which allow you to specify how often `Lwt_pool` will reattempt resource creation/acquisition when a `Resource_invalid` exception is caught.
- `add`: the possibility to add already existing resources (might raise exception `Resource_limit_exceeded`)
- `omit_max_check`: an optional parameter of `add` that allows for existing resources to be added even if this would exceed the defined resource limit.

So given that @aantron dares to call my(!) PR "unnatural" and given his insolent refusal to merge it, I'd like to defend/discuss the above functionality some more and see whether we can't arrive at a mergable version. I have tried to rank the above points by potential controversialness, so I'd say that:
- `Resource_invalid` is a very useful feature and that any complexity added is more than justified.
- `creation_attempts` and `usage_attempts` are also quite useful and add little complexity.
- `add` is useful for my specific case and might be a little bit (but not too) surprising to a potential user. (raises `Resource_limit_exceeded`)
- `omit_max_check` is indeed "unnatural" and a potential user might ask itself "what the hell?"

For my use case `add` and `Resource_invalid` are critical functionalities, so if people deem it unfit for `Lwt_pool` I'd have to create and maintain a fork (which I'd hate to do). I think we could avoid the `omit_max_check` if there we could change the pool size dynamically. So that's my proposition: I'd remove `omit_max_check` but add a function `set_max` or `set_max_size` to change the size of an already established pool. This is by the way something that has been implemented as a hack in `Os_core_db` (from ocsigen-start), so if I added such a function `Os_core_db` would become a bit cleaner.

Comments? Approval? More insolence?

aantron:
@jrochel, I'd still be in favor of keeping `Lwt_pool` as it is, and maintaining this separately. There is no reason why the separate pool implementation should be handled as a "fork." You can simply copy the code in `Lwt_pool` now to a new repo, make whatever changes you wish, and not look back. We will still be happy to link to it.

One of the main reasons for this is that you are the only user of the new functionality, and we don't foresee any others. Only you are in a position to maintain it, because nobody else that routinely works on Lwt is accumulating experience with your use case. It seems most efficient for both you and the rest of Lwt to maintain this functionality separately.

For comparison, we recently factored out several modules out of Lwt. We were already considering factoring out `Lwt_pool` itself before this PR. We think it's better for there to be several different Lwt-friendly resource pools in the community, each serving specific needs, and maintained by their interested users, rather than one `Lwt_pool` in a central repo, that tries to satisfy all, and is fully understood by probably nobody, thus being unmaintainable.

jrochel:
*admitting defeat*

I'll create a package then soon.

aantron:
Thank you, and sorry for being such an obstacle. Please do give a link to the package (if public) in this PR, we'll refer to it in `lwt_pool.mli` and the README.

jrochel:
No worries, I'll get over it. Someday. Thanks anyway for the feedback. I'll post a link here as soon as the package is available.

PR #586, opened 2018-04-29
Lwt_io.establish_server (TCP servers): expose client socket to connection-handling callback

aantron:
This is a follow-on to #323/#346 (cc @rgrinberg). It adds a function like:

```ocaml
val establish_server_with_client_socket :
  Unix.sockaddr ->
  (Unix.sockaddr -> Lwt_unix.file_descr -> (input_channel * output_channel) -> unit Lwt.t) ->
    server Lwt.t
```

...whereas existing versions of this function didn't pass the `Lwt_unix.file_descr` to the client-handling function, and only passed the `sockaddr` and the two channels.

I'm still not sure if this is the right API, but starting with this.

The PR also includes a pretty thorough cleanup of the existing code, so it should be easier to keep working on.

### DIFF
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1659,24 +1659,6 @@ let establish_server_generic
 
   server, server_has_started
 
-(* Old, deprecated version of [establish_server]. This function has to persist
-   for a while, in some form, until it is no longer exposed as
-   [Lwt_io.Versioned.establish_server_1]. *)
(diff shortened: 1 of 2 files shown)


aantron:
This was motivated by inhabitedtype/httpaf#54. inhabitedtype/httpaf#53 also anticipated the type signature of `establish_server_with_client_socket`, so merging.

PR #597, opened 2018-06-08
Ppx: refuse match cases with only exception cases

### LINKED ISSUE #596: match%lwt accepts all cases being exception cases
According to https://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec264:
> It is an error if all cases are exception cases in a given pattern matching.

Checking in toplevel, the following does not compile, as expected:
```
# match () with exception exn -> ();;
Error: None of the patterns in this 'match' expression match values.
```

But `match%lwt` gives a run-time error instead:
```
# Lwt_main.run @@ match%lwt Lwt.return_unit with exception exn -> Lwt.return_unit;;
Exception: Match_failure ("//toplevel//", 1, 16).
```

raphael-proust:
Fixes #596 

I guess this technically counts as a breaking change just in case someone was using the behaviour (even though it was incompatible with ocaml's).

### DIFF
--- a/src/ppx/ppx_lwt.ml
+++ b/src/ppx/ppx_lwt.ml
@@ -186,6 +186,12 @@ let lwt_expression mapper exp attributes =
         | {pc_lhs = [%pat? exception [%p? _]]; _} -> true
         | _ -> false)
     in
+    if cases = [] then
+        raise (Location.Error (
+          Location.errorf
+            ~loc:exp.pexp_loc
+            "match%%lwt must contain at least one non-exception pattern."
+        ));
     let exns =
       exns |> List.map (
         function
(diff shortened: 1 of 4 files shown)


aantron:
Thanks @raphael-proust! I think we can skip the full breaking-change treatment for this PR :)

For testing, we could add the ability to skip expect tests conditionally to the expect test framework. I'll push something into this PR tomorrow.

aantron:
Looking at the output, it may also be good enough to remove "Error: " prefixes from lines.

aantron:
Thanks @raphael-proust, and @cyberhuman for finding and reporting the issue. For the expect tests, I decided to restrict them to OCaml >= 4.04. We were already restricted to >= 4.03 before that.

PR #602, opened 2018-07-04
Ppx: fix loc info of sequence exp

kandu:
This fixes a tiny problem that merlin didn't show type/range info of sequence exp.


kandu:
the second commit makes the compilation error(pattern matching produced by ppx) more precise.
cc @hcarty

aantron:
Thanks, and thanks for catching the issue addressed by the second commit!

FTR, I used this code to trigger the error:

```ocaml
let () =
  let p = Lwt.return 1 ;%lwt Lwt.return () in
  ignore p
```

without the second commit, it triggers

```
File "foo.ml", line 2, characters 10-22:
Error: This pattern matches values of type unit
       but a pattern was expected which matches values of type int
```

with the commit, the error location is:

```
File "foo.ml", line 2, characters 25-28:
```

hcarty:
Thank you @kandu!

PR #605, opened 2018-08-20
Lwt_sequence: remove checks for active nodes in iter-like functions

### LINKED ISSUE #604: Lwt_sequence implementation question wrt active nodes
This is not an issue per-se, but more a question I have regarding an implementation detail in `Lwt_sequence`.

In iteration functions (`iter_*`, `fold_*`), the code checks if the nodes are active before calling the function on them (e.g. [here](https://github.com/ocsigen/lwt/blob/master/src/core/lwt_sequence.ml#L128)).
However, unless I'm mistaking, it seems to me that nodes that are linked to the sentinel node are **necessarily** active (nodes can only have `active = false` after `remove` has been called on them, but then they have been unlinked). Is my understanding correct, or am I missing something?

If this reasoning is correct the check could be removed, which might even make the code slightly faster.

Armael:
These checks can be removed because of the implicit invariant that only active
nodes can be reached from the sentinel node of the list. Indeed, a node's active
field is set to false when calling [remove] on it -- but then the node is
unlinked from the list, so it becomes unreachable by functions such as `iter_*`
or `fold_*`.

This is consistent with the previous removal of a similar check in the
implementation of [length] in #542

Closes #604

### DIFF
--- a/src/core/lwt_sequence.ml
+++ b/src/core/lwt_sequence.ml
@@ -125,7 +125,7 @@ let iter_l f seq =
   let rec loop curr =
     if curr != seq then begin
       let node = node_of_seq curr in
-      if node.node_active then f node.node_data;
+      f node.node_data;
       loop node.node_next
     end
   in
@@ -135,7 +135,7 @@ let iter_r f seq =
   let rec loop curr =
     if curr != seq then begin
       let node = node_of_seq curr in
-      if node.node_active then f node.node_data;
+      f node.node_data;
       loop node.node_prev
     end
   in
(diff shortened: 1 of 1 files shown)


Armael:
Oops wait, this is incorrect, as the tests show...

Armael:
Sorry for the noise! I didn't realize that removing arbitrary nodes was allowed in (all the) iteration functions.

aantron:
@Armael, I'd be happy to merge a PR adding a comment to `active` explaining this, if you'd like to make one :) It seems this has been a source of confusion for several people by now, myself included.

PR #606, opened 2018-08-20
fixed race condition in Lwt_react.E.limit

AvniFatehpuria:
Now, the function that runs on success ensures that Lwt.is_sleeping is false before pushing the event.

### DIFF
--- a/src/react/lwt_react.ml
+++ b/src/react/lwt_react.ml
@@ -52,7 +52,8 @@ module E = struct
              | None ->
                let cell = ref x in
                delayed := Some cell;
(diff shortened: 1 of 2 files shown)


aantron:
@AvniFatehpuria Thanks! We should probably rewrite this function completely eventually, because the code is pretty difficult to follow, if one wishes to think through all the corner cases.

There might also be a version of this bug in [`Lwt_react.S.limit`](https://github.com/ocsigen/lwt/blob/master/src/react/lwt_react.ml#L265), the signals counterpart of `Lwt_react.E.limit`. If you're interested, I suggest trying to come up with a test case to reveal the bug, and fixing it too, if it is there.

PR #609, opened 2018-08-23
Forbid nesting calls to Lwt_main.run

### LINKED ISSUE #607: Lwt_main.run should fail with prejudice if already running
[Original issue title: *Lwt_mvar.take fails to be woken up by put*]

Using lwt 3.3.0 on ocaml 4.06.1 from opam, I have one thread that does:

```
        Lwt_io.printf "Waiting for DB message on %d...\n%!"
          (2 * Obj.magic db_mailbox)
        >>= fun () ->
        Lwt_mvar.take db_mailbox
        >>= fun msg ->
        Printf.printf "RECEIVED DB MESSAGE\n%!";
```
while the other thread does:
```
  Lwt_io.printf "Sending DB message to %d: %s\n%!"
    (2 * Obj.magic db_mailbox)
    (string_of_yojson (yojson_of_db_request r))
  >>= fun () ->
  Lwt_mvar.put db_mailbox r
```
and a third watching thread in a loop does:
```
         Lwt_io.printf "The mailbox at %d is %sempty\n%!"
           (2 * Obj.magic db_mailbox)
           (if Lwt_mvar.is_empty db_mailbox then "" else "not ")
         >>= fun () ->
         Lwt_unix.sleep 1.0
```
The first (server) thread hangs after it prints its message.
The second (client) thread prints and queues three messages.
The third (watcher) thread tells me that the mvar is not empty.
The first thread looks like it's never woken up.

The whole thing works when I run my unit tests, that use the very same setup. Then when I try to compile the actual executable server, it hangs. I'm a bit at a loss regarding what I may be doing wrong, and would appreciate guidance in debugging, whether the bug is mine (likely) or yours (unlikely, but possible?).

aantron:
cc @fare

Resolves #607.

### DIFF
--- a/src/unix/lwt_main.ml
+++ b/src/unix/lwt_main.ml
@@ -60,29 +60,27 @@ let run t =
   end;
 
   let rec run_loop () =
-
-  (* Wakeup paused threads now. *)
-  Lwt.wakeup_paused ();
-  match Lwt.poll t with
-  | Some x ->
-    x
-  | None ->
-    (* Call enter hooks. *)
-    Lwt_sequence.iter_l (fun f -> f ()) enter_iter_hooks;
-    (* Do the main loop call. *)
-    Lwt_engine.iter (Lwt.paused_count () = 0 && Lwt_sequence.is_empty yielded);
-    (* Wakeup paused threads again. *)
+    (* Wakeup paused threads now. *)
     Lwt.wakeup_paused ();
-    (* Wakeup yielded threads now. *)
-    if not (Lwt_sequence.is_empty yielded) then begin
-      let tmp = Lwt_sequence.create () in
-      Lwt_sequence.transfer_r yielded tmp;
-      Lwt_sequence.iter_l (fun wakener -> Lwt.wakeup wakener ()) tmp
-    end;
-    (* Call leave hooks. *)
-    Lwt_sequence.iter_l (fun f -> f ()) leave_iter_hooks;
-    run_loop ()
-
+    match Lwt.poll t with
+    | Some x ->
+      x
+    | None ->
+      (* Call enter hooks. *)
+      Lwt_sequence.iter_l (fun f -> f ()) enter_iter_hooks;
+      (* Do the main loop call. *)
+      Lwt_engine.iter (Lwt.paused_count () = 0 && Lwt_sequence.is_empty yielded);
+      (* Wakeup paused threads again. *)
+      Lwt.wakeup_paused ();
+      (* Wakeup yielded threads now. *)
+      if not (Lwt_sequence.is_empty yielded) then begin
+        let tmp = Lwt_sequence.create () in
+        Lwt_sequence.transfer_r yielded tmp;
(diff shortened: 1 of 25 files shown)


src/unix/lwt_main.ml
```diff
@@ -20,28 +20,84 @@ let yielded = Lwt_sequence.create ()
...
+        backtrace_string)
+    | `From_somewhere ->
+      Some ("Nested calls to Lwt_main.run are not allowed")
+    | `No ->
+      let called_from =
+        if Printexc.backtrace_status () then
+          let backtrace =
+            try raise Exit
+            with Exit -> Printexc.get_backtrace ()
+          in
```
aantron: I decided to try recording stack traces of where `Lwt_main.run` was called, rather than add an optional argument.

Drup:
I don't really object to it, but have you checked on opam that this doesn't break anything ?

Drup:
Actually, opam will probably not tell you that anyway .... Anyway, I think it's a breaking change in general, even if it has been strongly verboten for years.

aantron:
@fare If at all easy to do, could you check that this actually catches the error in the specific setup you had? I don't know enough about dynamic loading in OCaml and/or ocsigenserver to know whether the new internal [`run_already_called`](https://github.com/ocsigen/lwt/pull/609/commits/9f832c7cb26327beb52a66673aac863097fdc9d1#diff-e0a5cf6b13a0e51544ecdb2f955843d8R23) flag will be shared or not.

aantron:
@Drup I just searched the revdeps for it, the number of occurrences seems to large to analyze by hand. Many are in tests and examples, but there are plenty in the main source code of packages. Let's see if this helps catch the problem @fare was having. If so, we'll tag it breaking and put it in in 5.0.0 with all the proper announcements, etc.

fare:
Thanks a lot for this patch!

Importing a new lwt requires some setup, because I need to pin not just lwt but several other packages that depend on it. I'll try tomorrow when I have more time.

raphael-proust:
I don't think it's that weird: if you don't call it, then it's probably called by `ocsigenserver` after initialisation and so it works; if you do call it, then it probably follows the sequential execution described in #607: “If you were calling `Lwt_main.run` under a top-level `let _ =`, it seems to me the other `Lwt_main.run` must have already returned, so the two calls shouldn't be interfering with each other.”

aantron:
@fare Can you say which versions of ocsigenserver and Lwt you were using before, and which you are using now?

fare:
Before: I was using ocsigenserver 2.9 and lwt 3.3.0 from OPAM.

After: I am using lwt from source (your branch) and still ocsigenserver 2.9 it seems. At some point I was also running from source because it was complaining about lwt being too recent, but it looks like this doesn't happen with your branch of lwt.

My understanding is that ocsigenserver was loading my code from within the lwt loop, but then again, I don't really know.

fare:
Actually, I don't understand OPAM, and I was probably using the wrong lwt and ocsigenserver. Now I made sure I was using the right one (pinning lwt, cohttp, ocsigenserver, eliom), and somehow, I can't reproduce the issue in this setting, and I don't get an error message either.

PR #612, opened 2018-08-24
Link to Mwt, alternative to Lwt_preemptive

aantron:
...created by @hcarty, thanks @hcarty!

Docs are here: https://hcarty.github.io/mwt/mwt/Mwt/index.html.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,5 +191,6 @@ serialization combinators
 logging
 - [lwt-parallel](https://github.com/ivg/parallel) —
 distributed computing
+- [mwt](https://github.com/hcarty/mwt) — preemptive (system) thread pools
 - [opium](https://github.com/rgrinberg/opium) —
 web framework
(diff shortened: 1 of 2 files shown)


Drup:
Is it supposed to replace Lwt_preemptive? What's the difference?

Come on, we need details to hype us up. :D

hcarty:
I'll get a proper discuss post together in the next few days!

In short, Mwt provides user-managed preemptive thread pools with per-thread state. So a program can have multiple thread pools, with each pool dedicated to its own purpose. So very much like Lwt_preemptive but with explicitly defined pools of threads. 

The per-thread state part can be useful when working with some C libraries which need a stateful entity to be managed (created, used and disposed of) from a single thread.

An an example, one pool could be used to distribute lacaml calls across cores while another could manage database handles which do not have a cooperative interface available.

Issue #371, opened 2017-05-22
Delete simple-top


aantron:
`simple-top` was deprecated in favor of [utop](https://github.com/diml/utop) in or before bb73e9834ceacb8e6c2aed7c8f19fdd81bc404af (summer 2011). There are no users in OPAM. A GitHub search for it returns only @bobzhang's [ocaml-book](https://github.com/bobzhang/ocaml-book) and @kkirstein's [proglang-playground](https://github.com/kkirstein/proglang-playground).

I think we can announce the removal of simple-top in the next minor release, and delete it 4.0.0.

Related #313.

kkirstein:
That's fine with me, just kick it out as planned...

PR #615, opened 2018-08-30
discover.ml: do not pipe ocamlc through ocamlfind

avsm:
This seems to be the only dependency on ocamlfind left in the codebase, and is (I think) unused. With this patch, I can build Lwt without having ocamlfind installed (which is useful in a Dune monorepo)

### DIFF
--- a/src/unix/config/discover.ml
+++ b/src/unix/config/discover.ml
@@ -280,7 +280,7 @@ CAMLprim value lwt_test()
    | Compilation                                                     |
    +-----------------------------------------------------------------+ *)
 
-let ocamlc = ref "ocamlfind ocamlc"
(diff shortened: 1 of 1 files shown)


avsm:
darn, this seems to break windows...

aantron:
I don't think we need that part of the configuration anymore, IIRC it was for `lwt.simple-top`, which has been deleted (#371). Looking into it.

aantron:
More detail, check was added in #313.

aantron:
#616 replaces using `ocamlfind` with another workaround and summarizes the issue `ocamlfind` was working around. I suggest to close this PR, and merge that one. Thanks for triggering this, it's been on the todo list for a while!

PR #616, opened 2018-08-30
Remove ocamlfind as a release dependency

aantron:
`ocamlfind` was used during the configure step of the Lwt release build. It is now only used by the test suite, so it is only a development dependency.

We were calling `ocamlfind ocamlc` instead of just `ocamlc`, because `ocamlfind` was working around a bug in the command line of `ocamlc`, where `.o` files compiled from `.c` files would be output in the current directory, instead of the same directory as the `.c` file. See https://github.com/ocaml/ocaml/commit/da56cf6dfdc13c09905c2e07f1d4849c8346eec8. This bug was fixed in OCaml 4.04, which explains why #615 failed on 4.03 and 4.02.

This PR changes `discover.ml` to detect and move the `.o` file if it is output to the current directory.

Alternative to #615. cc @avsm

### DIFF
--- a/lwt.opam
+++ b/lwt.opam
@@ -24,16 +24,10 @@ depends: [
   # Bisect_ppx is used only during development. This dependency should be
   # removed when preparing a release.
   "bisect_ppx" {>= "1.3.0"}
+  # ocamlfind is a development dependency of the test suite.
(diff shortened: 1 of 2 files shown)


avsm:
This works great over here -- thanks for the quick fix!

PR #618, opened 2018-08-31
rfc: disable libev build by default in the discovery scripts

avsm:
This simplifies the configure scripts to:

- disable libev build by default unless it is supplied on the configure line.  This does not affect most installations such as via opam (where the flag is explicitly specified depending on the presence of conf-libev) or OS packaging (where libev is typically a concrete dependency specified in the package data). However, this does make the build of Lwt much more portable by default when libev is not available.

- remove writing of setup.data, which seems to be vestigial from the oasis build days. Simplifies the discovery script.

- remove the auto-detection of libev on opam, which is not necessary since the opam package specifies it. Due to an unfortunate bug in opam 2.0.0 (ocaml/opam#3525) this is always true and so has different behaviour on opam1 vs 2.

The overall motivation for this PR is for embedding Lwt in the experimental [ocaml platform](https://github.com/avsm/platform) monorepo, which bootstraps on all supported OCaml platforms with just a C compiler.  It only needs libev for utop, and portability is significantly simpler if libev (and hence pkg-config) are not hard dependencies of this package for that purpose.

I'm opening this PR to discuss other options for out-of-the-box portability as well, in case I'm missing something far simpler!

### DIFF
--- a/src/unix/config/configure.ml
+++ b/src/unix/config/configure.ml
@@ -34,20 +34,6 @@ let main () =
 
   let config_file = "lwt_config" in
 
-  begin match !default_configuration with
-  | Some true ->
-    (* Check if we have the opam command and conf-libev is installed. If so,
-       behave as if -use-libev true was passed. *)
-    if not @@ Sys.file_exists config_file then
-      if Sys.command "opam --version > /dev/null" = 0 then
-        if Sys.command "opam list --installed conf-libev > /dev/null" = 0 then
-          use_libev := Some true
-        else
-          use_libev := Some false
-  | _ ->
-    ()
-  end;
-
   let f = open_out config_file in
   let print name var =
     match var with
--- a/src/unix/config/discover.ml
+++ b/src/unix/config/discover.ml
@@ -581,7 +581,7 @@ let () =
     (* set up the defaults as per the original _oasis file *)
     get android_target "android_target" false;
     get use_pthread "use_pthread" (!os_type <> "Win32");
-    get use_libev "use_libev" (!os_type <> "Win32" && !android_target = false);
+    get use_libev "use_libev" false;
     get libev_default "libev_default"
       (List.mem !system (* as per _oasis *)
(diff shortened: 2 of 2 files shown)


aantron:
@avsm:

> remove the auto-detection of libev on opam, which is not necessary since the opam package specifies it.

What does this refer to? The same commit that added auto-detection of `conf-libev` is the commit that removed the call to `configure.ml` from the `lwt.opam` file. The one in the repo now has:

https://github.com/ocsigen/lwt/blob/38bf913c12926c97f8207f5f8799d6f00b7a7dd9/lwt.opam#L18-L20

Removing that configure command was the purpose of adding the auto-detection of `conf-libev`.

aantron:
Before the auto-detection was added, we had

https://github.com/ocsigen/lwt/blob/88c745745e27681958cd48ccebf9cd7e179be7bb/lwt.opam#L18-L21

avsm:
Oops, I hadn't noticed that particular change.  I think we need a scheme where the configure phase can be run by package builds, but doesn't *have* to be (so it can be embedded easily in a bigger monorepo).

So if the opam package still explicitly runs the configure.ml as it used to and libev defaults to false when not specified, that would leave most users unchanged but the defaults more portable.  At least, I think it will...

aantron:
On the Lwt side, the consensus is that Lwt is much less useful without libev. We really want to compile with it if available on the target, and we want to make that the default in as many situations as possible.

avsm:
How about keeping the current detection, and just not terminating the build if libev isn't detected? We could print a warning in discover.ml and continue as if `-with-libev false` had been specified.

aantron:
Yeah, Lwt should use libev if available, not use it if not available, and only fail if a flag was specified but libev is not correspondingly present or absent. Would you like to do that in this PR, or do you prefer I fix it?

avsm:
That sounds like a good solution to me. If you do have time to fix it to your satisfaction, I'd appreciate it -- otherwise I can look at it next week!

Issue #539, opened 2018-01-15
Test the Unix binding


aantron:
This is a big issue for lots of PRs. Some parts of it could be quite easy, but on the whole I'd say it is of medium difficulty.

Some care should be taken to run the tests properly sandboxed, i.e. files created in temporary directories, etc. For that, the test framework in [`test/test.ml`](https://github.com/ocsigen/lwt/blob/4f0dabca5228b36a65ea7e49375a15d2070161fa/test/test.mli) will probably have to be extended with new helper functions. I'm not sure how to sandbox DNS and some of the other APIs. We should probably discuss that.

Besides correct behavior, we need really thorough tests for various error conditions. The typical system call has man pages on multiple systems, listing the various ways the system call can fail. We want to trigger as many of those as possible. See section *Help* below for links to man pages.

<br/>

## The work

To make this manageable, I would break the work into chunks, based on the current order of functions in the various `.mli` files of the Unix binding. These aren't listed necessarily in the order they should be tested in. We generally want simple and obvious things tested first, and more complex things that depend on them tested later.

There are a few tests already written, but I included all the functions in the category list, so make sure to check before choosing a category. I'll try to keep the check boxes up to date.

If you'd like to work on one of these items, please leave a comment, to avoid duplication of effort :)

- [ ] [Bigstrings](https://github.com/ocsigen/lwt/blob/4f0dabca5228b36a65ea7e49375a15d2070161fa/src/unix/lwt_bytes.mli#L22-L111). This is pure OCaml, and should be easy.
- [ ] [Sleeping and timeouts](https://github.com/ocsigen/lwt/blob/4f0dabca5228b36a65ea7e49375a15d2070161fa/src/unix/lwt_unix.cppo.mli#L130-L157), probably best tested with the help of [`Unix.gettimeofday`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html#VALgettimeofday).
- [ ] [`Lwt_throttle`](https://github.com/ocsigen/lwt/blob/4f0dabca5228b36a65ea7e49375a15d2070161fa/src/unix/lwt_throttle.mli). I guess this is in the Unix binding because it is aware of real time.
- [ ] [`Lwt_timeout`](https://github.com/ocsigen/lwt/blob/4f0dabca5228b36a65ea7e49375a15d2070161fa/src/unix/lwt_timeout.mli).
[...]

PR #619, opened 2018-09-08
Test unix bytes

cedlemo:
#539 

 Sorry for the delay,

I send this for a first review before I start the tests of the IO functions of Lwt_bytes.

### DIFF
--- a/test/unix/test_lwt_bytes.ml
+++ b/test/unix/test_lwt_bytes.ml
@@ -131,6 +131,19 @@ let suite = suite "lwt_bytes" [
       | _ -> Lwt.return false
     end;
 
+    test "blit source out of bound: upper limit" begin fun () ->
+      let str1 = "abc" in
+      let buf1 = Lwt_bytes.of_string str1 in
+      let str2 = "abcdef" in
+      let buf2 = Lwt_bytes.of_string str2 in
+      try
+        let() = Lwt_bytes.blit buf1 150 buf2 3 3 in
+        Lwt.return_false
+      with
+      | Invalid_argument _ -> Lwt.return true
+      | _ -> Lwt.return false
+    end;
+
     test "blit from bytes" begin fun () ->
       let bytes1 = Bytes.of_string "abc" in
       let str2 = "abcdef" in
--- a/test/unix/test_lwt_bytes.ml
+++ b/test/unix/test_lwt_bytes.ml
@@ -287,13 +287,78 @@ let suite = suite "lwt_bytes" [
       Lwt.return check
     end;
 
-    test "blit to bytes out of bounds" begin fun () ->
+    test "blit to bytes source lower limit out of bounds" begin fun () ->
       let str1 = "abc" in
       let buf1 = Lwt_bytes.of_string str1 in
       let str2 = "abcdef" in
       let bytes2 = Bytes.of_string str2 in
       try
-        let () = Lwt_bytes.blit_to_bytes buf1 150 bytes2 150 150 in
+        let () = Lwt_bytes.blit_to_bytes buf1 (-1) bytes2 3 3 in
+        Lwt.return_false
+      with
+      | Invalid_argument _ -> Lwt.return_true
+      | _ -> Lwt.return_false
+    end;
+
+    test "blit to bytes source upper limit out of bounds" begin fun () ->
+      let str1 = "abc" in
+      let buf1 = Lwt_bytes.of_string str1 in
+      let str2 = "abcdef" in
+      let bytes2 = Bytes.of_string str2 in
+      try
+        let () = Lwt_bytes.blit_to_bytes buf1 150 bytes2 3 3 in
+        Lwt.return_false
+      with
+      | Invalid_argument _ -> Lwt.return_true
+      | _ -> Lwt.return_false
+    end;
+
+    test "blit to bytes destination lower limit out of bounds" begin fun () ->
+      let str1 = "abc" in
+      let buf1 = Lwt_bytes.of_string str1 in
+      let str2 = "abcdef" in
+      let bytes2 = Bytes.of_string str2 in
+      try
+        let () = Lwt_bytes.blit_to_bytes buf1 0 bytes2 (-1) 3 in
+        Lwt.return_false
+      with
+      | Invalid_argument _ -> Lwt.return_true
+      | _ -> Lwt.return_false
+    end;
+
+    test "blit to bytes destination upper limit out of bounds" begin fun () ->
+      let str1 = "abc" in
+      let buf1 = Lwt_bytes.of_string str1 in
+      let str2 = "abcdef" in
+      let bytes2 = Bytes.of_string str2 in
+      try
+        let () = Lwt_bytes.blit_to_bytes buf1 0 bytes2 150 3 in
+        Lwt.return_false
+      with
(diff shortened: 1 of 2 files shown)


aantron:
Thanks, looking! It looks like `Bytes.equal` requires OCaml 4.03, but we support 4.02. I assume `=` (polymorphic equality) does the same thing, we can probably define a local `bytes_equal` that is `=` with the right constraints for use in the test suite.

test/unix/test_lwt_bytes.ml
```diff
@@ -0,0 +1,248 @@
+(* This file is part of Lwt, released under the MIT license. See LICENSE.md for
+   details, or visit https://github.com/ocsigen/lwt/blob/master/LICENSE.md. *)
+
+open Test
+
+let suite = suite "lwt_bytes" [
+    test "create" begin fun () ->
+      let len = 5 in
+      let buff = Lwt_bytes.create len in
+      let len' = Lwt_bytes.length buff in
```
aantron: Just a note, it's a bit more "authoritative" to test this using [`Bigarray.Array.dim`](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Bigarray.Array1.html#VALdim) instead of `Lwt_bytes.length`. Not suggesting to actually change the test case, not worth the time :)

test/unix/test_lwt_bytes.ml
```diff
@@ -0,0 +1,248 @@
...
+      let buff = Lwt_bytes.create 3 in
+      try
+        let () = Lwt_bytes.set buff 150 'a' in
+        Lwt.return true
+      with
+      | Invalid_argument _ -> Lwt.return true
+      | _ -> Lwt.return false
+    end;
+
+    test "unsafe_get/unsage_set" begin fun () ->
```
aantron: s/unsage/unsafe

test/unix/test_lwt_bytes.ml
```diff
@@ -0,0 +1,248 @@
...
+      Lwt.return check
+    end;
+
+    test "blit out of bound" begin fun () ->
+      let str1 = "abc" in
+      let buf1 = Lwt_bytes.of_string str1 in
+      let str2 = "abcdef" in
+      let buf2 = Lwt_bytes.of_string str2 in
+      try
+        let() = Lwt_bytes.blit buf1 150 buf2 150 150 in
```
aantron: I think fully thorough out-of-bounds access tests would check individually for both lower and upper bounds being exceeded, for both the source and destination, so four tests per function.

test/unix/test_lwt_bytes.ml
cedlemo: done

test/unix/test_lwt_bytes.ml
cedlemo: done

test/unix/test_lwt_bytes.ml
```diff
@@ -0,0 +1,526 @@
+(* This file is part of Lwt, released under the MIT license. See LICENSE.md for
+   details, or visit https://github.com/ocsigen/lwt/blob/master/LICENSE.md. *)
+
+open Test
+
+let bytes_equal b1 b2 = b1 = b2
```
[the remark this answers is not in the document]
cedlemo: So for the former I should I done something like:

```
let bytes_equal (b1:Bytes.t) (b2:Bytes.t) = b1 = b2
```
Sorry I did not understood why you wanted me to create the bytes_equal function, if I could use `=` directly.

test/unix/test_lwt_bytes.ml
```diff
@@ -0,0 +1,526 @@
...
+      let str1 = "abc" in
+      let buf1 = Lwt_bytes.of_string str1 in
+      let str2 = "abcdef" in
+      let bytes2 = Bytes.of_string str2 in
+      let () = Lwt_bytes.unsafe_blit_to_bytes buf1 0 bytes2 3 3 in
+      let check = "abcabc" = Bytes.to_string bytes2 in
+      Lwt.return check
+    end;
+
+    test "proxy" begin fun () ->
```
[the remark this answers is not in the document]
cedlemo: done.

cedlemo:
@aantron , do you think it would be cleaner to use the first number that generates an exception rather than a random number that is greater than the limit. ie -> should I change the random 150 value in the tests for out of bounds parameters?

test/unix/test_lwt_bytes.ml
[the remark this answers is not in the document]
aantron: I would just use `=` directly, but I thought you were using `Bytes.equal` deliberately instead, so I suggested to recreate it as `bytes_equal` for 4.02, with the definition you just suggested.

test/unix/test_lwt_bytes.ml
```diff
@@ -0,0 +1,497 @@
...
+      | _ -> Lwt.return false
+    end;
+
+    test "blit source out of bounds: upper limit" begin fun () ->
+      let str1 = "abc" in
+      let buf1 = Lwt_bytes.of_string str1 in
+      let str2 = "abcdef" in
+      let buf2 = Lwt_bytes.of_string str2 in
+      try
+        let() = Lwt_bytes.blit buf1 4 buf2 3 3 in
```
aantron: Sorry, I made a mistake and didn't see the source is only 3 bytes (not 6), so I'd pick 1 or 2 here as the starting offset: the offset of a byte that is actually within the source buffer, so that it is adding the length which triggers the upper limit check failure.

test/unix/test_lwt_bytes.ml
```diff
@@ -0,0 +1,497 @@
...
+      | _ -> Lwt.return false
+    end;
+
+    test "blit destination out of bounds: upper limit" begin fun () ->
+      let str1 = "abc" in
+      let buf1 = Lwt_bytes.of_string str1 in
+      let str2 = "abcdef" in
+      let buf2 = Lwt_bytes.of_string str2 in
+      try
+        let() = Lwt_bytes.blit buf1 0 buf2 150 3 in
```
aantron: Suggest 4 or 5 instead of 150, for the same reason as above.

aantron:
LGTM, so I will merge in a couple hours if there is no objection :) Thanks for your patience :)

test/unix/test_lwt_bytes.ml
```diff
@@ -37,7 +37,7 @@ let suite = suite "lwt_bytes" [
     test "get out of bounds : upper limit" begin fun () ->
       let buff = Lwt_bytes.create 3 in
       try
-        let _ = Lwt_bytes.get buff 150 in
+        let _ = Lwt_bytes.get buff 4 in
```
aantron: Actually, I think the first index out of bounds is 3, and the value of testing that one specifically is checking for `>=` instead of `>`. But this is just a nit.

cedlemo: done.

aantron:
Thanks again!

cedlemo:
You are welcome! , I will continue with the IO part. Are there some helpers in Lwt test suite like in OUnit with `OUnit2.bracket_tmpfile` that will help me to create temporary file or socket for the remaining tests ?

Furthermore the functions related to the memory seems hard to tests : 
* madvize
* page_size 
* mincore
* wait_mincore

any advices ?

PR #621, opened 2018-09-14
Test lwt bytes second part

cedlemo:
@aantron,

This part contains only two tests for now but a review could be helpfull:

- the files used to test read and write functions are named with the `.ml` extension so that they are copied in the `_build/default/test/unix` when the tests are run. I tried the `copy_files` stanza in the *test/unix/dune* file without success. Should I put the file in a dedicated directory in *src/test/unix* ?

- I had to generate a fd from the Unix module and transform it to an Lwt file descriptor. In the second test, I had to close the underlying Unix fd that I used to create the Lwt fd in order to flush everything before reading what has been written for the test. Should I close all the fd using the `Unix.close` at the end of the tests? I have the feeling that it is not really clean.


test/unix/test_lwt_bytes.ml
```diff
@@ -496,4 +497,32 @@ let suite = suite "lwt_bytes" [
...
+      let buf = Lwt_bytes.create 6 in
+      Lwt_bytes.read fd buf 0 6
+      >>= fun _n ->
+      let check = "abcdef" = Lwt_bytes.to_string buf in
+      Lwt.return check
+    end;
+
+    test "bytes write" begin fun () ->
+      let test_fd = "bytes_io_data_write.ml" in
+      let unix_fd = Unix.openfile test_fd [O_RDWR;O_TRUNC; O_CREAT] 0o666 in
```
aantron: > I had to generate a fd from the Unix module and transform it to an Lwt file descriptor. In the second test, I had to close the underlying Unix fd that I used to create the Lwt fd in order to flush everything before reading what has been written for the test. Should I close all the fd using the Unix.close at the end of the tests? I have the feeling that it is not really clean.

I recommend using `Lwt_unix.openfile` and `Lwt_unix.close`. I think we can rely, in these tests, on `Lwt_unix` tests having already run.

aantron: Yes, I think so.

cedlemo:
@aantron,

could you look at [this test](https://gist.github.com/cedlemo/ffec2ba4c461d246c461c3bb71b422a1) for the `recv` function. I created a thread for a server that listens in a unix socket and wait for the client to send a string. In another thread the client connect to the server and use the `Lwt_unix.recv` to get the string in a `Lwt_bytes.t`. My test blocks after those steps:
- server gen socket                                                                                                                                                    
- server bind socket                                                                                                                                    
- client gen socket                                                                                                                                     
- client connect socket

thanks.

aantron:
I don't see the problem immediately. When you say

> - client connect socket

Do you mean that print was triggered, or that the client actually was able to connect? The print comes before the `connect` call.

To check which call is blocking (or raising a synchronous exception), you can insert prints after each call, but outside the callback:

```
let p = Lwt_unix.connect … in
prerr_endline "if connect didn't block, this print runs immediately";
p >>= fun () ->
prerr_endline "this print runs only after connect succeeds";
...
```

Also, be sure to flush your output streams. It's possible that some output is buffered, but not visible to you. `prerr_endline` flushes STDERR after each print.

Finally, I recommend not using Unix domain sockets for these tests, because Unix domain sockets will not work on Windows.

aantron:
The test also looks non-deterministic: the client could run first.

cedlemo:
@aantron, in the last commit I had to make some choices:

- I don't catch exception for `Lwt_unix.gethostbyname` or `Lwt_unix.socket`. 
- I use an hard coded port for the socket to use. I have read that it should be better to let the system chose the port (https://stackoverflow.com/questions/1365265/on-localhost-how-do-i-pick-a-free-port-number for example.)
- I use `Lwt_unix.sleep` in order to be sure that when the client will try to read in the socket, the server is ready. I did not find a better way with `Lwt.join`, `Lwt.choose` or `Lwt.pick`.

test/unix/test_lwt_bytes.ml
```diff
@@ -530,4 +530,43 @@ let suite = suite "lwt_bytes" [
       >>= fun () ->
       Lwt.return check
     end;
+
+    test "bytes recv" begin fun () ->
+      let hostname = "127.0.0.1" in
+      let port = 20000 in
+      let get_socket_infos () =
+             Lwt_unix.gethostbyname hostname
```
aantron: You should be able to use [`Unix.inet_addr_loopback`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html#VALinet_addr_loopback) instead of calling `gethostbyname`.

test/unix/test_lwt_bytes.ml
```diff
@@ -530,4 +530,43 @@ let suite = suite "lwt_bytes" [
...
+             >>= fun entry ->
+             let addr = Array.get entry.Unix.h_addr_list 0 in
+             let sock = Lwt_unix.socket Lwt_unix.PF_INET Lwt_unix.SOCK_STREAM 0 in
+             let sockaddr = Lwt_unix.ADDR_INET (addr, port) in
+             Lwt.return (sock, sockaddr)
```
aantron: It's odd to pair the socket and the address. You should have one address that is shared between the client and server, but two different sockets. so that suggests one setup of `sockaddr` and two calls to `socket` at the appropriate places.

aantron:
Sorry, comment got cut.

> I use `Lwt_unix.sleep` in order to be sure that when the client will try to read in the socket, the server is ready. I did not find a better way with `Lwt.join`, `Lwt.choose` or `Lwt.pick`.

You just need to make the client wait on a promise that will be resolved by the server. This also lets you pass the server's system-assigned port to the client.

First create the promise:

```ocaml
let server_is_ready, notify_server_is_ready = Lwt.wait () in
...
```

in the server,

```ocaml
let server_address = Lwt_unix.getsockname server_socket in
Lwt.wakeup_later server_is_ready server_address;
```


At the beginning of the client:


```ocaml
server_is_ready
>>= fun server_address ->
(* The rest of the client *)
```

cedlemo:
@aantron, I have tested all the send/recv functions. Could you make a review? Do you think that I should try to refactor ?

test/unix/test_lwt_bytes.ml
```diff
@@ -496,4 +497,209 @@ let suite = suite "lwt_bytes" [
...
+    test "bytes recv" ~only_if:(fun () -> not Sys.win32) begin fun () ->
+      let server_is_ready, notify_server_is_ready = Lwt.wait () in
+      let buf = Lwt_bytes.create 6 in
+      let server () =
+        let sock = Lwt_unix.socket Lwt_unix.PF_INET Lwt_unix.SOCK_STREAM 0 in
+        let sockaddr = Lwt_unix.ADDR_INET (Unix.inet_addr_loopback, 0) in
+        Lwt_unix.bind sock sockaddr
+        >>= fun () ->
+        let server_address = Lwt_unix.getsockname sock in
+        Lwt.wakeup_later notify_server_is_ready server_address;
```
aantron: IIRC this should be moved under `listen`.

test/unix/test_lwt_bytes.ml
```diff
@@ -496,4 +497,209 @@ let suite = suite "lwt_bytes" [
       let check = "abcaaa" = Lwt_bytes.to_string buf in
       Lwt.return check
     end;
+
+    test "bytes read" begin fun () ->
+      let test_file = "bytes_io_data" in
+      Lwt_unix.openfile test_file [O_RDONLY] 0
+      >>= fun fd ->
+      let buf = Lwt_bytes.create 6 in
```
aantron: I recommend zeroing all new `Lwt_bytes.t`s in the test suite with `Lwt_bytes.fill`, because the contents can be anything when each buffer is first allocated.

test/unix/test_lwt_bytes.ml
cedlemo: You mean something like this ?:

```diff
diff --git a/test/unix/test_lwt_bytes.ml b/test/unix/test_lwt_bytes.ml
index 37d8a0c0..8ffca4ff 100644
--- a/test/unix/test_lwt_bytes.ml
+++ b/test/unix/test_lwt_bytes.ml
@@ -540,8 +540,8 @@ let suite = suite "lwt_bytes" [
         Lwt_unix.bind sock sockaddr
         >>= fun () ->
         let server_address = Lwt_unix.getsockname sock in
-        Lwt.wakeup_later notify_server_is_ready server_address;
         let () = Lwt_unix.listen sock 5 in
+        Lwt.wakeup_later notify_server_is_ready server_address;
         Lwt_unix.accept sock
         >>= fun (fd_client, _) ->
         Lwt_unix.write_string fd_client "abcdefghij" 0 9
```

aantron: Yes ;)

cedlemo:
@aantron,

tell me if there is something that can be improved. I will start the third part for the Lwt_bytes module after that.

test/unix/test_lwt_bytes.ml
```diff
@@ -6,6 +6,33 @@ open Test
...
+  let client () =
+    server_is_ready
+    >>= fun sockaddr ->
+    let sock = Lwt_unix.socket Lwt_unix.PF_INET Lwt_unix.SOCK_STREAM 0 in
+    Lwt_unix.connect sock sockaddr
+    >>= fun () ->
+    client_logic sock
+    >>= fun _n -> Lwt_unix.close sock
+  in
+  Lwt_main.run (Lwt.join [client (); server ()])
```
aantron: Sorry for not noticing this earlier, but the testing framework already calls `Lwt_main.run`: https://github.com/ocsigen/lwt/blob/c1e48e4807c3742bad632dc5905c71a815e5f042/test/test.ml#L204. Nested `Lwt_main.run` is not allowed, so this should just do the `join` without wrapping it in a `Lwt_main.run` call, and the code that calls this function should do `tcp_server_client_exchange >>= fun () -> ...` to wait for this to complete.

test/unix/test_lwt_bytes.ml
```diff
@@ -585,111 +606,63 @@ let suite = suite "lwt_bytes" [
...
+        let io_vectors = [Lwt_bytes.io_vector ~buffer ~offset ~length:6] in
+        Lwt_bytes.recv_msg ~socket ~io_vectors
       in
-      let client () =
-        server_is_ready
-        >>= fun sockaddr ->
-        let sock = Lwt_unix.socket Lwt_unix.PF_INET Lwt_unix.SOCK_DGRAM 0 in
-        Lwt_unix.connect sock sockaddr
+      let client_logic socket sockaddr =
+        Lwt_unix.connect socket sockaddr
```
aantron: Is this needed for the UDP socket? I have no experience with `send_msg`, but this looks suspicious.

test/unix/test_lwt_bytes.ml
```diff
@@ -33,6 +33,27 @@ let tcp_server_client_exchange server_logic client_logic =
...
+    >>= fun (_n, _sockaddr) -> Lwt_unix.close sock
+  in
+  let client () =
+    server_is_ready
+    >>= fun sockaddr ->
+    let sock = Lwt_unix.socket Lwt_unix.PF_INET Lwt_unix.SOCK_DGRAM 0 in
+    client_logic sock sockaddr
+    >>= fun (_n) -> Lwt_unix.close sock
+  in
+  Lwt_main.run (Lwt.join [client (); server ()])
```
aantron: Same comment as before about `Lwt_main.run`.

test/unix/test_lwt_bytes.ml
```diff
@@ -54,6 +54,11 @@ let udp_server_client_exchange server_logic client_logic =
   in
   Lwt_main.run (Lwt.join [client (); server ()])
 
+let gen_buf n =
+  let buf = Lwt_bytes.create n in
+  let () = Lwt_bytes.fill buf 0 n '0' in
```
aantron: Technically, this is filling it with `0x30`s, for zeroes you'd want `\x00`. But it doesn't matter.

test/unix/test_lwt_bytes.ml
cedlemo: Unlike with `Lwt_bytes.sendto socket message 0 9 [] sockaddr`, I need to "link" the socket to an address before calling `Lwt_bytes.send_msg ~socket ~io_vectors ~fds:[]`.

Furthermore according to the manpage of connect http://man7.org/linux/man-pages/man2/connect.2.html connect can be used with UDP socket too:

> If the socket sockfd is of type SOCK_DGRAM, then addr is the address
       to which datagrams are sent by default, and the only address from
       which datagrams are received.  If the socket is of type SOCK_STREAM
       or SOCK_SEQPACKET, this call attempts to make a connection to the
       socket that is bound to the address specified by addr.

I have to say that the `Lwt_bytes.send_msg` do not work like the  C system call : 

```
ssize_t sendmsg(int sockfd, const struct msghdr *msg, int flags);
```
so maybe there is a better solution.

aantron:
Looks good, thanks for the patch, and for the info about `send_msg` :)

PR #624, opened 2018-09-24
default_config: work when opam 2.0 is installed

avsm:
opam 2.0.0 has different behaviour for `opam list` exit code,
and so always passes. This results in libev being activated
in the default configuration even when conf-libev is not installed.

This PR uses opam config variables instead, which appears to
work on opam 1 and 2.  Since stdout needs to be captured,
it requires the Unix module for configure.exe.

It also works in the case where `opam` is not initialised,
such as in the platform monorepo where opam and lwt are being
built together from the same source tree.

With this changeset, the platform repo builds successfully
on the OS matrix with and without libev installed.

Signed-off-by: Anil Madhavapeddy <anil@recoil.org>

### DIFF
--- a/src/unix/config/configure.ml
+++ b/src/unix/config/configure.ml
@@ -35,15 +35,18 @@ let main () =
   let config_file = "lwt_config" in
 
   begin match !default_configuration with
-  | Some true ->
+  | Some true -> begin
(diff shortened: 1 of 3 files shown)


aantron:
Thanks!

Issue #625, opened 2018-09-30
about Lwt_bytes.mincore usage


cedlemo:
I am not able to use correctly `Lwt_bytes.mincore`, here is what I have tried:

```ocaml
      let test_file = "bytes_io_data" in
      let fd = Unix.openfile test_file [O_RDONLY] 0 in
      let shared = false in
      let size = 6 in
      let buffer = Lwt_bytes.map_file ~fd ~shared ~size () in
      let states = Array.make 1 false in
      let () = Lwt_bytes.mincore buffer 0 states in
```
this code throws: 

```
'mincore' raised 'Invalid_argument("Lwt_bytes.mincore")'
```

I looked at the [C code but it does not help much](https://github.com/ocsigen/lwt/blob/master/src/unix/unix_c/unix_mincore.c): 

```c
#ifdef HAVE_BSD_MINCORE
#define MINCORE_VECTOR_TYPE char
#else
#define MINCORE_VECTOR_TYPE unsigned char
#endif

CAMLprim value lwt_unix_mincore(value val_buffer, value val_offset,
                                value val_length, value val_states)
{
    long len = Wosize_val(val_states);
    MINCORE_VECTOR_TYPE vec[len];
    mincore((char *)Caml_ba_data_val(val_buffer) + Long_val(val_offset),
            Long_val(val_length), vec);
    long i;
    for (i = 0; i < len; i++) Field(val_states, i) = Val_bool(vec[i] & 1);
    return Val_unit;
}
```

any hints?

aantron:
The test is probably failing this check:

https://github.com/ocsigen/lwt/blob/ac8f7b15733bd9e3992764736d8860ca6c9fa379/src/unix/lwt_bytes.ml#L243

which *may* be too aggressive, because it checks there must be a number of *full* pages equal to the length of `states` after `offset` in the mapped buffer. This means the smallest possible buffer for which `Lwt_bytes.mincore` would usefully work has length 4096 (or 4M, depending on the page size).

To see if this is too restrictive, I suggest commenting out that check, and seeing if the underlying `mincore(2)` call still works, the test case asking about one page in a buffer of length 6. If it does work, we have caught Lwt's first bug with these tests :) In that case, we need to change the check to allow the last page queried to only partially overlap the buffer.

cedlemo:
It works when I comment the last check, I prepare you a PR that modifies the check in `mincore`.

PR #627, opened 2018-10-08
fix condition in mincore

cedlemo:
@aantron, 

#625 

I started to try to understand the behavior with different cases : 

when the last condition is : `offset > length buffer - (Array.length states * page_size)`

```ocaml
case where mem_len is > buffer_len

	buff_len = 6 offset = 0  pages = 1
		offset = 0
		mem_len = 1*4096
		offset > (buff_len - mem_len < 0) -> true -> Error

        buff_len = 4096 + 6 offset = 0  pages = 2
		offset = 0
		mem_len = 2*4096
		offset > (buff_len - mem_len < 0) -> true -> Error

	buff_len = 4096 + 6 offset = 4096 pages = 2
		offset = 4096
		mem_len = 2*4096
		offset > (buff_len - mem_len) < 0 -> true ->  Error

case where mem_len is < buffer_len

	buff_len = 4096 + 6 offset = 0 pages = 1
		offset = 0
		mem_len = 1*4096
		offset > buff_len - mem_len = 6  -> false -> ok

	buff_len = 4096 + 6 offset = 4096 pages = 1
		offset = 4096
		mem_len = 1*4096
		offset > (buff_len - mem_len) = 6 -> true ->  Error
```

and instead of *we need to change the check to allow the last page queried to only partially overlap the buffer.* like you said, I thought that we just have to be sure that the offset is not greater than the buffer length.
I have done the tests with the last condition `offset > buffer length` and it works as expected. 

```ocaml
case where mem_len is > buffer_len

	buff_len = 6 offset = 0  pages = 1
		offset = 0
		mem_len = 1*4096
		offset > buff_len -> false -> ok

        buff_len = 4096 + 6 offset = 0  pages = 2
		offset = 0
		mem_len = 2*4096
		offset > buff_len -> false -> ok

	buff_len = 4096 + 6 offset = 4096 pages = 2
		offset = 4096
		mem_len = 2*4096
		offset > buff_len > 0 -> false -> ok

case where mem_len is < buffer_len

	buff_len = 4096 + 6 offset = 0 pages = 1
		offset = 0
		mem_len = 1*4096
		offset > buff_len -> false -> ok

	buff_len = 4096 + 6 offset = 4096 pages = 1
		offset = 4096
		mem_len = 1*4096
		offset > buff_len -> false -> ok
```

I send this PR first for validation, then I will send you another PR with the tests for all the IO part of `Lwt_bytes`.

### DIFF
--- a/src/unix/lwt_bytes.ml
+++ b/src/unix/lwt_bytes.ml
@@ -240,7 +240,8 @@ external stub_mincore : t -> int -> int -> bool array -> unit = "lwt_unix_mincor
 let mincore buffer offset states =
   if (offset mod page_size <> 0
       || offset < 0
-      || offset > length buffer) then
+      || length buffer - offset >= (Array.length states - 1) * page_size + 1)
+  then
     invalid_arg "Lwt_bytes.mincore"
   else
     stub_mincore buffer offset (Array.length states * page_size) states
--- a/src/unix/lwt_bytes.ml
+++ b/src/unix/lwt_bytes.ml
@@ -240,7 +240,7 @@ external stub_mincore : t -> int -> int -> bool array -> unit = "lwt_unix_mincor
 let mincore buffer offset states =
   if (offset mod page_size <> 0
       || offset < 0
-      || length buffer - offset >= (Array.length states - 1) * page_size + 1)
+      || length buffer - offset < (Array.length states - 1) * page_size + 1)
   then
     invalid_arg "Lwt_bytes.mincore"
   else
--- a/src/unix/lwt_bytes.ml
+++ b/src/unix/lwt_bytes.ml
@@ -240,7 +240,7 @@ external stub_mincore : t -> int -> int -> bool array -> unit = "lwt_unix_mincor
 let mincore buffer offset states =
   if (offset mod page_size <> 0
       || offset < 0
-      || offset > length buffer - (Array.length states * page_size)) then
+      || offset > length buffer) then
     invalid_arg "Lwt_bytes.mincore"
   else
     stub_mincore buffer offset (Array.length states * page_size) states


src/unix/lwt_bytes.ml
```diff
@@ -240,7 +240,7 @@ external stub_mincore : t -> int -> int -> bool array -> unit = "lwt_unix_mincor
 let mincore buffer offset states =
   if (offset mod page_size <> 0
       || offset < 0
-      || offset > length buffer - (Array.length states * page_size)) then
+      || offset > length buffer) then
```
aantron: I think we want to make sure that the offset is far back enough in the buffer, that for a flag array for N pages, the first N-1 pages are fully in memory, and there is at least one byte for the last page.

So, we want there to be at least `(Array.length states - 1) * page_size + 1` bytes in the buffer starting at `offset`. So, we need something like `length buffer - offset >= (Array.length states - 1) * page_size + 1`. The only thing to watch out for is that `Array.length states - 1` might be negative, if the array is empty. It seems like the check will still give us the right behavior, though, but we should probably leave a comment about that in the code.

cedlemo: Ok, 
so if we want that there is at least `(Array.length states - 1) * page_size + 1` bytes in the buffer starting at `offset`, the condition that throws an exception for bad parameter is:

```
length buffer - offset < (Array.length states - 1) * page_size + 1
```
If I go through several cases like this:

* buff_len = 6 offset = 0 n_states = 1
```ocaml
6 - 0 < (1 - 1) * page_size + 1 (* -> false *)
```

* buff_len = page_size + 6 offset = 0 n_states = 1
```ocaml
page_size + 6 - 0 < (1 - 1) * page_size + 1 (* -> false *)
```

* buff_len = page_size + 6 offset = page_size n_states = 1

```ocaml
page_size + 6 - page_size < (1 - 1) * page_size + 1 (* -> false *)
```

* buff_len = page_size + 6 offset = page_size n_states = 2
```ocaml
page_size + 6 - page_size < (2 - 1) * page_size + 1 (*-> true (throw exception)*)
```

Do you want that I do all those tests (I have already almost done them) or do you see another case that are not covered?

aantron: I'd say to set the offset always to page_size, then do:

1. buffer length = page_size * 2, n_states = 1 (should be ok)

2. buffer length = page_size * 2, n_states = 2 (should fail)

3. buffer length = page_size * 2 + 1, n_states = 2 (should be ok)

I think that exercises the main corner cases completely and exactly, though something may be missing.

We should also test what happens when n_states = 0.

src/unix/lwt_bytes.ml
```diff
@@ -240,7 +240,8 @@ external stub_mincore : t -> int -> int -> bool array -> unit = "lwt_unix_mincor
 let mincore buffer offset states =
   if (offset mod page_size <> 0
       || offset < 0
-      || offset > length buffer - (Array.length states * page_size)) then
+      || length buffer - offset >= (Array.length states - 1) * page_size + 1)
```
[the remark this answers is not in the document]
cedlemo: argh you are right, I copy/pasted without checking sorry, the tests in the PR #628 are for `length buffer - offset < (Array.length states - 1) * page_size + 1`

aantron:
Thanks :) It happens, I guess I started talking in inverted terms (when the bounds are ok, rather than when to throw an exception). It was misleading, sorry about that.

PR #628, opened 2018-10-12
Test lwt bytes third part

cedlemo:
@aantron , this PR depends on the PR #627 because it contains the test for mincore.
I did not write test for `madvize`.

### DIFF
--- a/test/unix/test_lwt_bytes.ml
+++ b/test/unix/test_lwt_bytes.ml
@@ -772,15 +772,15 @@ let suite = suite "lwt_bytes" [
     test "wait_mincore offset < 0" ~only_if:(fun () -> not Sys.win32) begin fun () ->
       Lwt.catch
         (fun () ->
-        test_wait_mincore (Lwt_bytes.page_size * 2 + 1) (-1)
-        >>= fun _ -> Lwt.return false
+           test_wait_mincore (Lwt_bytes.page_size * 2 + 1) (-1)
+           >>= fun _ -> Lwt.return false
         )
         (function
-      | Invalid_argument message ->
-        if message = "Lwt_bytes.wait_mincore"
-        then Lwt.return true
-        else Lwt.return false
-      | _ -> Lwt.return false
+          | Invalid_argument message ->
+            if message = "Lwt_bytes.wait_mincore"
+            then Lwt.return true
+            else Lwt.return false
+          | _ -> Lwt.return false
         )
     end;
 
@@ -788,16 +788,16 @@ let suite = suite "lwt_bytes" [
       ~only_if:(fun () -> not Sys.win32) begin fun () ->
       Lwt.catch
         (fun () ->
-        let buff_len = Lwt_bytes.page_size * 2 + 1 in
-        test_wait_mincore buff_len (buff_len + 1)
-        >>= fun _ -> Lwt.return false
+           let buff_len = Lwt_bytes.page_size * 2 + 1 in
+           test_wait_mincore buff_len (buff_len + 1)
+           >>= fun _ -> Lwt.return false
         )
         (function
-      | Invalid_argument message ->
-        if message = "Lwt_bytes.wait_mincore"
-        then Lwt.return true
-        else Lwt.return false
-      | _ -> Lwt.return false
+          | Invalid_argument message ->
+            if message = "Lwt_bytes.wait_mincore"
+            then Lwt.return true
+            else Lwt.return false
+          | _ -> Lwt.return false
         )
     end;
-]
+  ]
--- a/test/unix/test_lwt_bytes.ml
+++ b/test/unix/test_lwt_bytes.ml
@@ -735,8 +735,8 @@ let suite = suite "lwt_bytes" [
 
     test "mincore buffer length = page_size * 2, n_states = 1"
       ~only_if:(fun () -> not Sys.win32) begin fun () ->
-           test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 1
-           >>= fun () -> Lwt.return true
+      test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 1
+      >>= fun () -> Lwt.return true
     end;
 
     test "mincore buffer length = page_size * 2, n_states = 2"
@@ -754,21 +754,21 @@ let suite = suite "lwt_bytes" [
 
     test "mincore buffer length = page_size * 2 + 1, n_states = 2"
       ~only_if:(fun () -> not Sys.win32) begin fun () ->
-           test_mincore (Lwt_bytes.page_size * 2 + 1) Lwt_bytes.page_size 2
-           >>= fun () ->
-           Lwt.return true
+      test_mincore (Lwt_bytes.page_size * 2 + 1) Lwt_bytes.page_size 2
+      >>= fun () ->
+      Lwt.return true
     end;
 
     test "mincore buffer length = page_size , n_states = 0"
       ~only_if:(fun () -> not Sys.win32) begin fun () ->
-           test_mincore (Lwt_bytes.page_size * 2 + 1) Lwt_bytes.page_size 0
-           >>= fun () -> Lwt.return true
+      test_mincore (Lwt_bytes.page_size * 2 + 1) Lwt_bytes.page_size 0
+      >>= fun () -> Lwt.return true
     end;
 
     test "wait_mincore correct bounds"
       ~only_if:(fun () -> not Sys.win32) begin fun () ->
-           test_wait_mincore (Lwt_bytes.page_size * 2 + 1) Lwt_bytes.page_size
-           >>= fun () -> Lwt.return true
+      test_wait_mincore (Lwt_bytes.page_size * 2 + 1) Lwt_bytes.page_size
+      >>= fun () -> Lwt.return true
     end;
 
     test "wait_mincore offset < 0"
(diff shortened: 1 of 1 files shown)


test/unix/test_lwt_bytes.ml
```diff
@@ -677,4 +713,91 @@ let suite = suite "lwt_bytes" [
...
+      ~only_if:(fun () -> not Sys.win32) begin fun () ->
+      test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 1
+      >>= fun states ->
+      Lwt.return (states.(0) = true)
+    end;
+
+    test "mincore buffer length = page_size * 2, n_states = 2"
+      ~only_if:(fun () -> not Sys.win32) begin fun () ->
+      try
+        test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 1
```
aantron: The number should be 2 here, not 1.

test/unix/test_lwt_bytes.ml
```diff
@@ -677,4 +713,91 @@ let suite = suite "lwt_bytes" [
...
+        >>= fun states ->
+        Lwt.return (states.(0) = true)
+      with
+      | Invalid_argument message ->
+        if message = "Lwt_bytes.mincore"
```
aantron: I think this string can be moved into the pattern, though IIRC that triggers a warning. We can either disable the warning by tagging the test with an `[@ocaml.warning]` annotation, or just not check for the string.

test/unix/test_lwt_bytes.ml
```diff
@@ -677,4 +713,91 @@ let suite = suite "lwt_bytes" [
...
+      >>= fun states ->
+      Lwt.return (states.(0) = true)
+    end;
+
+    test "mincore buffer length = page_size * 2, n_states = 2"
+      ~only_if:(fun () -> not Sys.win32) begin fun () ->
+      try
+        test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 1
+        >>= fun states ->
+        Lwt.return (states.(0) = true)
```
aantron: Tests in which we expect an exception shouldn't be able to finish with `true` if the exception isn't triggered. This should end with`Lwt.return false` unconditionally at this point.

test/unix/test_lwt_bytes.ml
```diff
@@ -59,6 +59,42 @@ let gen_buf n =
...
+  let buf_write = gen_buf buff_len in
+  Lwt_bytes.write fd buf_write 0 buff_len
+  >>= fun _n ->
+  Lwt_unix.close fd
+  >>= fun () ->
+  let fd = Unix.openfile test_file [O_RDONLY] 0 in
+  let shared = false in
+  let size = buff_len in
+  let buffer = Lwt_bytes.map_file ~fd ~shared ~size () in
+  Lwt_bytes.wait_mincore buffer offset
```
aantron: I think we should check that `mincore` returns `false` for the page being waited for. If we can't observe `false`, the `wait_mincore` tests aren't telling us anything, except that bounds checking is working.

If we can't get `false`, I suggest to rename the `wait_mincore` success test at the end to something like `wait_mincore correct bounds`, and leave a comment explaining why the test isn't testing the function "truly."

test/unix/test_lwt_bytes.ml
```diff
@@ -677,4 +713,91 @@ let suite = suite "lwt_bytes" [
...
+    test "mincore buffer length = page_size * 2, n_states = 1"
+      ~only_if:(fun () -> not Sys.win32) begin fun () ->
+      test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 1
+      >>= fun states ->
+      Lwt.return (states.(0) = true)
+    end;
+
+    test "mincore buffer length = page_size * 2, n_states = 2"
+      ~only_if:(fun () -> not Sys.win32) begin fun () ->
+      try
```
aantron: This `try` doesn't catch the rejection (it can only catch synchronous exceptions). I think it should be `Lwt.catch` instead (or `try%lwt` if we were using the PPX).

test/unix/test_lwt_bytes.ml
cedlemo: done

test/unix/test_lwt_bytes.ml
cedlemo: done

test/unix/test_lwt_bytes.ml
```diff
@@ -677,4 +713,92 @@ let suite = suite "lwt_bytes" [
...
+      Lwt.return (states.(0) = true)
+    end;
+
+    test "mincore buffer length = page_size * 2, n_states = 2"
+      ~only_if:(fun () -> not Sys.win32) begin fun () ->
+      Lwt.catch
+        (fun () ->
+        test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 2
+        >>= fun states ->
+        Lwt.return (states.(0) = false)
```
aantron: This test is still capable of passing even if the exception is not raised, if `states.(0)` turns out to be `false`. I think it needs to unconditionally do `Lwt.return false` on this line. Compare with the `wait_mincore` tests that expect an exception.

test/unix/test_lwt_bytes.ml
```diff
@@ -677,4 +713,92 @@ let suite = suite "lwt_bytes" [
...
+    test "page_size" begin fun () ->
+      let size = 4096 in
+      Lwt.return (size = Lwt_bytes.page_size)
+    end;
+
+    test "mincore buffer length = page_size * 2, n_states = 1"
+      ~only_if:(fun () -> not Sys.win32) begin fun () ->
+      test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 1
+      >>= fun states ->
+      Lwt.return (states.(0) = true)
```
aantron: I think this is brittle, because it will fail if `states.(0)` turns out to be `false`.

test/unix/test_lwt_bytes.ml
```diff
@@ -677,4 +713,92 @@ let suite = suite "lwt_bytes" [
...
+      | Invalid_argument _message -> Lwt.return true
+      | _ -> Lwt.return false
+        )
+    end;
+
+    test "mincore buffer length = page_size * 2 + 1, n_states = 2"
+      ~only_if:(fun () -> not Sys.win32) begin fun () ->
+      test_mincore (Lwt_bytes.page_size * 2 + 1) Lwt_bytes.page_size 2
+      >>= fun states ->
+      Lwt.return (states.(0) = true && states.(0) = true)
```
aantron: Same note about brittleness. Compare to the next test, which ignores `states` completely. In fact, it looks like we never want to check `_states` in these tests. The same comment as for `wait_mincore` applies. Checking `states` is only useful if we can force one of the `states` to reliably initially be `false`, then touch the page, and then we can expect it to reliably be `true`. We can then check these conditions.

If we can't have that, we should probably have `test_mincore` do `Lwt.return ()` instead of `Lwt.return states`. Alternatively, it should touch the pages (read them), to make sure that we can legitimately expect `true`, and then we check for `true`. I suggest also to move the comment to above both helper functions, explaining that we were only able to test bounds checking.

cedlemo: I don't think that we have the possibility to force the the states to be false, I checked the Unix module and I did not find any useful functions that could help me to do it.

test/unix/test_lwt_bytes.ml
```diff
@@ -93,7 +100,7 @@ let test_wait_mincore buff_len offset =
   >>= fun () ->
```
aantron: These next four lines should be redundant now.

test/unix/test_lwt_bytes.ml
```diff
@@ -732,55 +739,76 @@ let suite = suite "lwt_bytes" [
...
       ~only_if:(fun () -> not Sys.win32) begin fun () ->
-      test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 1
-      >>= fun states ->
-      Lwt.return (states.(0) = true)
+      Lwt.catch
+        (fun () ->
+           test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 1
+           >>= fun () -> Lwt.return true
+        )
+        (function
```
aantron: For the tests that we don't expect to raise an exception, I suggest not checking for it explicitly. The test framework will report it if it occurs. As an additional benefit, this will let us know if a different exception occurs. So, I think this test should be how it was, except for the new unconditional `Lwt.return true` at the end.

test/unix/test_lwt_bytes.ml
```diff
@@ -732,55 +739,76 @@ let suite = suite "lwt_bytes" [
...
-        >>= fun states ->
-        Lwt.return (states.(0) = false)
+           test_mincore (Lwt_bytes.page_size * 2) Lwt_bytes.page_size 2
+           >>= fun () -> Lwt.return false
         )
         (function
-      | Invalid_argument _message -> Lwt.return true
-      | _ -> Lwt.return false
+          | Invalid_argument _message -> Lwt.return true
+          | _ -> Lwt.return false
```
aantron: Here, it's better to reject the promise, so the testing framework reports which exception we actually got:

`| exn -> Lwt.fail exn`

test/unix/test_lwt_bytes.ml
cedlemo: yes

test/unix/test_lwt_bytes.ml
cedlemo: ok

test/unix/test_lwt_bytes.ml
aantron: I imagine opening a large-enough file will make it `false` fairly reliably, but we probably don't want to do that.

aantron:
Thanks! These were trickier than most.

PR #630, opened 2018-10-16
Test lwt unix sleeping and timeout functions

cedlemo:
#539,

I did not make a test for `Lwt_unix.yield`.

I am not sure of the usage of `Lwt_unix.auto_yield`. The documentation says:

> (** [auto_yield timeout] returns a function [f] that will yield
      every [timeout] seconds. *)

I expected this function to make my tests enter in a loop, but no it seems to have the same behavior than `Lwt.sleep`.

### DIFF
--- a/test/unix/test_sleep_and_timeout.ml
+++ b/test/unix/test_sleep_and_timeout.ml
@@ -39,8 +39,8 @@ let suite = suite "Lwt_unix sleep and timeout" [
 
     test "with_timeout : no timeout" begin fun () ->
       let duration = 1.0 in
-           Lwt_unix.with_timeout duration Lwt_unix.yield
-           >>= fun () -> Lwt.return_true
+      Lwt_unix.with_timeout duration Lwt_unix.yield
+      >>= fun () -> Lwt.return_true
     end;
 
     test "with_timeout : timeout" begin fun () ->
@@ -62,13 +62,13 @@ let suite = suite "Lwt_unix sleep and timeout" [
     end;
 
     test "yield" begin fun () ->
-       let bind_callback_ran = ref false in
-       Lwt.async (fun () -> Lwt.return () >|= fun () -> bind_callback_ran := true);
-       let bind_is_immediate = !bind_callback_ran in
-       let yield_callback_ran = ref false in
-       Lwt.async (fun () -> Lwt_unix.yield () >|= fun () -> yield_callback_ran := true);
-       let yield_is_immediate = !yield_callback_ran in
-       Lwt.return (bind_is_immediate && not yield_is_immediate)
+      let bind_callback_ran = ref false in
+      Lwt.async (fun () -> Lwt.return () >|= fun () -> bind_callback_ran := true);
+      let bind_is_immediate = !bind_callback_ran in
+      let yield_callback_ran = ref false in
+      Lwt.async (fun () -> Lwt_unix.yield () >|= fun () -> yield_callback_ran := true);
+      let yield_is_immediate = !yield_callback_ran in
+      Lwt.return (bind_is_immediate && not yield_is_immediate)
     end;
 
     test "auto_yield" begin fun () ->
@@ -78,14 +78,14 @@ let suite = suite "Lwt_unix sleep and timeout" [
         Lwt.async (fun () -> f () >|= fun () -> callback_ran := true);
         !callback_ran;
       in
-        let check1 = run_auto_yield () in
-        let check2 = run_auto_yield () in
-        Lwt_unix.sleep 1.0
-        >>= fun () ->
-        let check3 = run_auto_yield () in
-        let check4 = run_auto_yield () in
-        let check5 = run_auto_yield () in
-        let check = check1 && check2 && not check3 && check4 && check5 in
-        Lwt.return check
+      let check1 = run_auto_yield () in
+      let check2 = run_auto_yield () in
+      Lwt_unix.sleep 1.0
+      >>= fun () ->
+      let check3 = run_auto_yield () in
+      let check4 = run_auto_yield () in
+      let check5 = run_auto_yield () in
+      let check = check1 && check2 && not check3 && check4 && check5 in
+      Lwt.return check
     end;
   ]


test/unix/test_sleep_and_timeout.ml
```diff
@@ -0,0 +1,80 @@
...
+
+    test "auto_yield" begin fun () ->
+      let start_time = Unix.gettimeofday () in
+      let duration = 1.0 in
+      let wait = Lwt_unix.auto_yield duration in
+      wait ()
+      >>= fun () ->
+      let check = cmp_elapsed_time start_time duration in
+      Lwt.return check
+    end;
```
aantron: `let f = auto_yield timeout in` is something like:

- If it has been less than `timeout` time since the last time `f ()` was called, behave like `Lwt.return ()`.
- If it has been more than `timeout` time since the last time `f ()` was called, behave like `Lwt.yield ()`.

You can test whether `f ()` is yielding or not by comparing elapsed times. I think you want something like call `f ()`, it shouldn't yield (time should be near 0). Wait one second, call `f ()` again. It should now yield (time near 1). Call `f ()` immediately again, it should not yield. I'd then call it immediately a second time, and then wait one second and check for a second yield, just to make sure this is truly repeating.

Also, the docs of this function need improving. I'll open a separate issue.

This test I think doesn't "work" at all. I suspect the elapsed time is close to zero, and the test is succeeding because `cmp_elapsed_time` allows the elapsed time to be less than the duration right now.

test/unix/test_sleep_and_timeout.ml
```diff
@@ -0,0 +1,80 @@
...
+      Lwt.catch
+        (fun () ->
+           Lwt_unix.timeout duration
+           >>= fun () -> Lwt.return_false
+        )
+        (function
+          | Lwt_unix.Timeout ->
+            let check = cmp_elapsed_time start_time duration in
+            Lwt.return check
+          | _ -> Lwt.return false
```
aantron: I suggest to do `| exn -> Lwt.fail exn` here, so the test framework tells us what exception is raised if the wrong one is raised.

test/unix/test_sleep_and_timeout.ml
```diff
@@ -0,0 +1,80 @@
...
+      let duration = 1.0 in
+      Lwt.catch
+        (fun () ->
+           Lwt_unix.with_timeout duration Lwt_unix.yield
+           >>= fun () -> Lwt.return_true
+        )
+        (function
+          | Lwt_unix.Timeout ->
+            let check = cmp_elapsed_time start_time duration in
+            Lwt.return check
```
aantron: Even though this test is checking that `with_timeout` doesn't raise, it's still possible for it to raise, and the test to succeed. There shouldn't be a check in the exception case. In fact, you don't need to wrap in `Lwt.catch` at all. We can just let the test fail with the raised exception.

test/unix/test_sleep_and_timeout.ml
```diff
@@ -0,0 +1,80 @@
...
+          | _ -> Lwt.return false
+        )
+    end;
+
+    test "with_timeout does not throw exception" begin fun () ->
```
aantron: I'd rename this to `with_timeout: no timeout`, because the current name suggests that `with_timeout` doesn't throw exceptions :)

test/unix/test_sleep_and_timeout.ml
```diff
@@ -0,0 +1,80 @@
...
+        )
+        (function
+          | Lwt_unix.Timeout ->
+            let check = cmp_elapsed_time start_time duration in
+            Lwt.return check
+          | _ -> Lwt.return false
+        )
+    end;
+
+    test "with_timeout throws exception" begin fun () ->
```
aantron: Suggest `with_timeout: timeout`.

test/unix/test_sleep_and_timeout.ml
```diff
@@ -0,0 +1,80 @@
...
+        (fun () ->
+           Lwt_unix.with_timeout duration f
+           >>= fun () ->
+           Lwt.return false
+        )
+        (function
+          | Lwt_unix.Timeout ->
+            let check = cmp_elapsed_time start_time duration in
+            Lwt.return check
+          | _ -> Lwt.return false
```
aantron: Same note about using `Lwt.fail exn`.

test/unix/test_sleep_and_timeout.ml
aantron: Opened #631.

cedlemo: > This test I think doesn't "work" at all. I suspect the elapsed time is close to zero, and the test is succeeding because cmp_elapsed_time allows the elapsed time to be less than the duration right now

yes, that right.

cedlemo:
@aantron , 

I have the following test for `auto_yield` following your instructions 

```ocaml
    test "auto_yield" begin fun () ->
      let start_time = Unix.gettimeofday () in
      let duration = 1.0 in
      let f = Lwt_unix.auto_yield duration in
      f () (* first use of f *)
      >>= fun () ->
      if Unix.gettimeofday () -. start_time > 0.5 then Lwt.return_false (* this part works as expected *)
      else Lwt_unix.sleep duration
           >>= fun () ->
           let start_time = Unix.gettimeofday () in
           f () (* second use of f after one second *)
           >>= fun () ->
           let check = cmp_elapsed_time start_time duration in
           Lwt.return check (* this test fails *)
    end;
```

test/unix/test_sleep_and_timeout.ml
```diff
@@ -4,9 +4,11 @@
 open Test
 open Lwt.Infix
 
+(* None of the APIs make promises about how much larger the elapsed time will
+ * be, but they all promise that it won't be less than the expected time. *)
 let cmp_elapsed_time start_time expected_time =
   let elapsed_time = Unix.gettimeofday () -. start_time in
-  elapsed_time -. expected_time < 0.005
+  elapsed_time -. expected_time >= 0.0
```
aantron: Ok this is really a nit, but you could write `elapsed_time > expected_time` if keeping this condition :)

I still suggest maybe checking that it's within a factor of two, or maybe 10%, though.

cedlemo:
@aantron ,

when I try to play with your example, I have an error on this line:

    Lwt.async (fun () -> f () >|= callback_ran := true);

With this message:

      This expression has type bool ref
       but an expression was expected of type unit -> 'a
       Hint: Did you forget to wrap the expression using `fun () ->'?

I don't understand why.

aantron:
I forgot the `fun () ->` before `callback_ran := true`. It was OCaml-like pseudocode :)

cedlemo:
Sorry to bother @aantron, but I tried to follow your example with the following test:

```ocaml
    test "auto_yield" begin fun () ->
      let f = Lwt_unix.auto_yield 1.0 in
      let run_auto_yield () =
        let callback_ran = ref true in
        Lwt.async (fun () -> f () >|= fun () -> callback_ran := false);
        !callback_ran;
      in
      let test_loop () =
        let t1 = run_auto_yield () in
        let t2 = run_auto_yield () in
        if t1 || t2 then Lwt.return_false
        else Lwt.return_true
        (* else (Lwt_unix.sleep 1.0
             >>= fun () -> Lwt.return (not (run_auto_yield ())))*)
      in
      test_loop ()
      (* >>= fun check ->
      if check then test_loop ()
      else Lwt.return_false *)
    end;
```

You can see that:
* I inverted the value of `callback_ran`, its on purpose, the `if` condition at the end is easier to deal with
* I don't test yet auto_yield after one second (it is in comment). 

When I run this test, I expected that t1 and t2 were set to false so my test should return true. But it fails.

here is a patch for this branch to test it if you want

```diff
diff --git a/test/unix/test_sleep_and_timeout.ml b/test/unix/test_sleep_and_timeout.ml
index bb4ced73..3beb248b 100644
--- a/test/unix/test_sleep_and_timeout.ml
+++ b/test/unix/test_sleep_and_timeout.ml
@@ -70,4 +70,25 @@ let suite = suite "Lwt_unix sleep and timeout" [
           | exn -> Lwt.fail exn
         )
     end;
+
+    test "auto_yield" begin fun () ->
+      let f = Lwt_unix.auto_yield 1.0 in
+      let run_auto_yield () =
+        let callback_ran = ref true in
+        Lwt.async (fun () -> f () >|= fun () -> callback_ran := false);
+        !callback_ran;
+      in
+      let test_loop () =
+        let t1 = run_auto_yield () in
+        let t2 = run_auto_yield () in
+        if t1 || t2 then Lwt.return_false
+        else Lwt.return_true
+        (* else (Lwt_unix.sleep 1.0
+             >>= fun () -> Lwt.return (not (run_auto_yield ())))*)
+      in
+      test_loop ()
+      (* >>= fun check ->
+      if check then test_loop ()
+      else Lwt.return_false *)
+    end;
   ]
```

aantron:
@cedlemo, the test that's failing is the old `auto_yield` test above where you are adding this one (at least, if I only apply your patch, I am left with two `auto_yield` tests, the old one is the previous one we discussed).

This test is passing so far.

I recommend renaming `callback_ran`, if you invert its meaning. Also, setting whatever it is called to `false` in the callback makes it more difficult to reason about when reading the test. I suggest to leave its meaning the "natural" way, and change the condition instead.

Also, you can simply accumulate these conditions, like `let check_1 = run_auto_yield ()`, etc., and test them all in one big check at the end. That way, you don't have to write `if`s for early return. Then, the test will clearly show the logic that is being executed, without any interference from early returns.

I recommend not calling `test_loop` "loop" if it doesn't end up recursive. Also, without early return, you can just write a straight line of sequence expressions (`;`) and binds (`>>=`) to express the whole test.

aantron:
I'll write an easy issue about checking for duplicate test names, maybe tomorrow :)

cedlemo:
@aantron, I have pushed yield and auto_yield tests.

aantron:
Thanks!

The macOS build is broken due to some upstream issue. I'll wait a few days to see if it's resolved upstream, then work around it if not.

Issue #631, opened 2018-10-17
Improve docs of Lwt_unix.auto_yield


aantron:
Right now, [they say](http://ocsigen.org/lwt/4.1.0/api/Lwt_unix#VALauto_yield):

> auto_yield timeout returns a function f that will yield every timeout seconds.

which is pretty obscure.

[Here's a suggestion](https://github.com/ocsigen/lwt/pull/630#discussion_r225761693) for a clearer explanation:

> `let f = auto_yield timeout in ...` is something like:
>
> - If it has been less than timeout time since the last time `f ()` was called, behave like `Lwt.return ()`.
> - If it has been more than timeout time since the last time `f ()` was called, behave like `Lwt_unix.yield ()`.

The file to edit is [`Lwt_unix.cppo,mli`](https://github.com/ocsigen/lwt/blob/2b825c9c551025dd12656088c061a283ce710bb9/src/unix/lwt_unix.cppo.mli#L122-L124):

https://github.com/ocsigen/lwt/blob/2b825c9c551025dd12656088c061a283ce710bb9/src/unix/lwt_unix.cppo.mli#L122-L124

PR #634, opened 2018-10-29
Add auto_yield documentation

cedlemo:
#631,

I think that it would be great if we could explain what is `yield` and what problems solve `yield` and `auto_yield`.

For example I had a hard time understanding `yield` before you said `"yielded" in the old "thread" terminology` and that I found this definition: 

> yield() method pauses the currently executing thread temporarily for giving a chance to the remaining waiting threads of the same priority to execute.

How would you explain `yield`? can we say that `yield` pauses a promise in order to let the Lwt scheduler to run another promise?

### DIFF
--- a/src/unix/lwt_unix.cppo.mli
+++ b/src/unix/lwt_unix.cppo.mli
@@ -121,7 +121,22 @@ val yield : unit -> unit Lwt.t
 
 val auto_yield : float -> (unit -> unit Lwt.t)
   (** [auto_yield timeout] returns a function [f] that will yield
-      every [timeout] seconds. *)
+      every [timeout] seconds.
+      Consider a function f defined with let f = auto_yield timeout. If you run
+      multiple times this function, it will behave like either Lwt.return () or
+      Lwt_unix.yield depending on the elapsed time.
+
+      t = 0
+        f acts like Lwt.return ()
+      t <= timeout
+        f acts like Lwt.return ()
+      t >= timeout
+        f acts like Lwt_unix.yield ()
+
+      if f is re-ran after it has yielded, f will act like if t = 0.
+      An example of usage can be found in the test file
+      test/unix/test_sleep_and_timeout.ml.
+     *)
 
 exception Timeout
   (** Exception raised by timeout operations *)


src/unix/lwt_unix.cppo.mli
```diff
@@ -121,7 +121,22 @@ val yield : unit -> unit Lwt.t
 
 val auto_yield : float -> (unit -> unit Lwt.t)
   (** [auto_yield timeout] returns a function [f] that will yield
-      every [timeout] seconds. *)
+      every [timeout] seconds.
+      Consider a function f defined with let f = auto_yield timeout. If you run
+      multiple times this function, it will behave like either Lwt.return () or
```
aantron: This makes it seem like the user *has* to run `f ()` multiple times.

src/unix/lwt_unix.cppo.mli
```diff
@@ -121,7 +121,22 @@ val yield : unit -> unit Lwt.t
...
+      Lwt_unix.yield depending on the elapsed time.
+
+      t = 0
```
aantron: This condition is redundant with the second one.

src/unix/lwt_unix.cppo.mli
```diff
@@ -121,7 +121,22 @@ val yield : unit -> unit Lwt.t
...
+        f acts like Lwt.return ()
+      t <= timeout
```
aantron: These conditions have an intersection. If `t = timeout`, which one applies?

src/unix/lwt_unix.cppo.mli
```diff
@@ -121,7 +121,22 @@ val yield : unit -> unit Lwt.t
...
+        f acts like Lwt.return ()
+      t >= timeout
+        f acts like Lwt_unix.yield ()
+
+      if f is re-ran after it has yielded, f will act like if t = 0.
```
aantron: "re-r*u*n" for the subjunctive.

The comment about `t = 0` makes it seem like it will *always* behave like `Lwt.return` after the first `yield`, because it does not say that the timeout is restarted, and it says above that when `t = 0`, `f ()` is `Lwt.return ()`.

src/unix/lwt_unix.cppo.mli
```diff
@@ -121,7 +121,22 @@ val yield : unit -> unit Lwt.t
...
+      An example of usage can be found in the test file
+      test/unix/test_sleep_and_timeout.ml.
```
aantron: I suggest to inline an example. It should be something that can be copy-pasted by the user into a standalone program, like the little programs you can find in the docs in `lwt.mli`.

aantron:
> How would you explain `yield`? can we say that `yield` pauses a promise in order to let the Lwt scheduler to run another promise?

I wouldn't say that `yield` *pauses* something. Promises aren't threads, they can't be run, and so they can't be paused. Promises are inert pieces of data, like `ref`s. It is code in callbacks that is run.

This is what made the old *thread* terminology so misleading, and an obstacle to learning Lwt. Only changing the words "threads" to "promises," while keeping the surrounding description the same, means the description remains inaccurate and prevents users from understanding what Lwt actually does. The reason for the terminology change was not to replace "thread" by a synonym, but to choose *different* words that are more true to how Lwt works, because how threads work has little to do with how Lwt works.

Unfortunately, we still have the function name `yield`, because changing docs is much easier than changing APIs. It should really be called `defer`, `delay`, `a_bit_later`, `let_other_callbacks_run`, or something like that, so that should be kept in mind when reading this comment further.

`Lwt_unix.yield ()` creates a promise whose callbacks will be called *after*:

- Any other callbacks that can be immediately run.
- One poll for I/O by Lwt's I/O loop.

So, if you have `Lwt_unix.yield () >>= fun () -> foo`, the code `foo` will be run after any other callback that can already be run right now, including callbacks that become runnable after checking if there is any completed I/O.

This is in contrast to `Lwt.return () >>= fun -> foo`. In this expression, `foo` will run immediately.

Somewhere in the user's program, the user may have either:

- Some non-Lwt code that is CPU-intensive, and will run for a long time without calling into Lwt.
- Lwt code that is doing `>>=`, or otherwise calling into Lwt, but the promises that the user is adding callbacks to might tend to be already resolved most of the time.

If that's the case, that code will run continuously, without giving any other code that Lwt might be able to run (in callbacks that Lwt is ready to call) a chance to run, and also not giving Lwt a chance to "receive" any I/O. Then, the user should insert calls to `Lwt_unix.yield` into that code. Things like

```
some_code;
more_code;
```

become


```
some_code;
Lwt_unix.yield () >>= fun () ->
more_code;
```
 
Making sure that `some_code; more_code` is not a critical section, i.e. the program will not be harmed by other parts of the program running in between those two expressions.

How to summarize this is an interesting problem :)

raphael-proust:
Maybe it's difficult to summarise because `yield` is fundamentally an operation of the scheduler.

- `yield` is intrinsically complex because it requires a basic understanding of the scheduling of promises.
- Maybe the documentation should talk a bit about the scheduling. That'd make the explanation for `yield` a bit simpler.

I'd say we should have a small *Scheduling* section in the manual. And the documentation for `yield` should link to that section and then mention something along the lines of “`yield ()` is a promise that is scheduled at the back of the scheduling queue.”

Alternative names: `back_of_the_queue`, `schedule_myself_last`.

aantron:
We have a section like that, ["Execution model,"](http://ocsigen.org/lwt/4.1.0/api/Lwt#3_Executionmodel) though it needs more work, and it needs to be moved to its own page (the whole manual needs to be reorganized).

Note that promises aren't scheduled in Lwt (or any promise library) – that is a way of confusing them with threads. It is callbacks that are scheduled.

There is no promise scheduling, *except*, in a way, for `yield` itself, because there is a queue of promises created by `yield` for Lwt to resolve after the next I/O poll. However, it's not really important that the *promise* is put into the queue. `yield` could have been implemented to take one callback directly, putting it into a queue of bare callbacks, returning `unit` and no promise, and achieved its purpose that way. It's a design accident that this is a queue of promises, so it really shouldn't be thought of as any kind of *promise scheduler*.

Promises exist somewhere in memory, retained by the user's reference. The Lwt "loop," the closest thing it has to a scheduler, doesn't know that most promises even exist (does not retain references to them), and has no way of finding them. But, when something calls `Lwt.wakeup_later` on a pending promise, Lwt calls all of the callbacks that were added to the promise. So it's the *callbacks* that are scheduled.

The only promises that Lwt indirectly retains references to, are the ones that are inside closures, that will call `Lwt.wakeup_later` on promises in response to the completion of asynchronous I/O that the user asked Lwt to start. But this is again not about the scheduling of promises, but about how Lwt waits/polls for I/O. When some I/O completes, Lwt calls a callback that was associated with that I/O, which often happens to resolve a promise, which then triggers more callbacks that Lwt can (and will) call, etc.

cedlemo:
@aantron, thanks for all of your explications, this is really interesting. I did some modifications but I am not really satisfied with it.

aantron:
Yeah, I would even suggest to seriously consider leaving this for later. It might be much easier to solve the `yield`/`auto_yield` docs issue when the surrounding docs give readers better background information that can be relied upon in the `yield` docs.

cedlemo:
I totally agree.

PR #635, opened 2018-11-12
Expose Lwt_io.is_closed

andreas:
This PR exposes `Lwt_io.is_closed`. The intent is for this to be analogous to [`Async_unix.Reader.is_closed`](https://ocaml.janestreet.com/ocaml-core/latest/doc/async_unix/Async_unix/Reader/#val-is_closed).


aantron:
Thanks!

PR #637, opened 2018-11-20
Fix Lwt_io.of_bytes

### LINKED ISSUE #636: Lwt_io.of_bytes + position don't work
I you run the following in utop:
```ocaml
let ichan = Lwt_io.of_bytes ~mode:Lwt_io.input (Lwt_bytes.of_string "abcd") in
Lwt_io.position ichan;;
```
you get:
```ocaml
- : int64 = -4L
```
which obviously is a bug as you can't `set_position` on a negative value.

I'm not familiar with `Lwt`'s internals but had a look and it appears `of_bytes` behaves inconsistently with the broader `make`.
[Here](https://github.com/ocsigen/lwt/blob/master/src/unix/lwt_io.ml#L525), in make, `max` depends on the mode while [here](https://github.com/ocsigen/lwt/blob/master/src/unix/lwt_io.ml#L555), in `of_bytes`, it doesn't.

I found it by writing tests for a function that is supposed to reset the position after processing the channel. I assume it should work on regular channels as if it wasn't, someone would probably have found out by now. Still I'll try it out and let you know.

I'm happy to submit a fix if this is indeed the issue here!

NathanReb:
Fixes #636

My understanding is that it's `offset` that's improperly set in `Lwt_io.of_bytes`. Since `Lwt_bytes` based channels just use the `Lwt_bytes` straight as their internal buffer and never actually perform any io, `offset` is never updated. That means we can consider that everything has been already read in case of an input channel since it's already in the internal buffer. In the output channel case, writing to the internal buffer is sufficient and we never need to flush meaning offset should be set to the end of the buffer here as well.

Let me know if that seems right to you!

I'd like to add some tests for this. They should be in `ls test/unix/test_lwt_io.ml` right ?
I was thinking about testing the initial position after creating a channel from bytes. Let me know if you have any brighter ideas for good tests here!

### DIFF
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -545,38 +545,32 @@ let make :
    | Output -> Outputs.add outputs wrapper);
   wrapper
 
-let of_bytes :
-  type m.
-  mode : m mode ->
-  Lwt_bytes.t ->
-  m channel
-  =
-  fun ~mode bytes ->
-    let length = Lwt_bytes.length bytes in
-    let abort_waiter, abort_wakener = Lwt.wait () in
-    let rec ch = {
-      buffer = bytes;
-      length = length;
-      ptr = 0;
-      max = length;
-      close = lazy(Lwt.return_unit);
-      abort_waiter = abort_waiter;
-      abort_wakener = abort_wakener;
-      main = wrapper;
-      (* Auto flush is set to [true] to prevent writing functions from
-         trying to launch the auto-fllushed. *)
-      auto_flushing = true;
-      mode = mode;
-      offset = (match mode with
-          | Output -> 0L
-          | Input -> Int64.of_int length);
-      typ = Type_bytes;
-    } and wrapper = {
-        state = Idle;
-        channel = ch;
-        queued = Lwt_sequence.create ();
-      } in
-    wrapper
+let of_bytes (type m) ~(mode : m mode) bytes =
+  let length = Lwt_bytes.length bytes in
+  let abort_waiter, abort_wakener = Lwt.wait () in
+  let rec ch = {
+    buffer = bytes;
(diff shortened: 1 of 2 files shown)


aantron:
Great, taking a look!

Yes, `test/unix/test_lwt_io.ml` is a good place for the tests. I would test the initial position, the position after some I/O is done, and the position after `set_position`, on both `Lwt_bytes` and ordinary channels. Feel free to write only a subset of those tests :)

aantron:
The reasoning seems correct for `Input`, but I think it will give the wrong result for `Output` (I guess it's important to vary `Input`/`Output` in the test cases as well).

`offset` seems to be the number of bytes either actually read into the buffer, or actually written *from* the buffer to the outside world, not *to* the buffer from the user (when `set_position` is not being used). That means that when an `Lwt_bytes` channel is first created, with this change, the position that will be reported will be the end of the buffer, but I think it should be zero instead.

Our writing-from-the-buffer operation is just a no-op in the `Lwt_bytes` case.

aantron:
Please disregard the noisy macOS build rows :) I need to adjust timeouts in some of the tests.

NathanReb:
Yes I just realized by writing the tests and reading the code further.
I did add a bunch of tests (which were definitely useful since I found out about the new bug I introduced), both for output and input but just for `Lwt_bytes` since the only function I modified was `of_bytes`.

I'll gladly add more tests for regular channels in a separate PR if find some time to do so!

NathanReb:
Is there anything blocking here? Can I do something to help get it merged?

aantron:
Nothing inherent to the PR :) I'm just continuously busy then traveling, but I should be able to review this fully and very likely merge later today, or early tomorrow.

aantron:
Thanks!

PR #640, opened 2018-11-27
ppx: recognize constrained catch-all cases 

Octachron:
This is a small tweak to lwt's ppx that makes it recognize `(_:ty)` as a catch-all pattern, avoiding the generation of a redundant case in

```OCaml
try%lwt
...
with (error: exn) -> ...
```


aantron:
Thanks!

Drup:
Nice!

Issue #644, opened 2018-12-11
Add Lwt_unix.​IO_vectors.length


madroach:
Add ``Lwt_unix.​IO_vectors.length`` to calculate the length of the resulting write resp. available space for read.

PR #645, opened 2018-12-14
Lwt_unix.IO_vectors.length

raphael-proust:
addresses #644

### DIFF
--- a/src/unix/lwt_unix.cppo.ml
+++ b/src/unix/lwt_unix.cppo.ml
@@ -681,6 +681,10 @@ struct
 
   let create () = {prefix = []; reversed_suffix = []; count = 0}
 
+  let length {prefix; reversed_suffix; _} =
+    (List.fold_left (fun acc {length; _} -> acc + length) 0 prefix)
+    + (List.fold_left (fun acc {length; _} -> acc + length) 0 reversed_suffix)
+
   let append io_vectors io_vector =
     io_vectors.reversed_suffix <- io_vector::io_vectors.reversed_suffix;
     io_vectors.count <- io_vectors.count + 1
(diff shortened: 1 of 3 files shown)


src/unix/lwt_unix.cppo.ml
```diff
@@ -681,6 +681,10 @@ struct
 
   let create () = {prefix = []; reversed_suffix = []; count = 0}
 
+  let length {prefix; reversed_suffix; _} =
+    (List.fold_left (fun acc {length; _} -> acc + length) 0 prefix)
+    + (List.fold_left (fun acc {length; _} -> acc + length) 0 reversed_suffix)
```
aantron: I suggest factoring out the common code on these two lines.

src/unix/lwt_unix.cppo.mli
```diff
@@ -409,6 +409,11 @@ sig
   (** [is_empty vs] is [true] if and only if [vs] has no I/O vectors, or all I/O
       vectors in [vs] have zero bytes. *)
 
+  val length : t -> int
```
aantron: I think it's better to rename this, so it's completely clear from the name that it is not the number of vectors, but the number of bytes in the vectors. Something like `byte_count` may work.

src/unix/lwt_unix.cppo.mli
```diff
@@ -409,6 +409,11 @@ sig
...
+  (** [length vs] is the total length of the I/O vectors of [t]. In other words,
+      it is the sum of all the lengths of I/O vectors that have been appended
+      minus the sum of all the bits that have been dropped. *)
```
aantron: The part about dropping seems confusing. Why not just say it's the number of bytes in the I/O vectors in `t`? It seems that the `drop` docs are sufficient to know that bytes can be dropped from the front of `t`, and so won't be counted here. I think the user can assume that an API is logically consistent with itself :)

raphael-proust:
I addressed all three comments in a second commit. I think the two commits should be squashed if they are to be merged.

aantron:
Great, thanks for doing this! I'll merge after CI completes.

PR #646, opened 2018-12-18
Add Lwt_stream.of_seq

hcarty:
The Seq module is in stdlib since OCaml 4.07.0 and there is a seq
compatibility package available on opam for older versions of OCaml.

`Lwt_stream.of_seq` gives us a way to lazily traverse a `_ Seq.t` via `Lwt_stream`.

### DIFF
--- a/test/core/test_lwt_stream.ml
+++ b/test/core/test_lwt_stream.ml
@@ -33,13 +33,19 @@ let suite = suite "lwt_stream" [
 
   test "of_seq"
     (fun () ->
-       let seq = fun () -> Seq.Cons (1, Seq.empty) in
+       let x = ref false in
+       let nil = fun () -> x := not !x; Seq.Nil in
+       let seq = fun () -> Seq.Cons (1, nil) in
        let stream = Lwt_stream.of_seq seq in
+       let x_before = !x in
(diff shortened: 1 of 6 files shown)


hcarty:
Travis failed on 4.05.0 but seems to be unrelated to this PR.

test/core/test_lwt_stream.ml
```diff
@@ -31,6 +31,17 @@ let suite = suite "lwt_stream" [
        t3 >>= fun x3 ->
        return ([x1; x2; x3] = [1; 1; 1]));
 
+  test "of_seq"
+    (fun () ->
+       let seq = fun () -> Seq.Cons (1, Seq.empty) in
```
aantron: To test "The elements are evaluated from [s] and pushed onto the stream as the stream is consumed," I suggest to cause side effects (setting a ref) on evaluation of either the element 1 or `Nil`, check the side effects are not triggered at the beginning, and are triggered after the corresponding `Lwt_stream.get` call.

hcarty: Sorry for the looong delay on this!  69a1f2b should implement something like you're talking about.

aantron:
Thank you!

PR #652, opened 2019-01-10
Lwt_unix: missing header on Android

mjustus:
Building lwt on Android currently fails because the macro `LWT_NOT_AVAILABLE2` from `lwt_unix.h` is not included in `unix_wait4.c`.


aantron:
Thank you!

PR #656, opened 2019-01-29
check for duplicate test names per suite

### LINKED ISSUE #632: Catch duplicate test names in the test framework
We should edit [one of these two functions](https://github.com/ocsigen/lwt/blob/ce0b32fb1180f2b0d454b78e2ad99d456c1f4ef9/test/test.ml#L85-L109) in the test framework so that it raises an exception if two [tests](https://github.com/ocsigen/lwt/blob/ce0b32fb1180f2b0d454b78e2ad99d456c1f4ef9/test/test.ml#L6-L10) in the same suite have the same test name. Not checking this has tripped up developers during test refactoring.

To test the fix, temporarily edit any test suite ([example](https://github.com/ocsigen/lwt/blob/ce0b32fb1180f2b0d454b78e2ad99d456c1f4ef9/test/core/test_lwt_mvar.ml#L17-L27)) so that two tests have the same name.

See [`CONTRIBUTING.md`](https://github.com/ocsigen/lwt/blob/master/docs/CONTRIBUTING.md#Workflow) for instructions on how to clone the repo and run tests.

metadave:
Here's some sample output: (with 2 tests modified to use the name `basic take`)

```
$ make test
dune build
dune runtest -j 1 --no-buffer
Done: 890/904 (jobs: 1)Fatal error: exception Test.Duplicate_Test_Names("suite:lwt_mvar test:basic take")
        main alias test/core/runtest (exit 2)
(cd _build/default/test/core && ./main.exe)
Done: 891/904 (jobs: 1)Testing library 'ppx'...
.................
```

### Notes
- I realize that `contains_dup_tests` is Not Very Efficient, but it seemed like a simple implementation so I kept it.
- I removed/renamed a few tests that were flagged as duplicates
- closes #632

### DIFF
--- a/test/core/test_lwt.ml
+++ b/test/core/test_lwt.ml
@@ -2060,7 +2060,7 @@ let nchoose_split_tests = suite "nchoose_split" [
     end [@ocaml.warning "-4"]
   end;
 
-  test "pending, rejected" begin fun () ->
+  test "pending, rejected 2" begin fun () ->
     let p, r = Lwt.wait () in
     let p = Lwt.nchoose_split [p; fst (Lwt.wait ())] in
     Lwt.wakeup_exn r Exception;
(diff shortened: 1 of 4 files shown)


aantron:
Thanks!

PR #659, opened 2019-02-13
Lwt.ignore: convenience function

jrochel:
Once more I'm testing the waters to see whether such trivial (but convenient) contributions are welcome. If so, more (more useful and complex ones) might follow...

### DIFF
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -1495,6 +1495,8 @@ sig
   val return_error : 'e -> (_, 'e) Result.result t
   val return_nil : _ list t
 
+  val ignore : _ -> unit t
+
   val fail_with : string -> _ t
   val fail_invalid_arg : string -> _ t
 end =
@@ -1516,6 +1518,7 @@ struct
   let return_false = return false
   let return_ok x = return (Result.Ok x)
   let return_error x = return (Result.Error x)
+  let ignore _ = return_unit
 
   let fail_with msg =
     to_public_promise {state = Rejected (Failure msg)}
@@ -1643,7 +1646,7 @@ struct
 
         let State_may_have_changed p' =
           resolve ~allow_deferring:false p' p_result in
-        ignore p'
+        Pervasives.ignore p'
       in
 
       let remove_the_callback =
@@ -1675,7 +1678,7 @@ struct
 
         let State_may_have_changed p' =
           resolve ~allow_deferring:false p' p_result in
-        ignore p'
+        Pervasives.ignore p'
       in
       add_implicitly_removed_callback p_callbacks callback;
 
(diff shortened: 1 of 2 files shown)


hcarty:
I think a function named `Lwt.ignore` may be more useful as:
```ocaml
val ignore : _ Lwt.t -> unit Lwt.t

let ignore promise =
  bind promise (fun _ -> Lwt.return_unit)
```

jrochel:
You might be right. However in practise I come across the pattern `x; Lwt.return_unit` much more often. I could just write `Lwt.return x` if `x` were a `unit` value as in `Lwt.return @@ print_endline "Jehova's Witnesses"` but the word return does give a wrong impression here.

So I think that while the `_ Lwt.t -> unit Lwt.t` does make a bit more sense it is IMO rarely applicable and I don't miss a bit.

hcarty:
FWIW, I've had the opposite experience - I have no need for `ignore x; Lwt.return_unit` but do sometimes have a use for `bind x (fun _ -> Lwt.return_unit)`.

Maybe a name like `wrap_ignore` for the function proposed in this PR, to match the naming of the existing `wrap*` functions?  Though looking at your proposal again, it's not wrapping a function, so I'm not sure the extra definition is worth it.

aantron:
This saves a small amount of typing for the writer of code, but the reader of code has to learn an extra Lwt API. Its exact behavior is not self-evident. There is the basic issue of the exact types and language-level behavior, as suggested by @hcarty. More troubling to me, in the minds of many Lwt users, Lwt has to do with I/O and side effects. What does it mean *for Lwt* to "ignore" something? Is it as simple as ignoring a result, or is there a non-trivial effect on side effects? Writing out `ignore x; Lwt.return_unit` makes it immediately clear what exactly is being ignored *and how*, and what behavior of Lwt is being triggered (in this case, only the simplest one).

Myself, as a writer writing for readers, I would choose to write out `x; Lwt.return_unit` each time, relying on simple constructs that readers will be aware of anyway.

@hcarty That ignore is `Lwt.map ignore p`, also `p >|= ignore`.

I would suggest not to merge this.

> Once more I'm testing the waters to see whether such trivial (but convenient) contributions are welcome. If so, more (more useful and complex ones) might follow...

Trivial contributions are not welcome/unwelcome as a class, but each one on a case-by-case basis. So, please submit/suggest anything else you may have, regardless of the outcome for this one :)

hcarty:
@aantron Agreed, and thanks for the `Lwt.map ignore p` tip!

aantron:
@hcarty I should note that the `map` version allocates one more promise than your version with `Lwt.return_unit`. However, that should only be a consideration in code that is known to be a performance bottleneck.

jrochel:
Alright, fine. Next time I'll try something more useful. :-)

Issue #361, opened 2017-05-18
Factor out Lwt_sequence (doubly-linked list)


aantron:
[`Lwt_sequence`](https://ocsigen.org/lwt/3.0.0/api/Lwt_sequence) ([.mli](https://github.com/ocsigen/lwt/blob/6d1a16c4672c71abd7fc3d3f11e0739f2079e9cb/src/core/lwt_sequence.mli), [.ml](https://github.com/ocsigen/lwt/blob/6d1a16c4672c71abd7fc3d3f11e0739f2079e9cb/src/core/lwt_sequence.ml)) is pretty much an ordinary mutable doubly-linked list. As in the `Lwt_pqueue` issue (#360), it belongs in the standard library.

Unlike `Lwt_pqueue`, `Lwt_sequence` is actually referenced elsewhere in the Lwt API, in two places: [`Lwt.add_task_r`](https://ocsigen.org/lwt/3.0.0/api/Lwt#VALadd_task_r) and [`Lwt.add_task_l`](https://ocsigen.org/lwt/3.0.0/api/Lwt#VALadd_task_l). `Lwt_sequence` will therefore need to stick around, or become a bunch of aliases, even if the stdlib gets a linked list. 

Perhaps we should also start gradually deprecating these two functions. Their functionality can be implemented in terms of other Lwt functions, and they seem to be used in few places. I see them in use in:

- [fdopen/uwt](https://github.com/fdopen/uwt) in several places (OPAM package `uwt`)
- [nojb/ocaml-utp](https://github.com/nojb/ocaml-utp/blob/bbe647bf09faad438961a3a2cbeb4168be33f0ce/lib/utp_lwt.ml#L292) (OPAM package `utp`)
- [ocamllabs/ocamlot](https://github.com/ocamllabs/ocamlot/blob/1304889b97af994d0a9e4db7e9618e9a995ef138/lib/ocamlot.ml#L514)
- [ANSSI-FR/parsifal](https://github.com/ANSSI-FR/parsifal/blob/4dd5f7e93642c52c9a53005fe4e068af98d60f8f/lwt/lwt_semaphore.ml#L32)

avsm:
ocamllabs/ocamlot is not in active use, so I've archived that repository

aantron:
This seems good to me.

aantron:
We can later add a link to lwt-dllist in the `Lwt_sequence` deprecation message.

avsm:
[...]

PR #660, opened 2019-02-14
Abstract Lwt_sequence from Lwt_main hooks APIs

aantron:
Part of #361.

cc @toolslive -- just want to check by you that this API is sufficient for the needs of your projects :) Thanks for getting me moving on this.


toolslive:
Yes this is (more than) sufficient.

PR #661, opened 2019-02-14
Travis: test on 4.08.0+beta1

aantron:
This is currently blocked on ocamlbuild not having a compatible release, see https://github.com/ocaml/ocamlbuild/issues/294. ocamlbuild is an indirect dependency of `lwt_react`, which we test in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,9 @@ matrix:
     - os: linux
       env: COMPILER=4.07.1
     - os: linux
-      env: COMPILER=4.07.1 FLAMBDA=yes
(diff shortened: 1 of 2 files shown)


aantron:
Probably also blocked on https://github.com/ocaml-ppx/ocaml-migrate-parsetree/issues/50.

gasche:
(ocamlbuild 0.14.0 is now released and should work with 4.08)

aantron:
We are testing on 4.08 in Travis since ac19689caf622cccac663b6741889e07cd2b0bee, though as an allowed failure for now.

Issue #667, opened 2019-02-21
Include Let_syntax module for ppx_let


brendanlong:
I'm trying to convert some code from Async to Lwt and it would be convenient if I could easily keep using ppx_let. ppx_let is nice because it works with other monad modules, and has syntax for both `map` and `bind`.

I could add the module myself but it gets messy.

I think all you need to do is add this module somewhere to `Lwt.Infix`:

```
module Let_syntax = struct
  let return = Lwt.return
  let map t ~f = Lwt.map f t
  let bind t ~f = Lwt.bind t f
  let both a b =
    bind a ~f:(fun a ->
      map b ~f:(fun b ->
        a, b))

  module Open_on_rhs = Lwt.Infix
end
```

(Ideally `Lwt.both` would exist too and then this could just reference it)

aantron:
This seems like a good idea.

If we add this to `Lwt.Infix`, is it correct that it will be used as:

```ocaml
open Lwt.Infix

(* ... *)

let%bind x = foo in (* ... *)
```

Since `Lwt.Infix` will be already open to bring `Lwt.Infix.Let_syntax` into scope, do we still need the line

```ocaml
module Open_on_rhs = Lwt.Infix
```

?

Drup:
Note that the need for ppx_let will be greatly reduced starting next version, with let operators.

On the other hand, with let operators, we will implement something that look exactly like that module *anyway*, so ...

brendanlong:
Yeah, it should make ppx_let just automatically work if `Lwt.Infix` is open.

I think you're right that Open_on_rhs isn't necessary. I'm slightly confused about what that piece of the definition is for, but I think you're fine without it.

aantron:
@brendanlong would you be willing to make a quick PR for this that suits your needs? I'll be happy to review, test, and merge it.

brendanlong:
@aantron I created #668 for this. Let me know if I didn't approach this right or something.

Issue #325, opened 2017-03-10
More general join and choose operators


yallop:
The [`<&>` operator][join] has the following type:

```ocaml
val ( <&> ) : unit t -> unit t -> unit t
```

It'd be useful to have a variant that works for any promise type, returning both results, with the following type: 

```ocaml
val ( <&> ) : 'a t -> 'b t -> ('a * 'b) t
```

that behaved something like this:

```ocaml
let (<&>) l r =
   let lv = ref None and rv = ref None in
   (    (l >>= fun v -> lv := Some v; return_unit)
    <&> (r >>= fun v -> rv := Some v; return_unit)) >|= fun () ->
   match !lv, !rv with
   | Some lv, Some rv  -> (lv, rv)
   | _ -> assert false
```

Similarly, the [`<?>`][choose] operator has the following type:

```ocaml
val ( <?> ) : 'a t -> 'a t -> 'a t
```

It'd be useful to have a variant that doesn't constrain the result types to be the same, and that preserves the information about which promise delivered the result, with the following type:

```ocaml
val ( <?> ) : 'a t -> 'b t -> ('a, 'b) either t
```

that behaved something like this:

```ocaml
let (<?>) l r =
     map (fun l -> Left l) l
 <?> map (fun r -> Right r) r
```

[join]: https://github.com/ocsigen/lwt/blob/ce9b201587ff3dec5c9fef07eee8a6421c74a9ea/src/core/lwt.mli#L246-L247
[choose]: https://github.com/ocsigen/lwt/blob/ce9b201587ff3dec5c9fef07eee8a6421c74a9ea/src/core/lwt.mli#L243-L244

c-cube:
This seems interesting to me, but I'm wondering if it's worth breaking compatibility. Maybe providing both versions would help, e.g. with `<??>` and `<&&>` as the generalized versions?

mfp:
On Fri, Mar 10, 2017 at 04:04:52AM -0800, Simon Cruanes wrote:
> This seems interesting to me, but I'm wondering if it's worth breaking
> compatibility. Maybe providing both versions would help, e.g. with `<??>`
> and `<&&>` as the generalized versions?

My gut reaction is a very strong "don't break compatibility over this".
Typing `<&&>` is just a tiny little bit less convenient than `<&>`, and this
improved convenience in new code that only targets newer Lwt is minuscule
compared to the major inconvenience of fixing older code, or rewriting
[...]

PR #668, opened 2019-02-22
Add support for ppx_let

brendanlong:
See #667

### DIFF
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -3072,6 +3072,18 @@ struct
   let (=|<) = map
   let (<&>) p p' = join [p; p']
   let (<?>) p p' = choose [p; p']
+
+  module Let_syntax =
+  struct
+    let return = return
+    let map t ~f = map f t
+    let bind t ~f = bind t f
+    let both = both
+
+    module Open_on_rhs =
+    struct
+    end
+  end
 end
 include Infix
 
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -2445,6 +2445,7 @@ sig
   val async : (unit -> _ t) -> unit
   val ignore_result : _ t -> unit
 
+  val both : 'a t -> 'b t -> ('a * 'b) t
   val join : unit t list -> unit t
 
   val choose : 'a t list -> 'a t
@@ -2567,6 +2568,16 @@ struct
     in
 
     attach_callback_or_resolve_immediately ps
+
+  let both p1 p2 =
+    let v1 = ref None in
+    let v2 = ref None in
+    let p1' = bind p1 (fun v -> v1 := Some v; return_unit) in
+    let p2' = bind p2 (fun v -> v2 := Some v; return_unit) in
+    join [p1'; p2'] |> map (fun () ->
+      match !v1, !v2 with
+      | Some v1, Some v2 -> v1, v2
+      | _ -> assert false)
 
 
 
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -891,8 +891,8 @@ v}
 (** {3 Multiple wait} *)
 
 val both : 'a t -> 'b t -> ('a * 'b) t
-(** [Lwt.both a b] returns a promise that is pending until both promises in
-    the arguments [a] and [b] become {{: #TYPEt} {e resolved}}.
+(** [Lwt.both p_1 p_2] returns a promise that is pending until {e both} promises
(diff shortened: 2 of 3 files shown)


hcarty:
Could this be put in a separate module, something like `Lwt.Ppx_let`?  That way the infix operators aren't pulled into scope along with the `Let_syntax` module.

aantron:
I'm also not sure about where this should go. A separate module seems reasonable, since, presumably, the two preferences, using infix operators, and using ppx_let, are mutually exclusive.

@brendanlong I suggest omitting `Lwt.both`, and only defining it in the `Let_syntax` module.

brendanlong:
>  presumably, the two preferences, using infix operators, and using ppx_let, are mutually exclusive.

It's very common to mix them since point-free coding style is still useful (for the same reason `|>` is common in code that also uses `let`), so I suspect almost every user of ppx_let will also want the infix operators in scope. For what it's worth, Jane Street libraries put `Let_syntax` inside of the `Monad_infix` modules.

You can see examples of this in [sentry-ocaml](https://github.com/brendanlong/sentry-ocaml/blob/master/src/client.ml#L54) and [mssql](https://github.com/arenadotio/ocaml-mssql/blob/master/src/client.ml#L249) (even using `>>|` inside of `let%map`).

That said, it's also true that people opening `Lwt.Infix` who don't use ppx_let won't necessarily want `Let_syntax` in scope. I'm not sure what the downside of this is but I can move it if you want.

I can also move `both` although it seems strange to have it and only expose it in the ppx_let bindings. I couldn't really follow the implementation of `Lwt.join`, so I'm also wondering if I implemented `both` correctly or if I should have done something different?

aantron:
Ok, then let's leave it in `Lwt.Infix`.

The issue with `Lwt.both` is its rejection semantics. The implementation favors the `'a t`: as soon as it is rejected, the state (or fate) of the `'b t` doesn't matter, even if it was rejected first, or is still pending. Once the `'a t` is rejected, first or last, the rejection is immediately propagated to the `('a * 'b) t`. However, if the `'b t` is rejected first, this `Lwt.both` continues to "wait" for the `'a t`, potentially forever.

These asymmetric semantics might make sense if `Lwt.both` is only bound under `Lwt.Infix.Let_syntax.both`, and is therefore mostly hidden behind `and` syntax, because in that case, at least the `let` promise arises first syntactically, and we can imagine that there is some kind of evaluation order issue.

But, for an explicit `Lwt.both`, with such a type signature, I would expect the semantics to be symmetric, that is, in some sense, commutative. These two pieces of code:

```
Lwt.on_failure (Lwt.both a b) f
Lwt.on_failure (Lwt.both b a) f
```

should be equivalent.

If you'd like to add `Lwt.both` to `Lwt`, I think the easiest way is to take advantage of the existing `Lwt.join`, and use `Result.result ref`s to store values/exceptions.

Alternatively, I can merge a version of this PR with only `Lwt.Infix.Let_syntax.both`.

Or, please discuss :)

aantron:
See also #325, "More general join and choose operators."

brendanlong:
Hm yeah maybe it makes sense to implement `Lwt.both` in terms of `Lwt.join` for now, and then optimize it later if someone has time (to ensure that we don't get stuck maintaining a weird implementation for backwards compatibility). Do you think it makes sense to add that as `Lwt.both` or should it stay in `Lwt.Infix.Let_syntax.both` for now?

aantron:
An `Lwt.both` implemented in terms of `Lwt.join` (thus with its semantics) would be a welcome addition to the main module, `Lwt` :) There might be some trickiness with cancelation semantics (another Lwt annoyance :/), but there is a high chance of getting them right, and if anything is wrong, I can comment and give options on how to adjust the implementation.

aantron:
@brendanlong I'd like to release Lwt 4.2.0 soon (#675), with this PR included. Would you mind if I took over and added the new `Lwt.both`? Or, you are very welcome if you have the time to do it :)

brendanlong:
@aantron That would be fine to me. Sorry, I wanted to make a better version of `both` but haven't had time to look into this much again. If you want to do it that would be great!

aantron:
@brendanlong I replaced (most of) the first commit (adding `Lwt.both`) by a different implementation, then cherry-picked your docs and `Let_syntax` module, and added some new tests (spread over the commits). Thanks!

I should note that the `Lwt.both` in the new first commit is structurally identical to `<&>` proposed by @yallop in #325.

PR #669, opened 2019-02-22
OpenBSD no longer has mincore syscall - add required #ifdef dance

krwesterback:
OpenBSD issued an errata on January 27, 2019 to remove the mincore syscall:

 011: SECURITY FIX: January 27, 2019   All architectures
The mincore() system call can be used to observe memory access patterns of other processes.
A source code patch exists which remedies this problem.

The actual patch for 6.4 is here:

https://ftp.openbsd.org/pub/OpenBSD/patches/6.4/common/011_mincore.patch.sig

This branch allows lwt to compile on OpenBSD -current and 6.4 with the errata applied.

To make lwt compile on patched 6.3, the condition would be OpenBSD >= 201805.

I just run -current so I used >= 201811.

### DIFF
--- a/src/unix/unix_c/unix_mincore.c
+++ b/src/unix/unix_c/unix_mincore.c
@@ -19,7 +19,7 @@
 LWT_NOT_AVAILABLE4(unix_mincore)
 #elif defined __OpenBSD__
 #include <sys/param.h>
-#if OpenBSD >= 201811
+#if OpenBSD >= 201805
 LWT_NOT_AVAILABLE4(unix_mincore)
 #endif
 #else
--- a/src/unix/unix_c/unix_mincore.c
+++ b/src/unix/unix_c/unix_mincore.c
@@ -22,7 +22,6 @@ LWT_NOT_AVAILABLE4(unix_mincore)
 #if !defined SYS_mincore
 LWT_NOT_AVAILABLE4(unix_mincore)
 #endif
-#endif
 #else
 
 #ifdef HAVE_BSD_MINCORE
--- a/src/unix/unix_c/unix_mincore.c
+++ b/src/unix/unix_c/unix_mincore.c
@@ -18,9 +18,10 @@
 #ifdef __CYGWIN__
 LWT_NOT_AVAILABLE4(unix_mincore)
 #elif defined __OpenBSD__
-#include <sys/param.h>
-#if OpenBSD >= 201805
+#include <sys/syscall.h>
+#if !defined SYS_mincore
 LWT_NOT_AVAILABLE4(unix_mincore)
+#endif
 #endif
 #else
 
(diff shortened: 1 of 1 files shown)


aantron:
Looks good, thanks!

> To make lwt compile on patched 6.3, the condition would be OpenBSD >= 201805

I would suggest making the condition `>= 201805`, then. Is that wrong under some circumstances?

krwesterback:
>= 201805 should not be wrong in any circumstances. I just didn't think of it until I read the errata. :-)

Should I create a new pull request? Does the first one just get closed somehow?

krwesterback:
Or maybe I just push a commit to the branch. :-)

If that was wrong, let me know.

aantron:
Just a commit is right :)

Just to finally confirm, 201805 is not the original 6.3, but only the patched 6.3 with the erratum applied?

aantron:
Looking in the OpenBSD errata (6.3 and 6.4), the patches are from January 2019. This makes me suspect that neither 201811 nor 201805 are the right constants, since OpenBSD 6.3 and 6.4 still have `mincore`, although it is "neutered" by the patch. If possible, I'd suggest to use a constant that matches the first version of OpenBSD that actually won't have `mincore`.

I guess we don't know what that constant will be, but we can probably get away with using 201901.

aantron:
Thinking about this more, in OpenBSD current, is the constant still 201811?

Also,

> This branch allows lwt to compile on OpenBSD -current and 6.4 with the errata applied.
>
> To make lwt compile on patched 6.3, the condition would be OpenBSD >= 201805.

Did the patch remove the system call in 6.3 and 6.4, or just make it do nothing useful?

aantron:
Ok, looking at the OpenBSD source, I see that OpenBSD -current still has 201811 for `OpenBSD`, and no apparent way to tell that it is -current rather than 6.4.

I see that commits about removing `mincore` also removed `#define SYS_mincore 78` from `sys/syscall.h`. Is this file installed and visible to the user on OpenBSD? If so, the right macro magic might be

```c
#elif defined __OpenBSD__
#include <sys/syscall.h>
#if !defined SYS_mincore
LWT_NOT_AVAILABLE4(unix_mincore)
#endif
```

...and, if this is fragile, we can change that to whatever date constant in `sys/param.h` OpenBSD defines for its next release, once it is available.

aantron:
I was looking here: https://github.com/openbsd/src/commit/54e4f6b9a1dc183e1dc7c4a722c2646aa309246a#diff-4568f7b51aa0a96555ce3672b472f01dL241

krwesterback:
The values of OpenBSD are only changed by releases. i.e. the only values will be the dates
the version ticks up from 6.2 to 6.3 (201805), 6.3 to 6.4 (201811), 6.4 to 6.5 (likely 201905), etc.

So setting it to 201901 wouldn't allow lwt to work with -current, although it would start working with 6.5.

Checking the syscall file directly would likely work, but I'm not sure if including the file would have side effects. I can give it a whirl tomorrow.

krwesterback:
the direct test for SYS_mincore after #include <sys/syscall.h> works fine. lwt, lambda-term both build w/o compaint.

So I'm happy with this approach.

aantron:
Thanks for this PR!

PR #673, opened 2019-03-15
Add tests for user id and group id

anuragsoni:
relates to #539

Signed-off-by: Anurag Soni <anurag@sonianurag.com>

### DIFF
--- a/test/unix/test_lwt_unix.cppo.ml
+++ b/test/unix/test_lwt_unix.cppo.ml
@@ -726,6 +726,39 @@ let lwt_preemptive_tests = [
   end;
 ]
 
+let lwt_user_tests = [
+  test "getlogin and Unix.getlogin" ~only_if:(fun () -> not Sys.win32) begin fun () ->
+    let unix_user = Unix.getlogin () in
+    Lwt_unix.getlogin () >>= fun user ->
+    Lwt.return (user = unix_user)
+  end;
+  test "getpwnam and Unix.getpwnam" ~only_if:(fun () -> not Sys.win32) begin fun () ->
+    let unix_user = Unix.getlogin () in
+    let unix_password = Unix.getpwnam unix_user in
+    Lwt_unix.getpwnam unix_user >>= fun password ->
+    Lwt.return (password = unix_password)
+  end;
+  test "getpwuid and Unix.getpwuid" ~only_if:(fun () -> not Sys.win32) begin fun () ->
+    let pwnam = Unix.getpwnam (Unix.getlogin ()) in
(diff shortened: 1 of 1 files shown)


anuragsoni:
The MacOS builds failed. I'll try and find out how travis does the setup for it. I don't have access to a Mac but i'd have expected these calls to work there.

aantron:
I don't have immediate access to a Mac I can try this on. However, looking at the tests, I think they need to be adjusted.

It looks like all the tests use `getlogin` to get a user name for testing purposes, and then call the various functions on it. This is reasonable for `getpwnam` and `getpwuid`, because those functions expect a user name. However, for `getgrnam` and `getgrgid`, you should provide a group name, not a user name. What I suspect is happening is that on Ubuntu in Travis, there is a group with the same name as the user under which the tests are running, but this is not the case in Travis on macOS, and it also won't be the case in general.

aantron:
(Respectively, group id and user id for `getgrgid` and `getpwuid`).

anuragsoni:
Ah i completely missed that i was using a user login name to retrieve groups. I've updated the tests to use group ids instead. The `Not_found` exception should have been a hint :)

test/unix/test_lwt_unix.cppo.ml
```diff
@@ -726,6 +726,47 @@ let lwt_preemptive_tests = [
...
+    Lwt.return (password = unix_password)
+  end;
+  test "getpwuid and Unix.getpwuid" ~only_if:(fun () -> not Sys.win32) begin fun () ->
+    let pwnam = Unix.getpwnam (Unix.getlogin ()) in
+    let unix_pwuid = Unix.getpwuid pwnam.pw_uid in
+    Lwt_unix.getpwuid pwnam.pw_uid >>= fun pwuid ->
+    Lwt.return (pwuid = unix_pwuid)
+  end;
+  test "getgrgid and Unix.getgrgid" ~only_if:(fun () -> not Sys.win32) begin fun () ->
+    let groups = Unix.getgroups () in
```
aantron: Have you considered `Unix.getgid` instead? It looks like it should always return a gid for the test to use, so you don't have to have the pattern-matching below, and a fall-through case.

aantron:
Happens to us all :)

test/unix/test_lwt_unix.cppo.ml
anuragsoni: Updated to use `Unix.getgid` instead.

aantron:
Great, thanks for doing this :)

PR #678, opened 2019-03-21
unix_termios_conversion.c: Support for higher baud rates.

GreenArchon:
This makes Lwt_unix.tcsetattr/tcgetattr functions compatible with the corresponding Unix library functions at higher baudrates.

These extra baudrates are in OCaml's standard library since v.4.02 (see https://github.com/ocaml/ocaml/issues/6781 for details). Code is taken directly from there.

### DIFF
--- a/src/unix/unix_c/unix_termios_conversion.c
+++ b/src/unix/unix_c/unix_termios_conversion.c
@@ -48,6 +48,9 @@ static struct {
                   {B110, 110},
                   {B134, 134},
                   {B150, 150},
+#ifdef B200
+                  {B200, 200},
+#endif
                   {B300, 300},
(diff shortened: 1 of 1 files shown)


aantron:
Thanks. Disregard the AppVeyor failure. Hopefully this will be fixed soon. If not, we will move the build to allowed failures.

aantron:
Looks good, will merge after Travis finishes.

It looks like the commit that merged this support into the compiler was actually merged for 4.03: https://github.com/ocaml/ocaml/commit/860c670848440f791d1b9c68a1ace8fb629da234. However, Lwt support doesn't look like it depends on the compiler having the corresponding code, so it should work in 4.02 as well.

PR #679, opened 2019-03-27
Esy compatibility for libev detection

anmonteiro:
Esy does not have an `opam` binary in the path, and the libev detection introduced in https://github.com/ocsigen/lwt/commit/c1e48e4807c3742bad632dc5905c71a815e5f042 and released in 4.2.0 does not work with esy.

The proposed fix defaults to looking for the environment variables that the libev packages for esy sets (https://github.com/esy-packages/libev/blob/master/package.json#L16-L26)

### DIFF
--- a/src/unix/config/configure.ml
+++ b/src/unix/config/configure.ml
@@ -52,7 +52,7 @@ let main () =
         use_libev := Some false
       | exception Not_found ->
         use_libev := Some false
-      | flags, libs -> ()
+      | _flags, _libs -> ()
   end
   | _ ->
     ()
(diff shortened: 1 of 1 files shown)


aantron:
Thanks, I'll be able to take a good look in 48+ hours. We'll do a 4.2.1 release after fixing this (or is this something that can/should be patched in the esy overlay?).

anmonteiro:
This would need a patch release, since we can't use Lwt + libev at all in esy with 4.2.0 right now.

aantron:
Thanks!

jordwalke:
Thank you.

aantron:
(FYI releasing this today to opam)

PR #693, opened 2019-06-08
Better custom configuration

hongchangwu:
I was tripped up by the example in the travis script when trying to
build lwt with libev without opam.
`dune exec src/unix/configurator.exe -- -use-libev true` doesn't really
do anything since the `lwt_config` file will be generated at the project root
instead of where dune expects it.

Also update the `dune` file to allow proper fallback. Currently dune
gives a warning when a file is both in the source tree and generated by a rule.
This will be an error in future versions of dune.

### DIFF
--- a/src/unix/config/configure.ml
+++ b/src/unix/config/configure.ml
@@ -32,10 +32,11 @@ let args = [
 let main () =
   Arg.parse args ignore usage;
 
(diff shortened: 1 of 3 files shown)


aantron:
Thanks!

PR #694, opened 2019-06-13
Allow setting the cwd when spawning a process

### LINKED ISSUE #163: Lwt_process.exec should support a chdir option
Sometimes we need to start a background process in a specific directory – containing resources or whatever else. It would be very easy to support in `Lwt_process.exec` by just adding an optional `chdir` in the appropriate branch of the fork. If we do not want to support this in `Lwt_process.exec` this leads to absurd complications because a the current working directory must be protected by a global _mutex_ and every part of the application needs to collaborate on it – esp. third party libraries creating files must be wrapped.

talex5:
Fixes #163.

This patch adds a new `?cwd` argument after every `?env` argument in `Lwt_process`.

Note: Windows changes have not been tested. Someone with access to a Windows machine should try it :-)

### DIFF
--- a/src/unix/lwt_process.ml
+++ b/src/unix/lwt_process.ml
@@ -45,7 +45,7 @@ let win32_get_fd fd redirection =
     Some fd'
 
 external win32_create_process :
-  string option -> string -> string option ->
+  string option -> string -> string option -> string option ->
   (Unix.file_descr option * Unix.file_descr option * Unix.file_descr option) ->
     proc = "lwt_process_create_process"
 
@@ -56,7 +56,7 @@ let win32_quote arg =
     Filename.quote arg
 
 let win32_spawn
-    (prog, args) env
+    (prog, args) env ?cwd
     ?(stdin:redirection=`Keep)
     ?(stdout:redirection=`Keep)
     ?(stderr:redirection=`Keep)
@@ -88,7 +88,7 @@ let win32_spawn
   and stderr_fd = win32_get_fd Unix.stderr stderr in
   let proc =
     win32_create_process
-      (if prog = "" then None else Some prog) cmdline env
+      (if prog = "" then None else Some prog) cmdline env cwd
       (stdin_fd, stdout_fd, stderr_fd)
   in
   let close = function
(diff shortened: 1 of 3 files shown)


src/unix/lwt_process.ml
```diff
@@ -153,6 +153,15 @@ let unix_spawn
     List.iter Unix.close toclose;
     begin
       try
+        begin match cwd with
+          | None -> ()
+          | Some dir ->
+            try Sys.chdir dir
+            with ex ->
+              let err = Printf.sprintf "chdir(%S) failed: %s\n" dir (Printexc.to_string ex) in
+              let _ : int = Unix.write_substring Unix.stderr err 0 (String.length err) in
```
aantron: I don't think Lwt should have the child process write to STDERR. I suggest to let the surrounding `try` handler handle this error the same as it handles the rest of the potential errors in this part of the code. If that's not sufficient, we can improve all the error handling together in the future.

talex5: Having Lwt do the chdir is an alternative for processes that don't support the `-C` option. Processes that are able to chdir themselves usually respond in this way (writing a message to stderr and exiting). e.g.

```bash
~ $ make -C /missing >/dev/null
make: *** /missing: No such file or directory.  Stop.
~ $ git -C /missing show  >/dev/null
fatal: cannot change to '/missing': No such file or directory
```

Without this, it is very difficult to debug errors (e.g. permission denied) because it will be reported instead as the process failing to start.

If the parent process is behaving correctly then it won't be passing an invalid directory here and so this won't make any difference. If it is passing an invalid directory, then getting a message about that is surely an improvement?

aantron: I agree that the existing error handling in child processes is bad. What I am saying is that I prefer not to:

1. Special-case it for `chdir`. For comparison, if the parent process passed a bad program path, or the child process can't be started due to a similar error (e.g. another permission denied), we don't print to STDERR in the surrounding code.

2. Commit to printing to STDERR from a library (Lwt). In particular, we shouldn't choose "for `make`" and "for `git`" which message to print, whether to print a message (or respond otherwise), etc. `make` and `git` aren't even printing from a `fork` here, but if they were, and were using Lwt, they would have no control over what is shown (or whether it is shown).

We need a better solution for all these errors, not just `chdir`.

I'm just proposing to push that to a later PR, so we can do it as necessary, and not conflate `chdir` with fixing the error handling. Unless you'd like to do that, of course.

aantron:
@talex5, I merged this with the error message removed.

PR #696, opened 2019-07-04
Fix #695

### LINKED ISSUE #695: Add the missing flags for madvise
For example:
```
MADV_HUGEPAGE (since Linux 2.6.38)
    Enables Transparent Huge Pages (THP) for pages in the range specified by addr and length. Currently, Transparent Huge Pages only work with private anonymous pages (see mmap(2)). The kernel will regularly scan the areas marked as huge page candidates to replace them with huge pages. The kernel will also allocate huge pages directly when the region is naturally aligned to the huge page size (see posix_memalign(2)). This feature is primarily aimed at applications that use large mappings of data and access large regions of that memory at a time (e.g., virtualization systems such as QEMU). It can very easily waste memory (e.g., a 2MB mapping that only ever accesses 1 byte will result in 2MB of wired memory instead of one 4KB page). See the Linux kernel source file Documentation/vm/transhuge.txt for more details. The MADV_HUGEPAGE and MADV_NOHUGEPAGE operations are only available if the kernel was configured with CONFIG_TRANSPARENT_HUGEPAGE. 
MADV_NOHUGEPAGE (since Linux 2.6.38)
    Ensures that memory in the address range specified by addr and length will not be collapsed into huge pages. 
MADV_DONTDUMP (since Linux 3.4)
    Exclude from a core dump those pages in the range specified by addr and length. This is useful in applications that have large areas of memory that are known not to be useful in a core dump. The effect of MADV_DONTDUMP takes precedence over the bit mask that is set via the /proc/PID/coredump_filter file (see core(5)). 
MADV_DODUMP (since Linux 3.4)
    Undo the effect of an earlier MADV_DONTDUMP. 
```
Currently there are only:
```ocaml
type advice =
  | MADV_NORMAL
  | MADV_RANDOM
  | MADV_SEQUENTIAL
  | MADV_WILLNEED
  | MADV_DONTNEED
```

There are others as well: https://linux.die.net/man/2/madvise

XVilka:
While [working with huge data chunks](https://discuss.ocaml.org/t/working-with-a-huge-data-chunks/3955) learned about [madvise](https://linux.die.net/man/2/madvise), checked that some options are not available, and decided to add.


hannesm:
you may need some conditional compilation for this, they're not part of the posix standard (see https://pubs.opengroup.org/onlinepubs/9699919799/functions/posix_madvise.html), but a GNU/Linux extension (FWIW, on FreeBSD there's in addition to the posix defined ones: `MADV_FREE | MADV_NOSYNC | MADV_AUTOSYNC | MADV_NOCORE | MADV_CORE | MADV_PROTECT`).

XVilka:
It is more complicated than that. Seems certain values available depending on the kernel configuration options, e.g. HUGEPAGES, but at the same time they are quite important to have. Not sure how to proceed.

aantron:
We could support this with discovery. If a flag is not present, we could define its value as zero.

aantron:
@XVilka, we need to add discovery code to `discover.ml` for each flag that is supported on only some platforms.

Do you have information about whether these flags are supported/not supported in groups, or should we test for each flag individually?

aantron:
I have a branch in which I started rewriting `discover.ml` using Configurator, so I'll finish that up. If you want, wait for that branch, and it should be much easier to add the detection code then.

aantron:
@XVilka, the new `discover.ml` is in (#699). If you rebase this PR on `master`, you can add new feature tests in the `Features` module. Here's an example of a simple one:

https://github.com/ocsigen/lwt/blob/e4f70d66aa4a5f965bdaff78b43e0dd1c44ab821/src/unix/config/discover.ml#L449-L463

If this PR needs several very similar feature tests, you can make a templated one like so:

https://github.com/ocsigen/lwt/blob/e4f70d66aa4a5f965bdaff78b43e0dd1c44ab821/src/unix/config/discover.ml#L520-L564

There's a comment at the top of `discover.ml` that should put everything into context, and help with debugging:

https://github.com/ocsigen/lwt/blob/e4f70d66aa4a5f965bdaff78b43e0dd1c44ab821/src/unix/config/discover.ml#L6-L58

In the translation lookup table, for each constant, we'll probably need something like

```c
#if defined(HAVE_THIS_CONSTANT)
THIS_CONSTANT,
#else
0,
#endif
```

aantron:
Closing to keep the repo tidy. Please comment if you'd like to pick this up again!

XVilka:
Yes, will rebase in a couple days.

XVilka:
I force pushed the changes, but for some reason it didn't show in the PR. Detection is easy. The question how to define the `type advice` conditionally?

PR #712, opened 2019-09-06
Run Lwt_unix tests concurrently

### LINKED ISSUE #708: Run tests concurrently and/or parallelize the test suite

aantron:
This decreases testing time for `Lwt_unix` from about 47 seconds to about 6 seconds, and the overall testing time is reduced from 53 seconds to 12 seconds (all measurements on my system).

In the remaining 6 seconds:

- 2 seconds are taken up by some tests that must be run sequentially, because they use `Lwt.async_exception_hook`. Actually, most of the 2 seconds is due to only one test, which spends most of that time in `Lwt_unix.sleep`. We can probably refactor or delete it, if we ever really care about these seconds.
- The last concurrent test to finish is started at 3.1 seconds. This is one of the many test that consists mostly of waiting on `Lwt_unix.sleep`, which is the only reason it takes this long.
- The last concurrent test to start also starts around 3.1 seconds.

So, after the 11 sequential tests run for 2 seconds, all the remaining tests, 183 concurrent tests, start and finish within the next 4 seconds. So, the concurrency level is very high.

Resolves #708.

cc @CraigFe this is probably relevant for https://github.com/mirage/alcotest/issues/177 :)

The implementation in this PR is pretty sloppy, because I just coded as fast as possible, and the Lwt tester is private to Lwt. However, it shows what some projects can gain from running tests concurrently.

### DIFF
--- a/test/core/test_lwt.ml
+++ b/test/core/test_lwt.ml
@@ -1466,7 +1466,7 @@ let on_success_tests = suite "on_success" [
     later (fun () -> !f_ran = true)
   end;
 
-  test "fulfilled, f raises" begin fun () ->
+  test ~sequential:true "fulfilled, f raises" begin fun () ->
     let saw = ref None in
     let restore =
       set_async_exception_hook (fun exn -> saw := Some exn) in
@@ -1497,7 +1497,7 @@ let on_success_tests = suite "on_success" [
     later (fun () -> !f_ran = true)
   end;
 
-  test "pending, fulfilled, f raises" begin fun () ->
+  test ~sequential:true "pending, fulfilled, f raises" begin fun () ->
     let saw = ref None in
     let p, r = Lwt.wait () in
     Lwt.on_success p (fun () -> raise Exception);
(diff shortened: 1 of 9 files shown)


aantron:
(and of those 4 seconds of concurrent testing, there is probably only slightly over 1 second of CPU time used, and the extra 3 seconds are from inflation by `Lwt_unix.sleep`).

aantron:
For future reference, some commands I used to extract information from the testing logs:

- List names of tests which finished:

    ```
    find . -name '*.log' | /bin/grep unix | xargs cat | grep finished | sed 's/^\([^\[]*\)\[.*/\1/' | sort > finished
    ```

- List names of tests which started:

    ```
    find . -name '*.log' | /bin/grep unix | xargs cat | grep starting | sed 's/^\([^\[]*\)\[.*/\1/' | sort > started
    ```

- The diff of the above two files is the tests which started, but did not finish. Useful for debugging stuck tests. Usually requires multiple runs to figure out which test is the common one.

- Show test starting and ending times for each test, on adjacent lines:

    ```
    find . -name '*.log' | /bin/grep unix | xargs cat | sort | less
    ```

PR #716, opened 2019-09-23
Fix #695 - add more Linux-specific madvise() options

### LINKED ISSUE #695: Add the missing flags for madvise

XVilka:
Since https://github.com/ocsigen/lwt/pull/696 was closed, GitHub doesn't update closed pull requests.
I added the detection for two most important `madvise()` Linux-only options, but the question how to define the `type advice` in `src/unix/lwt_bytes.ml` conditionally? It's not a preprocessor option, right?

### DIFF
--- a/src/unix/unix_c/unix_madvise.c
+++ b/src/unix/unix_c/unix_madvise.c
@@ -14,6 +14,16 @@
 
 static int advise_table[] = {
     MADV_NORMAL, MADV_RANDOM, MADV_SEQUENTIAL, MADV_WILLNEED, MADV_DONTNEED,
+#if defined(MADV_MERGEABLE) && defined(MADV_UNMERGEABLE)
+	MADV_MERGEABLE, MADV_UNMERGEABLE,
+#else
+	0, 0,
+#endif
+#if defined(MADV_HUGEPAGE) && defined(MADV_NOHUGEPAGE)
+	MADV_HUGEPAGE, MADV_NOHUGEPAGE,
+#else
+	0, 0,
+#endif
 };
 
 CAMLprim value lwt_unix_madvise(value val_buffer, value val_offset,
--- a/src/unix/unix_c/unix_madvise.c
+++ b/src/unix/unix_c/unix_madvise.c
@@ -14,15 +14,25 @@
 
 static int advise_table[] = {
     MADV_NORMAL, MADV_RANDOM, MADV_SEQUENTIAL, MADV_WILLNEED, MADV_DONTNEED,
-#if defined(MADV_MERGEABLE) && defined(MADV_UNMERGEABLE)
-	MADV_MERGEABLE, MADV_UNMERGEABLE,
+#if defined(MADV_MERGEABLE)
+	MADV_MERGEABLE,
 #else
-	0, 0,
+	0,
 #endif
-#if defined(MADV_HUGEPAGE) && defined(MADV_NOHUGEPAGE)
-	MADV_HUGEPAGE, MADV_NOHUGEPAGE,
+#if defined(MADV_UNMERGEABLE)
+	MADV_UNMERGEABLE,
 #else
-	0, 0,
+	0,
+#endif
+#if defined(MADV_HUGEPAGE)
+	MADV_HUGEPAGE,
+#else
+	0,
+#endif
+#if defined(MADV_NOHUGEPAGE)
+	MADV_NOHUGEPAGE,
+#else
+	0,
 #endif
 };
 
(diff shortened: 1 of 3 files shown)


aantron:
Use `#ifdef` on the C side in `advise_table` to define constants that are not available as zero.

XVilka:
@aantron done. Rebased and cleaned up.

src/unix/config/discover.ml
```diff
@@ -705,6 +705,44 @@ struct
...
+        int main()
+        {
+        #if !defined(MADV_HUGEPAGE) && !defined(MADV_NOHUGEPAGE)
+        #error \"not available\"
+        #else
+            return 0;
+        #endif
+        }
+      |}
+  }
```
aantron: Given you don't use `HAVE_MADVISE_HUGEPAGE` or `HAVE_MADVISE_MERGEABLE` in the `#fidef`s in C, you probably don't need to detect them in `discover.ml` at all.

If you do decide that you need detection in `discover.ml`, I'd ask that, since the code between these is almost the same, you make one function for all the flags, and call it once for each flag to detect. See `nanosecond_stat` detection code below.

src/unix/unix_c/unix_madvise.c
```diff
@@ -14,6 +14,16 @@
 
 static int advise_table[] = {
     MADV_NORMAL, MADV_RANDOM, MADV_SEQUENTIAL, MADV_WILLNEED, MADV_DONTNEED,
+#if defined(MADV_MERGEABLE) && defined(MADV_UNMERGEABLE)
```
aantron: I suggest to define each constant independently, i.e. four `#ifdef`s without the `&&`.

XVilka: @aantron they were introduced together and have no meaning without each other. So they should be presented or both ignored.

src/unix/config/discover.ml
XVilka: I will remove detection then, it's not really usable outside `madvise()` calls.

XVilka:
Removed the discovery.

src/unix/unix_c/unix_madvise.c
aantron: If that's the case, then Lwt checking for both is redundant, since they will always either both be present or both be absent. On the other hand, if, for any reason, some platform ever has only one flag, then Lwt won't be able to expose the behavior of that platform, because of this check.

I suggest we pass through the platform's own flags as transparently as possible, so that we rely on the platform's logic fully.

aantron:
@XVilka could you avoid force-pushing? It can make review more complicated. I will squash-merge anyway, so it's not a problem if there are several commits :)

XVilka:
Ok.

XVilka:
Is anything required of me? Because it's finished I think.

aantron:
I'd ask to untangle the flags, see the last comment in the thread: https://github.com/ocsigen/lwt/pull/716#discussion_r327032653

I can also do it myself, if you wish.

src/unix/unix_c/unix_madvise.c
XVilka: Done.

aantron:
Thanks! I'll release this once some other recent issues are also resolved (see [milestone](https://github.com/ocsigen/lwt/milestone/16)).

PR #733, opened 2019-10-06
Deprecate let%lwt at the module item level

### LINKED ISSUE #728: Deprecate let%lwt structure items (they expand to Lwt_main.run)
See (1) here: https://github.com/ocsigen/lwt/issues/715#issuecomment-536333948

aantron:
Using `let%lwt` as a module item now produces a warning like this:

```
File "test/ppx/main.ml", line 162, characters 9-41:
162 |          let%lwt result = Lwt.return_true
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning 22: let%lwt should not be used at the module item level.
Replace let%lwt x = e by let x = Lwt_main.run (e)
```

Resolves #728.

<br/>

I completely removed support for `let%lwt ... and ...`, because it was broken. This line

https://github.com/ocsigen/lwt/blob/18e990e5b8875e654c165e2fdfa834340f36108d/src/ppx/ppx_lwt.ml#L140

was accidentally inserting the identifier `gen_exp` into the output AST, because it is missing an `[%e ...]` antiquotation, resulting in errors like

```
File "run_2.ml", line 1, characters 0-49:
Error: Unbound value gen_exp
```

This means that nobody has ever used `let%lwt ... and ...` successfully.

I removed the warning that we were trying to print on `let%lwt rec ... and ...`, because, in this case, the warning seems to be ignored, and the output is simply

```
File "run_2.ml", line 1, characters 4-7:
Error: Uninterpreted extension 'lwt'.
```

whether we try to add the warning or not. Given that we will already be discouraging `let%lwt` at structure item level without `and`, I think it's fine to have just this uglier message.

The result of these two changes is that we never try to interpret `let%lwt ... and ...` at all, simplifying the PPX a bit.

### DIFF
--- a/src/ppx/dune
+++ b/src/ppx/dune
@@ -6,5 +6,5 @@
  (libraries compiler-libs.common ocaml-migrate-parsetree ppx_tools_versioned)
  (ppx_runtime_libraries lwt)
  (kind ppx_rewriter)
- (preprocess (pps ppx_tools_versioned.metaquot_409))
+ (preprocess (pps ppx_tools_versioned.metaquot_409 bisect_ppx --conditional))
  (flags (:standard -w +A-4)))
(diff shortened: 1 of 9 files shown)


aantron:
(And I checked the blame about the `gen_exp` issue; the bug has been in the PPX since the beginning.)

PR #735, opened 2019-10-10
`deprected` -> `deprecated`

anmonteiro:


aantron:
Argh, that's embarrassing (for me!). Thanks, I'll add the warning suppression at the use site that I originally intended to add, but forgot because of this typo, and merge.

Issue #747, opened 2019-11-26
lwt_unix linker flags cause linker warnings


oandrieu:
The `unix_c_library_flags.sexp` flags as determined by `discover.exe` cause unseemly linker warnings on my system whenever I link in `lwt_unix` :
```
/usr/bin/ld: skipping incompatible /usr/lib/libpthread.so when searching for -lpthread
/usr/bin/ld: skipping incompatible /usr/lib/libpthread.so when searching for -lpthread
/usr/bin/ld: skipping incompatible /usr/lib/libm.so when searching for -lm
/usr/bin/ld: skipping incompatible /usr/lib/libdl.so when searching for -ldl
/usr/bin/ld: skipping incompatible /usr/lib/libc.so when searching for -lc
```

Indeed, `unix_c_library_flags.sexp`  is `("-lev" "-L/usr/lib" "-lpthread")`, but on Fedora/Redhats system `/usr/lib` contains 32-bits libraries whereas the system libraries directory is `/usr/lib64`.

aantron:
Fixed by #748.

PR #748, opened 2019-12-04
some discover.ml fixes

oandrieu:
addresses #747

### DIFF
--- a/src/unix/config/discover.ml
+++ b/src/unix/config/discover.ml
@@ -466,8 +466,13 @@ struct
           if !Arguments.android_target = Some true then
             no
           else begin
-            C_library_flags.detect context ~library:"pthread";
-            compiles context code
+            match compiles context code ~link_flags:["-lpthread"] with
+            | Some true ->
+              C_library_flags.add_link_flags ["-lpthread"];
+              Some true
+            | _ ->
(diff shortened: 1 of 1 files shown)


aantron:
IIRC, on at least one platform (Androd), linking fails when `-lpthread` is passed, but the symbols needed by Lwt are still available in the standard libraries chosen by the linker, so that's why we try to link without any flags first.

We could special-case this for Android only. I'll look through the history of Lwt in git and GitHub for any other platforms on which this could be important.

aantron:
To summarize my understanding at the moment, we have some platforms on which `-lpthread` *must not* be passed, and other platforms or setups on which `-lpthread` *should* be passed to avoid `discover.ml` searching for libraries and finding the wrong ones.

After this PR, I'll add a comment to this part of `discover.ml` about all the conditions.

aantron:
I believe everything will be accounted for if you move the new attempt with `-lpthread` to the `else` case currently on line 470 (in this PR), and do it before falling back to the current call to `detect` that is there.

oandrieu:
Ah yes thanks, indeed I suspected that `pthread` had a special status on some platforms.

aantron:
Thanks! And sorry about the delay on the original issue; I was away and only am now draining the backlog.

PR #750, opened 2019-12-06
Fix dead links

imbsky:


aantron:
Thanks!

imbsky:
That was too fast! Thank you, too!

PR #752, opened 2019-12-14
mock wait4 on android using waitpid like on win32

EduardoRFS:
Right now Lwt_process fails when using Android NDK because `wait4` isn't available on every Android device. To solve it this PR implements the same behavior as on win32 it just uses `waitpid` and fills the `rusage` with a mock.

This is important because esy does use Lwt_process and it fails without this patch to work when using an Android device


aantron:
Thanks!

Issue #680, opened 2019-03-27
Add esy + libev to the CI build matrix


anmonteiro:
Would make catching regressions like #679 easily

aantron:
Thanks! (prompted by #679)

aantron:
@anmonteiro, how does `package.json` typically look for an esy project that's depending on Lwt and libev?

PR #755, opened 2019-12-18
Fix esy libev detection

anmonteiro:
I'm gonna look separately into getting a test for this as suggested in https://github.com/ocsigen/lwt/issues/680

### DIFF
--- a/src/unix/config/discover.ml
+++ b/src/unix/config/discover.ml
@@ -397,8 +397,9 @@ struct
         in
         match result.stdout, result.exit_code with
         | "true\n", 0 -> true
-        | _, 0 -> false
-        | _, _ -> raise Not_found
+        | "false\n", 0 -> false
+        | _, _ ->
+         failwith "OPAM exited with an error code, or isn't even installed."
       in
 
       let detect_esy_wants_libev () =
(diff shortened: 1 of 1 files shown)


src/unix/config/discover.ml
```diff
@@ -395,9 +395,10 @@ struct
           Configurator.Process.run context
             "opam" ["config"; "var"; "conf-libev:installed"]
         in
-        match result.stdout with
-        | "true\n" -> true
-        | _ -> false
+        match result.stdout, result.exit_code with
+        | "true\n", 0 -> true
+        | _, 0 -> false
+        | _, _ -> raise Not_found
```
aantron: Could you do one of these two things:

1. Separate the test for the error code into an outer `match` statement, and raise `Failure` instead of `Not_found`, where the payload of `Failure` will explain what it is that failed, as a form of documentation in the source (even though the message will never be used at runtime).

2. (My preference) Keep this combined `match`, but change the second case from `_, 0` to `"false\n", 0`, which will make it clear that the `opam` command doesn't indicate that libev is not installed by returning a non-zero error code, but a non-zero error code means either a more serious failure in `opam`, or the `opam` command is missing. In this case, I'd still suggest to change the exception `Not_found` to `Failure` with a descriptive payload. I found `Not_found` to be misleading, because it first reads like the code means that it is specifically `libev` is not found by opam.

anmonteiro: For sure, I'll do 2.

aantron:
Thank you!

PR #756, opened 2019-12-18
Test esy gets libev

anmonteiro:
Initial stab at https://github.com/ocsigen/lwt/issues/680

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,125 @@
 language: generic
 
+scripts:
+  - &opam
+    before_install:
+      - '[ "$TRAVIS_EVENT_TYPE" != cron ] || rm -rf ~/.opam ./_opam ./_cache'
+
+    install:
+      - |
+        if [ "$LIBEV" != no ]
+        then
+          sudo apt-get update -qq
+          if [ "$MUSL" != yes ]
+          then
+            sudo apt-get install -qq libev-dev
+          else
+            sudo apt-get install musl-tools
+            wget http://dist.schmorp.de/libev/Attic/libev-4.27.tar.gz
+            tar xf libev-4.27.tar.gz
+            cd libev-4.27
+            CC=musl-gcc ./configure --prefix `pwd`/install
+            make
+            make install
+            sudo cp install/include/* /usr/include/x86_64-linux-musl
+            sudo cp install/lib/* /usr/lib/x86_64-linux-musl
+            cd ..
+          fi
+        fi
+
+      - OS=linux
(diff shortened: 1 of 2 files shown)


aantron:
It looks like you may want to split up the configs, which can be done like this:

https://github.com/aantron/bisect_ppx/blob/02dfb10188033a26d07d23480c2bc44a3a670357/.travis.yml#L129-L139

```yaml
matrix:
  include:
    - <<: *bucklescript
      os: osx
      env: BUCKLESCRIPT=YES SAVE_BINARIES=YES
    - <<: *bucklescript
      os: linux
      env: BUCKLESCRIPT=YES SAVE_BINARIES=YES
    - <<: *opam
      os: osx
      env: COMPILER=4.09.0
```

above that, there are:

https://github.com/aantron/bisect_ppx/blob/02dfb10188033a26d07d23480c2bc44a3a670357/.travis.yml#L4-L17

```yaml
scripts:
  - &bucklescript
    before_install:
      - |
        if [ "$TRAVIS_EVENT_TYPE" == cron ]
        then
          rm -rf ./test/bucklescript/node_modules
          rm -rf ./node_modules
          rm -rf ./_cache
        fi

    install:
      - npm --version
      - |
        if [ "$TRAVIS_BRANCH" != binaries ]
```

and

https://github.com/aantron/bisect_ppx/blob/02dfb10188033a26d07d23480c2bc44a3a670357/.travis.yml#L53-L63

```yaml
  - &opam
    before_install:
      - '[ "$TRAVIS_EVENT_TYPE" != cron ] || rm -rf ~/.opam ~/.esy ./_opam'


    install:
      - mkdir -p _cache


      - |
        case $TRAVIS_OS_NAME in
```

This uses YAML references to splice in different `before_install:`, `install:`, etc., stages into the different matrix rows, according to the build system in each row.

.travis.yml
```diff
@@ -1,25 +1,125 @@
...
       env: COMPILER=4.03.0
-    - os: linux
+    - <<: *opam
+      os: linux
       env: COMPILER=4.02.3 PACKAGING=yes
+    - <<: *esy
+      os: linux
+      env: ESY=yes
 
   allow_failures:
```
aantron: Could you add the esy row to `allow_failures`? Rather than treating these as allowed to fail, I use `allow_failures` for rows that are less likely to fail, given the main rows succeed. Together with `fast_finish`, this gives a reasonably accurate answer from CI sooner than if CI had to wait for all the rows. Rows in `allow_failures` still trigger emails, and I check them manually from time to time and before each release, so we won't miss any regression.

aantron:
Thanks!

PR #757, opened 2019-12-20
Don't extend library flags if the library isn't found

anmonteiro:
This fixes an issue where after #755 Esy would _only_ work if libev was
available in the sandbox.


aantron:
Thanks!

PR #769, opened 2020-03-05
Reduce the number of system calls for accept

chambart:
Use `accept4` instead of `accept` to avoid 2 `fcntl` system calls

`accept4` takes flags to tell to create the file descriptor in non blocking mode. This way we don't need to set it later with `Unix.set_nonblock.`

Also, since we know that all the calls to accept will be on non blocking sockets we can avoid releasing the runtime lock.

### DIFF
--- a/src/unix/config/discover.ml
+++ b/src/unix/config/discover.ml
@@ -808,6 +808,24 @@ struct
         }
       |}
   }
+
+  let () = feature {
+    pretty_name = "accept4";
+    macro_name = "HAVE_ACCEPT4";
+    detect = fun context ->
+      skip_if_windows context @@ fun () ->
+      compiles context {|
+        #define _GNU_SOURCE
+        #include <sys/socket.h>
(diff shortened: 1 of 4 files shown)


aantron:
Thanks! I'll merge this shortly.

Just a comment:

> Also, since we know that all the calls to accept will be on non blocking sockets we can avoid releasing the runtime lock.

The fd might be blocking, as a user can call `Lwt_unix.set_blocking listen_fd true` (though I know of no good reason to do so).

However, the current implementation in Lwt seems to assume that `Unix.accept` on a blocking listening socket will not block anyway, as Lwt will only call `Unix.accept` if the listening socket becomes readable (i.e., has a connection pending). I don't know if that's correct &mdash; perhaps the listening socket can become readable and then `Unix.accept` still blocks. However, if that's possible, this PR will be about equally as buggy as Lwt currently is here, so I think we can ignore this for now :)

PR #775, opened 2020-04-17
Define let*, and*

code-ghalib:
Proposal to add a syntax module to both `Lwt` and `Lwt_result` now that `let*` and `and*` are available in vanilla OCaml without ppx.

I've made an arbitrary choice while defining `Lwt_result.both` that works for most cases, because I cannot think of a solution that is both general and easy to use. I could avoid the arbitrary choice like so:
```ocaml
val both : ('e -> 'e -> 'e) -> ('a, 'e) Result.t Lwt.t -> ('b, 'e) -> Result.t Lwt.t

let both combine r1 r2 =
  Lwt.map (function
    | Ok p1, Ok p2 -> Ok (p1,p2)
    | Error e1, Error e2 -> Error (combine e1 e2)
    | Error e,_ | _,Error e -> Error e) (Lwt.both a b)

(* In case of string, we would say *)
let (and*) = Lwt_result.both (^)
```

But then defining the syntax module would probably involve a functor which is not that nice to use.

I'm open to suggestions - the main motivation is to not have to redefine these all the time for being able to use monadic syntax.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,1 +1,1 @@
-(lang dune 1.1)
+(lang dune 1.8)
--- a/lwt.opam
+++ b/lwt.opam
@@ -23,6 +23,7 @@ depends: [
   "dune-configurator"
   "mmap" {>= "1.1.0"} # mmap is needed as long as Lwt supports OCaml < 4.06.0.
   "ocaml" {>= "4.02.0"}
+  ("ocaml" {>= "4.08.0"} | "ocaml-syntax-shims")
   "ocplib-endian"
   "result" # result is needed as long as Lwt supports OCaml 4.02.
   "seq" # seq is needed as long as Lwt supports OCaml < 4.07.0.
--- a/src/core/dune
+++ b/src/core/dune
@@ -1,13 +1,10 @@
 (* -*- tuareg -*- *)
 
 let preprocess =
-  let res =
-    match Sys.getenv "BISECT_ENABLE" with
-    | "yes" -> "(preprocess (pps bisect_ppx))"
-    | _ -> ""
-    | exception _ -> ""
-  in
-  res^"\n(preprocess (future_syntax))"
+   match Sys.getenv "BISECT_ENABLE" with
+   | "yes" -> "(preprocess (pps bisect_ppx) (future_syntax))"
+   | _ -> "(preprocess (future_syntax))"
+   | exception _ -> "(preprocess (future_syntax))"
 
 let () = Jbuild_plugin.V1.send @@ {|
 
--- a/src/core/dune
+++ b/src/core/dune
@@ -1,10 +1,20 @@
 (* -*- tuareg -*- *)
 
 let preprocess =
-   match Sys.getenv "BISECT_ENABLE" with
-   | "yes" -> "(preprocess (pps bisect_ppx) (future_syntax))"
-   | _ -> "(preprocess (future_syntax))"
-   | exception _ -> "(preprocess (future_syntax))"
+  let older =
+    Scanf.sscanf
+      Sys.ocaml_version
+      "%d.%2d.%s"
+      (fun maj min _ -> maj < 4 || (maj = 4 && min < 8))
+  in
+  let bisect =
+    match Sys.getenv "BISECT_ENABLE" with
+    | "yes" -> "(preprocess (pps bisect_ppx))"
+    | _ -> ""
+    | exception _ -> ""
+  in
+  let future_syntax = "(preprocess (future_syntax))" in
+  if older && bisect = "" then future_syntax else bisect
 
 let () = Jbuild_plugin.V1.send @@ {|
 
--- a/src/core/dune
+++ b/src/core/dune
@@ -1,13 +1,13 @@
 (* -*- tuareg -*- *)
 
 let preprocess =
-   let res =
-	   match Sys.getenv "BISECT_ENABLE" with
-     | "yes" -> "(preprocess (pps bisect_ppx))"
-     | _ -> ""
-     | exception _ -> ""
-	 in
-	 res^"\n(preprocess (future_syntax))"
+  let res =
+    match Sys.getenv "BISECT_ENABLE" with
+    | "yes" -> "(preprocess (pps bisect_ppx))"
+    | _ -> ""
+    | exception _ -> ""
+  in
+  res^"\n(preprocess (future_syntax))"
 
 let () = Jbuild_plugin.V1.send @@ {|
 
--- a/src/core/dune
+++ b/src/core/dune
@@ -1,10 +1,13 @@
 (* -*- tuareg -*- *)
 
 let preprocess =
-   match Sys.getenv "BISECT_ENABLE" with
-   | "yes" -> "(preprocess (pps bisect_ppx))"
-   | _ -> ""
-   | exception _ -> ""
+   let res =
+	   match Sys.getenv "BISECT_ENABLE" with
+     | "yes" -> "(preprocess (pps bisect_ppx))"
+     | _ -> ""
+     | exception _ -> ""
+	 in
+	 res^"\n(preprocess (future_syntax))"
 
 let () = Jbuild_plugin.V1.send @@ {|
 
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -3116,6 +3116,11 @@ struct
 end
 include Infix
 
+module Syntax =
+struct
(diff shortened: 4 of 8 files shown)


anuragsoni:
Lwt supports ocaml versions starting from 4.02, so these new operators will fail on versions older than 4.08. I think this will need some dune machinery to ensure that this new syntax only gets exposed on ocaml >= 4.08. One example of this that i've seen in the ecosystem is the containers library that supports the new operators on ocaml > 4.08 https://github.com/c-cube/ocaml-containers/commit/bf0227d4046713384aee4e454da5244fedf8c91d

aantron:
I haven't had a chance to look at the PR in detail yet, but the `let*` syntax support on < 4.08 can be done like this:

1. https://github.com/aantron/luv/blob/0eae7f30ef99157bda77c62e0cb82169410de583/luv.opam#L20

2. https://github.com/aantron/luv/blob/0eae7f30ef99157bda77c62e0cb82169410de583/src/syntax/dune#L4

aantron:
I think the choice for `Lwt_result.both` is fine, given that the error type is unconstrained, and `Lwt_result.Syntax` is meant for use with `open`. If someone would like a safer `both` that combines the errors in some useful way, they would have to define `(and*)` anyway.

A variant of `Lwt_result.both` that knew something about the error type would be able to solve this more "intelligently." I guess this is an advantage of something like `Or_error`.

I'll port to < 4.08, add a couple tests, and merge.

aantron:
Actually, a safer choice is not to define `Lwt.result.(and*)`. Since the purpose of `Lwt_result` is explicit error handling, we shouldn't mislead users into dropping errors silently.

raphael-proust:
> A variant of `Lwt_result.both` that knew something about the error type would be able to solve this more "intelligently." I guess this is an advantage of something like `Or_error`.

I also think that Lwt should just provide Lwt (i.e., promise) monad-like combinators. Combining monads is its own can of worm and each project might want to do it on its own term. It's relatively easy for a project to write `promise.ml` that basically exports the Lwt primitives and wraps result-promise the way they want to use it.

code-ghalib:
> Actually, a safer choice is not to define `Lwt.result.(and*)`. Since the purpose of `Lwt_result` is explicit error handling, we shouldn't mislead users into dropping errors silently.

I can live with that:
```ocaml
let (and*) = Lwt_result.both
```

is not too annoying to duplicate across modules, and makes the choice *somewhat* more explicitly available for code reviewers. Although, this does limit the utility of having a `Syntax` module at all.  

Also, the error is not **dropped** per-se, the resulting promise is still wrapped in a `Result`.  The distinction between these two situations is dropped:
- *either* promise resulted in an error
- *both* promises resulted in an error  

This has practically not mattered to me personally, because in either case, I cannot continue with the remaining computation - it's a small debugging inconvenience to not know which of the two situations happened.

Not sure if you want to add the applicative syntax `let+`, `and+` as well. Personally, I have not used them in practice.  

> I'll port to < 4.08, add a couple tests, and merge.

That'd be great thanks! Not sure if this involves packaging it as a separate libraries:`lwt.syntax`, `lwt.result.syntax`. I will make an attempt to do this at my end too, just to get more familiar with `dune`.

aantron:
> I can live with that:

I had in mind not to add `Lwt.result.both`, either. But I'm very open to debating that. Do you often need such a function?



> The distinction between these two situations is dropped:
> 
> * _either_ promise resulted in an error
> * _both_ promises resulted in an error

Apart from the distinction, in the "both" case, one of the errors is also dropped. If we could guarantee that the error type has an operation `error -> error -> error` (like `^` as you suggested for strings), i.e. some kind of *logging*, we wouldn't have to drop one of the errors silently. That's what makes me suspect that `Lwt_result` as it is now is too underconstrained to have a reasonable `both` with the reasonable type signature, i.e. `Lwt_result`'s original design has missed a certain subtlety. It just seems like it would mislead users about the rigidity of error propagation.



> I will make an attempt to do this at my end too, just to get more familiar with `dune`.

Could you try it in this PR, then? It doesn't require anything fancy, I linked in a [prior comment](https://github.com/ocsigen/lwt/pull/775#issuecomment-615470084) to the only two lines you should have to add for this.

aantron:
Well, I guess `Lwt_result.both`'s error handling is not more questionable than value handling in `Lwt.pick`, so maybe it's okay. Perhaps a comment is enough to warn people. Still interested in opinions on this :)

code-ghalib:
> I had in mind not to add `Lwt.result.both`, either. But I'm very open to debating that. Do you often need such a function?

Yes, in practice, I'm mostly using `Lwt` to parallelize i/o heavy stuff, and as a result, almost every promise is stateful and can can result in errors, so `Lwt_result.t` is used much more often than an unwrapped `Lwt.t`

> Apart from the distinction, in the "both" case, one of the errors is also dropped. If we could guarantee that the error type has an operation `error -> error -> error` (like `^` as you suggested for strings), i.e. some kind of _logging_, we wouldn't have to drop one of the errors silently. That's what makes me suspect that `Lwt_result` as it is now is too underconstrained to have a reasonable `both` with the reasonable type signature, i.e. `Lwt_result`'s original design has missed a certain subtlety. It just seems like it would mislead users about the rigidity of error propagation.

I agree it's not ideal. Not sure how it might be done better though. I'm not familiar enough with haskell to understand some of their monad transformers examples, but I'm sure the situation would have come up there before? Maybe we can seek out opinions on discourse?

> > I will make an attempt to do this at my end too, just to get more familiar with `dune`.
> 
> Could you try it in this PR, then? It doesn't require anything fancy, I linked in a [prior comment](https://github.com/ocsigen/lwt/pull/775#issuecomment-615470084) to the only two lines you should have to add for this.

Will do!

code-ghalib:
Looked for inspiration at `async`. I may be mis-reading this, but I think `async` is getting the sequential version of `both` for their `Deferred.Result.t`:
https://github.com/janestreet/async_kernel/blob/master/src/deferred_result.ml#L17
https://github.com/janestreet/base/blob/master/src/monad.ml#L48

That seems not so useful.  

If we were to go with a functorized solution, it would look something like this:

```ocaml
let _ =
  let open Lwt_result.Syntax.Make(struct type t = string let combine_err = (^) end) in
  let* p1 = thing1 ()
  and* p2 = thing2 ()
  and* p3 = thing3 () in
  (* do your thing *)
```

I would frankly not enjoy using/reading that. I vote to warn people with comments in the absence of a smarter proposal.

kohlivarun5:
Personally, I would be okay with using `both` that just picks the first error, since for most of my use cases, that is sufficient and cleaner/easier to use. 

If people do want a version with combine, that can be exposed as an additional functor, so clients who prefer to combine can use the functor

aantron:
> I think `async` is getting the sequential version of `both` for their `Deferred.Result.t`

Not in that code. The promises `a` and `b` are already constructed, and any underlying operations already running, by the time `both` is called. So no sequencing of the underlying operations is introduced by waiting first on `a` and then on `b`. Lwt is not able to do this because Lwt promises can be rejected with `exn`, so `Lwt.both` uses a more complicated definition to ensure commutativity:

https://github.com/ocsigen/lwt/blob/2412006f228622437f5625d6e71dcdd0ea5cbcb2/src/core/lwt.ml#L2574-L2582

Here, it relies on `Lwt.join`'s inner semantics to ensure the behavior is the same no matter which promise is written on the left and which on the right.

Actually, we probably want this property for `Lwt_result.both`, too. See https://github.com/ocsigen/lwt/pull/668#issuecomment-466606801 and the tests:

https://github.com/ocsigen/lwt/blob/2412006f228622437f5625d6e71dcdd0ea5cbcb2/test/core/test_lwt.ml#L1948-L2159

aantron:
If getting commutativity is too much of a pain, we can probably just leave it as a note in this issue. I think you should be able to get it by having `both` keep track of which promise resolved first, and if both promises resolve with `Error`, choosing to propagate the `Error` from the first promise to resolve.

aantron:
If you're having trouble with Bisect, you can drop support for it. I will work it back in later.

Alternatively, you can test the OCaml version and add `(future_syntax)` only if it is < 4.08, otherwise add Bisect (conditionally). We don't build with coverage on old versions.

aantron:
Thanks @code-ghalib! And to all for discussion!

aantron:
I'll release this in 5.3.0 this week.

aantron:
This is now out in Lwt 5.3.0.

PR #776, opened 2020-04-22
Add {let+,and+} applicative syntax

CraigFe:
A follow up to https://github.com/ocsigen/lwt/pull/775 proposing `Syntax.(( let+ ), ( and+ ))` operators for the `Lwt` and `Lwt_result` modules.

When working with Lwt, I find myself using `>|=` often enough that a switch to `open Lwt.Syntax` would be frustrating without an equivalent syntactic form. As a sample from the [`mirage/irmin`](https://github.com/mirage/irmin) codebase:

```bash
ᐅ git grep ">>=" | wc -l
1480

ᐅ git grep ">|=" | wc -l
347
```

The proposed `( and+ )` operations behave the same as `( and* )`: I find it convenient / aesthetic to use the corresponding `and<op>` for whichever `let<op>` I'm using. We've been using similar aliases in [OCurrent syntax](https://ocurrent.github.io/ocurrent/current/Current/Syntax/index.html).

### DIFF
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -1400,7 +1400,6 @@ end
 (** {3 Let syntax} *)
 module Syntax :
 sig
-  (**)
 
   (** {1 Monadic syntax} *)
 
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -1402,6 +1402,9 @@ module Syntax :
 sig
   val (let*) : 'a t -> ('a -> 'b t) -> 'b t
   val (and*) : 'a t -> 'b t -> ('a * 'b) t
+
+  val (let+) : 'a t -> ('a -> 'b) -> 'b t
+  val (and+) : 'a t -> 'b t -> ('a * 'b) t
 end
 
 
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -1400,11 +1400,23 @@ end
 (** {3 Let syntax} *)
 module Syntax :
 sig
+  (**)
+
+  (** {1 Monadic syntax} *)
+
   val (let*) : 'a t -> ('a -> 'b t) -> 'b t
+  (** Syntax for {!bind}. *)
+
   val (and*) : 'a t -> 'b t -> ('a * 'b) t
+  (** Syntax for {!both}. *)
+
+  (** {1 Applicative syntax} *)
 
   val (let+) : 'a t -> ('a -> 'b) -> 'b t
+  (** Syntax for {!map}. *)
+
   val (and+) : 'a t -> 'b t -> ('a * 'b) t
+  (** Syntax for {!both}. *)
 end
 
 
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -3120,6 +3120,9 @@ module Syntax =
 struct
   let (let*) = bind
   let (and*) = both
+
+  let (let+) x f = map f x
+  let (and+) = both
 end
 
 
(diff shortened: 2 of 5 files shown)


src/core/lwt.mli
```diff
@@ -1400,8 +1400,23 @@ end
 (** {3 Let syntax} *)
 module Syntax :
 sig
+  (**)
```
aantron, line 1403: Would a single blank line suffice here, instead of an empty comment?

CraigFe: Yep, I think it would. I default to the empty comments now because I'm used to OCamlformat killing those blank lines...

aantron:
Thanks! I'll merge this tomorrow and release 5.3.0 with it.

src/core/lwt.mli
aantron: Ah, good to know. Is that a current bug in OCamlFormat?

In any case, Lwt doesn't use it, at least not yet.

CraigFe: Whether or not it's a bug is somewhat up to interpretation. OCamlformat doesn't always preserve Odoc's view of the world (in instances where Odoc is using more than just the AST structure to determine placement). It only guarantees to preserve the AST of the input program (disregarding location information).

A similar case happens for "grouped" items:

```diff
 val foo : unit
+
 val bar : unit
 (** conventional OCamlformat won't let me group these! *)
```

(although it might be possible to find some combination of options that _will_ preserve that information)

aantron: Ok, I am taking this further off topic, but in neither case (and in no case at all, today) is odoc interpreting whitespace.

In the case of the first comment, it is the compiler's parser which associates the comment with either the module or a module item, and attaches the documentation attribute to different AST nodes according to the decision. If OCamlFormat is effectively moving the comment so that it affects that decision, it is indeed not preserving the AST, regardless of the involvement of odoc.

In the second case, that was a former interpretation of odoc, but odoc wasn't reading the whitespace or the locations. It was just interpreting sequences of items of the same kind, in which only the last had a documentation attribute, as a group. This is not analogous to the first case &mdash; it would be analogous if OCamlFormat changed the spacing in such a way, that the comment would no longer be associated with `bar` *by the compiler*.

CraigFe: My understanding is that the following module types have the same AST:
```ocaml
(** Outside *)
module type S = sig
  (** Inside *)
end

(** Outside *)
module type S = sig

  (** Inside *)
end
```
In both cases, the `ocaml.text` attribute is attached as a signature item:
```
[{pstr_desc =
   Pstr_modtype
    {pmtd_name = {txt = "S"};
     pmtd_type =
      Some
       {pmty_desc =
         Pmty_signature
          [{psig_desc =
             Psig_attribute
              {attr_name = {txt = "ocaml.text"};
               attr_payload =
                PStr
                 [{pstr_desc =
                    Pstr_eval
                     ({pexp_desc =
                        Pexp_constant (Pconst_string (" Foo ", None));
                       pexp_loc_stack = []; pexp_attributes = []},
                     [])}]}}];
        pmty_attributes = []};
     pmtd_attributes = []}}]
```

Indeed, [OCamlformat would crash](https://github.com/ocaml-ppx/ocamlformat/blob/master/lib/Translation_unit.ml#L319-L334) if this were the case, since it checks for AST equality using `Mirage_ast.Printast`.

I only vaguely remember the empty comment changed the Odoc rendered output (something to do with including a module type?), but if what you say is true: perhaps neither an empty doc comment nor an empty line makes any observable difference :slightly_smiling_face:

aantron:
This is now out in Lwt 5.3.0.

PR #785, opened 2020-05-15
Cancellable

### LINKED ISSUE #782: RFC / Feature request: `cancelable`
Whilst coding I stumbled upon a situation where I needed a promise to be reliably cancelable. As in `Lwt.cancel p` needed to cause `p` to be rejected with `Canceled` if `p` was in the `Sleep` state. I came up with the following:

```
let cancelable p =
   match Lwt.state p with
   | Return _ | Fail _ -> p
   | Sleep -> Lwt.pick [p; fst (Lwt.task ())]
```

1. (RFC) This does what I need, but: have I missed some corner cases?

2. (Feature request) This would live nicely along `protected` and `no_cancel` in the `Lwt` module. I'm open to a change of name because `cancelable` could also be the name of a predicate to test whether a promise is cancelable. Maybe `unprotected` (but it suggests that this cancels out `protected`) or `always_cancelable` (but it's not if the argument is resolved)?

-----------------------------------------

Considering `protected`, `no_cancel`, and `cancelable` suggests a fourth function: one that propagates the cancelation search backwards but doesn't propagate the cancelation error forwards. This completes the cross product of “carries cancelation backwards”/“doesn't” with “fails with `Canceled` when canceled”/“doesn't”.

Something along the lines of

```
let hide_cancelation p =
  Lwt.catch
    (fun () -> p)
    (function Lwt.Canceled -> fst @@ Lwt.wait () | e -> Lwt.fail e)
```

Although I can't find uses for it right now, and returning a never-resolving promise is a bit strange.

raphael-proust:
Fixes #782 

This PR provides a new `cancelable : 'a t -> 'a t` function such that `cancelable p` is a promise `p'` that is cancelable even if `p` isn't.

I introduced a new constructor in `how_to_cancel`. I'm not 100% sure it is necessary but I couldn't find a way to do without it.

### DIFF
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -1550,6 +1550,7 @@ sig
 
   val protected : 'a t -> 'a t
   val no_cancel : 'a t -> 'a t
+  val wrap_in_cancelable : 'a t -> 'a t
 end =
 struct
   let new_pending ~how_to_cancel =
@@ -1684,6 +1685,28 @@ struct
       in
       add_implicitly_removed_callback p_callbacks callback;
 
+      to_public_promise p'
+
+  let wrap_in_cancelable p =
+    let Internal p_internal = to_internal_promise p in
+    let p_underlying = underlying p_internal in
+    match p_underlying.state with
+    | Fulfilled _ -> p
+    | Rejected _ -> p
+    | Pending p_callbacks ->
+      let rec p' = lazy (
+        new_pending
+          ~how_to_cancel:(
+            Propagate_cancel_to_several [p_underlying ; Lazy.force p'])) in
+      let p' = Lazy.force p' in
+      let callback p_result =
+        let State_may_now_be_pending_proxy p' = may_now_be_proxy p' in
+        let p' = underlying p' in
+        let State_may_have_changed p' =
+          resolve ~allow_deferring:false p' p_result in
+        ignore p'
+      in
+      add_implicitly_removed_callback p_callbacks callback;
       to_public_promise p'
 end
 include Pending_promises
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -1550,7 +1550,6 @@ sig
 
   val protected : 'a t -> 'a t
   val no_cancel : 'a t -> 'a t
-  val wrap_in_cancelable : 'a t -> 'a t
 end =
 struct
   let new_pending ~how_to_cancel =
@@ -1685,28 +1684,6 @@ struct
       in
       add_implicitly_removed_callback p_callbacks callback;
 
-      to_public_promise p'
-
-  let wrap_in_cancelable p =
-    let Internal p_internal = to_internal_promise p in
-    let p_underlying = underlying p_internal in
-    match p_underlying.state with
-    | Fulfilled _ -> p
-    | Rejected _ -> p
-    | Pending p_callbacks ->
-      let rec p' = lazy (
-        new_pending
-          ~how_to_cancel:(
-            Propagate_cancel_to_several [p_underlying ; Lazy.force p'])) in
-      let p' = Lazy.force p' in
-      let callback p_result =
-        let State_may_now_be_pending_proxy p' = may_now_be_proxy p' in
-        let p' = underlying p' in
-        let State_may_have_changed p' =
-          resolve ~allow_deferring:false p' p_result in
-        ignore p'
-      in
-      add_implicitly_removed_callback p_callbacks callback;
       to_public_promise p'
 end
 include Pending_promises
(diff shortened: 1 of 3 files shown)


hcarty:
The concept makes sense, but it would be nice to give this a clearer name since it doesn't make a promise cancelable.  Maybe `wrap_in_cancelable`?

raphael-proust:
@hcarty I agree that the name is not really good as is. The type of the function hints that it wraps the promise (rather than modify the promise in place) but the name should also hint at it.

I like `wrap_in_cancelable`. I tend to like past participles for this kind of things – e.g., how Python has `sort` (for essentially `'a list -> unit`) and `sorted` (`'a list -> 'a list`). And that would mirror `protected` nicely too. But I can't find a good past participle.

Anyway, I'll rename it to something like `wrap_in_cancelable` if that's the best suggestion I get.

src/core/lwt.ml
```diff
@@ -455,10 +455,11 @@ struct
   and 'a resolved_state = ('a, underlying, resolved) state
 
   and how_to_cancel =
-    | Not_cancelable              :                           how_to_cancel
-    | Cancel_this_promise         :                           how_to_cancel
-    | Propagate_cancel_to_one     : (_, _, _) promise      -> how_to_cancel
-    | Propagate_cancel_to_several : (_, _, _) promise list -> how_to_cancel
+    | Not_cancelable                           :                           how_to_cancel
+    | Cancel_this_promise                      :                           how_to_cancel
+    | Cancel_this_promise_and_propagate_to_one : (_, _, _) promise      -> how_to_cancel
```
aantron, line 460: Why is it necessary to add a new constructor? Given you were able to implement this function outside Lwt, I imagine it should be possible to implement it inside without changing these constructors.

Note that the existing constructors have a certain "algebraic" nature to them, in that `Not_cancelable` and `Cancel_this_promise` correspond to zero cancelable "ancestors," and arise when entirely independent promises are introduced ex nihilo with `Lwt.task` and `Lwt.wait` (functions that take zero other promise arguments), `Propagate_cancel_to_one` corresponds to one cancelable "ancestor," and arises from sequential composition upon one existing promise, and `Propagate_cancel_to_several` corresponds to an arbitrary number of cancelable "ancestors," and arises from concurrent composition upon an arbitrary number of existing promises. So these existing constructors really are in some way basic, they are a kind of 0, 1, and *.

Note also that `protected` and `no_cancel` are implemented without introducing new constructors.

raphael-proust: Oh, I can do a `Propagate_cancel_to_several` and pass it “itself” I guess. I'll do that instead.

raphael-proust: I had to use `lazy` to fit in the recursion. But it seems to work. At least the tests written for it pass.

test/core/test_lwt.ml
```diff
@@ -2720,6 +2720,94 @@ let protected_tests = suite "protected" [
...
+    Lwt.return (Lwt.state p = Lwt.Sleep && Lwt.state p = Lwt.Fail Lwt.Canceled)
+  end;
+
+  test "pending(task), canceled, fulfilled" begin fun () ->
+    let p, r = Lwt.task () in
+    let p' = Lwt.protected p in
+    Lwt.cancel p';
+    Lwt.wakeup r "foo";
+    Lwt.return
+      (Lwt.state p = Lwt.Fail Lwt.Canceled && Lwt.state p' = Lwt.Fail Lwt.Canceled)
```
aantron, line 2780: What causes `p` to become canceled here? IIRC `protected` should prevent propagation of cancellation from `p'` to `p`.

test/core/test_lwt.ml
```diff
@@ -2720,6 +2720,94 @@ let protected_tests = suite "protected" [
...
+    let p3, r3 = Lwt.wait () in
+    let _ = Lwt.bind p3 (fun () -> p2) in
+    Lwt.wakeup r3 ();
+
+    (* It should now be possible to resolve p2 by resolving p1. *)
+    Lwt.wakeup r1 "foo";
+    Lwt.return (Lwt.state p2 = Lwt.Return "foo")
+  end;
+]
+let suites = suites @ [protected_tests]
```
aantron, line 2809: This should append `cancelable_tests `, not `protected_tests`.

test/core/test_lwt.ml
raphael-proust: Yes, this should be `Return "foo"`. I changed it now.

Also, it doesn't actually test the added function at all. And several of those tests have the same issue. I'll fix that too.

test/core/test_lwt.ml
raphael-proust: Fixed!

raphael-proust:
So I actually made it point to the right test (thanks for spotting the error) and there's a bunch of failures where tests raise `CamlinternalLazy.Undefined`. I'll investigate the feature and maybe I'll have to go without Lazy and reintroduce the `Cancel_this_promise_and_propagate_to_one`.

aantron:
I think, in the worst case, you should be able to define the function using higher-level helpers like `on_cancel`. I could be wrong. But if that's correct, I suggest to go that route rather than introducing a new internal constructor.

raphael-proust:
I implemented it with `on_cancel` for backwards cancellation propagation and `on_any` for forward resolution propagation.

It doesn't fit with the `protected`/`no_cancel` syntactically because of the order the primitives are declared in, but it works.

aantron:
Thanks!

PR #786, opened 2020-05-22
Doc: Clarify that sleep() and yield()

bikallem:
Clarify that sleep and yield resolves a thread/promise rather than terminate it.


raphael-proust:
Good spot. You can also change `thread` for `promise` and `suspending` for `pending`:

```
[yeild ()] is a pending promise that resolves …
```

Note that another open PR also updates the documentation of `yield`: https://github.com/ocsigen/lwt/pull/784/files#diff-6b263cc9435dfbe9ea99fced61920cbbR46-R52 but that merging/rebasing should be easy either way.

bikallem:
@raphael-proust Updated PR with review comments.

aantron:
Thanks!

PR #788, opened 2020-05-25
Lwt_unix: more documentation fixes

raphael-proust:
A few more tweaks to remove “thread” and other associated terminology


aantron:
Thanks!

PR #790, opened 2020-05-31
String_val returns const char *

MisterDA:
> `String_val(v)` returns a pointer to the first byte of the string
> `v`, with type `char *` or, when OCaml is configured with
> `-force-safe-string`, with type `const char *`.

https://caml.inria.fr/pub/docs/manual-ocaml/intfc.html#ss:c-block-access

Safe strings have been the default since OCaml 4.06.

https://caml.inria.fr/pub/docs/manual-ocaml/libref/String.html

The function CreateProcess has two parameters that are non-const:

    LPSTR  lpCommandLine
    LPVOID lpEnvironment

https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa

Passing a const string triggers warnings. One can cast away the const
but that does not seem safe, or duplicate the string. Here, I’m
duplicating the string.

### DIFF
--- a/src/unix/lwt_process_stubs.c
+++ b/src/unix/lwt_process_stubs.c
@@ -36,8 +36,21 @@ CAMLprim value lwt_process_create_process(value prog, value cmdline, value env,
   CAMLparam5(prog, cmdline, env, cwd, fds);
   CAMLlocal1(result);
 
+  size_t count;
+  LPSTR commandLine;
+  LPVOID environment = NULL;
   STARTUPINFO si;
   PROCESS_INFORMATION pi;
+
+  count = caml_string_length(cmdline);
+  commandLine = lwt_unix_malloc(count);
+  memcpy(&commandLine, String_val(cmdline), count);
+
+  if (Is_block(env)) {
+    count = caml_string_length(Field(env, 0));
+    environment = lwt_unix_malloc(count);
+    memcpy(&environment, String_val(Field(env, 0)), count);
+  }
 
   ZeroMemory(&si, sizeof(si));
   ZeroMemory(&pi, sizeof(pi));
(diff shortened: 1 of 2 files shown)


aantron:
MSDN [has](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa):

> The Unicode version of this function, **CreateProcessW**, can modify the contents of this string. Therefore, this parameter cannot be a pointer to read-only memory (such as a const variable or a literal string). If this parameter is a constant string, the function may cause an access violation.

Given the typical representation of strings in OCaml, do we ever end up in an environment where `CreateProcess` is `CreateProcessW`? I guess this is a question for anyone watching the repo. @dra27?

MSDN is silent on whether either `CreateProcess` function can write into `lpEnvironment`.

dra27:
Note that since 4.06, OCaml itself always uses Unicode versions of functions. It's not related here, but unless you use Windows '95/'98/ME, the ANSI versions of functions are stubs which convert their arguments to Unicode and call the "wide" version. As it happens, that's the explanation for why this only affects `CreateProcessW` (see [blog post on this here](https://devblogs.microsoft.com/oldnewthing/20090601-00/?p=18083)).

Why not upgrade the LWT Unix stubs to use Unicode on Windows? That way you'd be copying anyway because you'll need to use `caml_stat_strdup_to_os` on both strings (cf. https://github.com/ocaml/ocaml/blob/trunk/otherlibs/win32unix/createprocess.c#L100)

aantron:
Thanks @dra27.

I agree. We should aim to change Lwt to use the -`W` functions on Windows, as you suggest, in the long term (if this isn't superseded by libuv conversion or some other such work).

For right now, I think the best thing to do is cast, and add a comment that this is only valid because we are calling `CreateProcessA`, perhaps with links to this PR, the blog post, and/or the MSDN entry. Or we can change this specific code to call `CreateProcessW`, as in the code @dra27 linked.

Issue #798, opened 2020-08-25
Segfault on Lwt_unix.tcsetattr with bad parameters


GreenArchon:
Hello,

The following code

```ocaml
let run_lwt () =
  let%lwt fd = Lwt_unix.openfile "test1" [O_RDWR;O_NONBLOCK] 0o644 in
  let%lwt ta = Lwt_unix.tcgetattr fd in
  let ta' = { ta with c_obaud = 115201; c_ibaud = 115201 } in
  let%lwt () =
    Lwt.try_bind
      (fun () -> Lwt_unix.tcsetattr fd TCSANOW ta')
      Lwt.return
      (fun exn -> Lwt_io.printf "LWT Got %s\n%!" (Printexc.to_string exn))
  in
  Lwt_io.printf "Done LWT\n%!"
```
, accompanied by the prior creation of a pseudo terminal in the background in the same directory, eg. with 

```bash
socat -d -d pty,raw,echo=0,link=test1 pty,raw,echo=0,link=test2
```

segfaults, with the following output:

```
Fatal error: exception Unix.Unix_error(Unix.EINVAL, "tcsetattr", "")
Raised by primitive operation at file "src/unix/lwt_engine.ml", line 403, characters 26-60
Called from file "src/unix/lwt_engine.ml", line 344, characters 8-19
Called from file "src/unix/lwt_main.ml", line 37, characters 6-49
Called from file "src/unix/lwt_main.ml", line 114, characters 8-13
Re-raised at file "src/unix/lwt_main.ml", line 120, characters 4-13
Called from file "test_crash.ml", line 25, characters 2-28
[1]    13202 segmentation fault (core dumped)  OCAMLRUNPARAM=b _build/install/default/bin/test_crash
```

Note the 115201 baudrate is not a standard baudrate (115200 is, and setting it does not crash), so there's a reason for the failure - but I would expect it to call the exception handler and continue its merry way, not a hard crash.

The equivalent standard OCaml code:

```ocaml
let run_normal () =
  let fd = Unix.openfile "test1" [O_RDWR;O_NONBLOCK] 0o644 in
  let ta = Unix.tcgetattr fd in
  let ta' = { ta with c_obaud = 115201; c_ibaud = 115201 } in
  (try
    Unix.tcsetattr fd TCSANOW ta'
  with exn -> Printf.printf "UNIX: Got %s\n%!" (Printexc.to_string exn));
  Printf.printf "Done Unix\n%!"
```

[...]

PR #799, opened 2020-08-25
Lwt_unix.tcsetattr: Fix segfault on bad arguments (#798)

GreenArchon:
This MR fixes the segfaults I had in #798  .

A review would be appreciated!

Notably,

* I'm not sure about setting errno in userspace..
* decode_terminal_status() is basically argument validation - would it be better placed in the job creation or do we not care?
* Is the premise that exceptions shouldn't be thrown from worker code correct at all?

### DIFF
--- a/src/unix/unix_c/unix_tcsetattr_job.c
+++ b/src/unix/unix_c/unix_tcsetattr_job.c
@@ -34,8 +34,12 @@ static void worker_tcsetattr(struct job_tcsetattr *job)
         job->result = result;
         job->error_code = errno;
     } else {
-        decode_terminal_status(&termios, &(job->termios[0]));
-        job->result = tcsetattr(job->fd, job->when, &termios);
+        int result_decode = decode_terminal_status(&termios, &(job->termios[0]));
+        if (result_decode != 0) {
+            job->result = -1;
+        } else {
+            job->result = tcsetattr(job->fd, job->when, &termios);
+        }
         job->error_code = errno;
     }
 }
--- a/src/unix/unix_c/unix_termios_conversion.c
+++ b/src/unix/unix_c/unix_termios_conversion.c
@@ -207,7 +207,7 @@ void encode_terminal_status(struct termios *terminal_status, value *dst)
     }
 }
 
-void decode_terminal_status(struct termios *terminal_status, value *src)
+int decode_terminal_status(struct termios *terminal_status, value *src)
 {
     long *pc;
     int i;
@@ -232,7 +232,8 @@ void decode_terminal_status(struct termios *terminal_status, value *src)
                 if (i >= 0 && i < num) {
                     *dst = (*dst & ~msk) | pc[i];
                 } else {
-                    unix_error(EINVAL, "tcsetattr", Nothing);
+                    errno = EINVAL;
+                    return EINVAL;
                 }
                 pc += num;
                 break;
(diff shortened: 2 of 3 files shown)


raphael-proust:
Hi,

Thank you for the error report and the quick patch. I'll have a look at it in more detail within a few days.

raphael-proust:
I had a look. I haven't reproduced the error yet because I don't have a good setup to do so (no `socat` to start with). I'll work on that.

In the mean time, I notice that your changes affect some existing error handling that seems somewhat strange to begin with. Specifically, the use of `uerror` vs `unix_error` looks suspicious to me. I'll look into it more. In the meantime, can you try to reproduce the segfault when applying the following diff on top of master:

```
$ git diff
diff --git a/src/unix/unix_c/unix_termios_conversion.c b/src/unix/unix_c/unix_termios_conversion.c
index 7b61bcf8f..b16a118bc 100644
--- a/src/unix/unix_c/unix_termios_conversion.c
+++ b/src/unix/unix_c/unix_termios_conversion.c
@@ -253,7 +253,7 @@ void decode_terminal_status(struct termios *terminal_status, value *src)
                                                   speedtable[i].speed);
                                 break;
                         }
-                        if (res == -1) uerror("tcsetattr", Nothing);
+                        if (res == -1) unix_error(EINVAL, "tcsetattr", Nothing);
                         goto ok;
                     }
                 }
```

GreenArchon:
Hi @raphael-proust !

Thanks for having a look. Your diff doesn't fix the problem and still causes a crash. Are exceptions supposed to be thrown from workers (if not for the whole crash, there's at least a memory leak involved by doing so)?

If you don't have socat or similar installed, you could have a look at opening, instead of the "test1" symlink, a numbered file in `/dev/pts/`, if there are any, or a real device if you have eg. an arduino you could plug in, then access via `/dev/ttyUSB*`. But that might start to be too complicated :)

raphael-proust:
Hi,

I have installed `socat` and made the other small necessary changes to my setup. I can confirm that your fix works on my machine as well.

For a better review, I'll ask people that are more familiar with C-bindings in general because it isn't my area of expertise.

src/unix/unix_c/unix_tcsetattr_job.c
```diff
@@ -34,8 +34,12 @@ static void worker_tcsetattr(struct job_tcsetattr *job)
         job->result = result;
         job->error_code = errno;
     } else {
-        decode_terminal_status(&termios, &(job->termios[0]));
-        job->result = tcsetattr(job->fd, job->when, &termios);
+        int result_decode = decode_terminal_status(&termios, &(job->termios[0]));
+        if (result_decode < 0) {
```
aantron, line 38: I suggest to change this comparison to `!= 0`, and return `EINVAL` rather than `-EINVAL` from `decode_terminal_status`. Set `job->result` to `-1` in the then-branch, as we don't use that value beyond checking that it's negative. We take the error result from `error_code`. `-EINVAL` appears fragile.

aantron (changes requested):

src/unix/unix_c/unix_tcsetattr_job.c
GreenArchon: Sure, that makes sense. Amended the commit!

aantron (approved):

PR #800, opened 2020-08-26
lwt_sequence: update some function names in docs

Sudha247:
This patch fixes some function names in `lwt_sequence.mli` file to match their corresponding functions.

Note: I see that `Lwt_sequence` is deprecated. Since it is used internally in Lwt, people looking at Lwt source code might be referring this doc (like myself), so I thought it might be worth updating it. :)


raphael-proust:
Looks good! Thanks for the fix!

PR #807, opened 2020-09-14
Switch to ppxlib

pveber:
This is my shot at converting the `lwt_ppx` to `ppxlib` instead of `ppx_tools_versioned`, which [is going to be deprecated](https://discuss.ocaml.org/t/ocaml-migrate-parsetree-2-0-0/5991). This new version passes all tests performed after a `dune runtest` invocation, I only see this warning:
```
$dune runtest
File "test/ppx/main.ml", line 9, characters 0-46:
9 | let%lwt structure_let_result = Lwt.return true
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning 22: let%lwt should not be used at the module item level.
Replace let%lwt x = e by let x = Lwt_main.run (e)
```
which is I believe is expected.

I tried to stay as close as possible to the original code, by sticking to a whole-module transformation. An alternative might have been to resort to `ppxlib`'s context-free rules which are preferable but maybe not expressive enough here, I'm not sure.

One recurrent difference with the original code is that I think there is no equivalent of the `[@metaloc ]` attribute in `ppxlib.metaquot` which forced me to introduce all quotations with a `let loc in ...` prelude. 

I'd be surprised if there were nothing to change to this patch, so if you think it can lead to something, I'd be happy to work on it further.

### DIFF
--- a/src/ppx/dune
+++ b/src/ppx/dune
@@ -13,10 +13,10 @@ let () = Jbuild_plugin.V1.send @@ {|
  (public_name lwt_ppx)
  (synopsis "Lwt PPX syntax extension")
  (modules ppx_lwt)
- (libraries compiler-libs.common ocaml-migrate-parsetree ppx_tools_versioned)
+ (libraries ocaml-compiler-libs.common ppxlib)
  (ppx_runtime_libraries lwt)
  (kind ppx_rewriter)
- (preprocess (pps ppx_tools_versioned.metaquot_411 |} ^ bisect_ppx ^ {|))
+ (preprocess (pps ppxlib.metaquot|} ^ bisect_ppx ^ {|))
  (flags (:standard -w +A-4)))
 
 |}
--- a/src/ppx/ppx_lwt.ml
+++ b/src/ppx/ppx_lwt.ml
@@ -1,16 +1,11 @@
-open! Migrate_parsetree
-open! OCaml_411.Ast
-open Ast_mapper
+open! Ppxlib
+open Ast_builder.Default
 open! Ast_helper
-open Asttypes
-open Parsetree
-
-open Ast_convenience_411
 
 (** {2 Convenient stuff} *)
 
-let with_loc f {txt ; loc = _loc} =
-  (f txt) [@metaloc _loc]
+let with_loc f {txt ; loc } =
+  f ~loc txt
 
 (** Test if a case is a catchall. *)
 let is_catchall case =
@@ -27,7 +22,7 @@ let add_wildcard_case cases =
     List.exists is_catchall cases
   in
   if not has_wildcard
-  then cases @ [Exp.case [%pat? exn] [%expr Lwt.fail exn]] [@metaloc Location.none]
+  then cases @ (let loc = Location.none in [Exp.case [%pat? exn] [%expr Lwt.fail exn]])
   else cases
 
 (** {3 Internal names} *)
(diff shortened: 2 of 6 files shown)


raphael-proust:
I'm not familiar with the ppx (and especially with the ppx implementation), I'll review the MR next week, but I might have to ask for additional review.

Drup:
At the very least, I suggest we wait for the new ppxlib to be released, which is probably going to break compat one way or the other.

And yeah, if we switch to ppxlib, might as well try to use their fancy hook system...

pveber:
Waiting for the next release of `ppxlib` certainly is an option, but in that case it would be better that `ppx_tools_versioned` releases a version compatible with `ocaml-migrate-parsetree` 2.0.0. Otherwise `lwt` would hold back other libraries that switched to `ppxlib`.

pveber:
Seems there won't be any release of `ppx_tools_versioned` compatible with OMP 2.0.0, see ocaml-ppx/ppx_tools_versioned#28

kit-ty-kate:
There is currently no release of ppxlib compatible with both omp2 and OCaml 4.12. The latest release is compatible with OCaml 4.12 but requires omp2.

This makes packages relying on both lwt_ppx and some other ppxlib-based ppxs, uninstallable when using the soon to come OCaml 4.12.
It would be very helpful to have this out sooner rather than later.

src/ppx/dune
```diff
@@ -13,10 +13,10 @@ let () = Jbuild_plugin.V1.send @@ {|
  (public_name lwt_ppx)
  (synopsis "Lwt PPX syntax extension")
  (modules ppx_lwt)
- (libraries compiler-libs.common ocaml-migrate-parsetree ppx_tools_versioned)
+ (libraries compiler-libs.common ppxlib)
```
aantron, line 16: AFAICT we are now using `ocaml-compiler-libs.common` (e.g. new line 268).

src/ppx/ppx_lwt.ml
```diff
@@ -240,37 +235,37 @@ let lwt_expression mapper exp attributes ext_loc =
...
-let mapper =
-  { default_mapper with
+class mapper = object (self)
+  inherit Ast_traverse.map as super
 
-    structure = begin fun mapper structure ->
-      if !warned then
-        default_mapper.structure mapper structure
+  method! structure = begin fun structure ->
+      if !warned then super#structure structure
```
aantron, line 263: Please reduce formatting changes like here, where the change on old line 268 also wrongly appears to affect old line 267. This applies to many of the other diff hunks also, parts of which have changed only in indentation or splitting of tokens across lines. The formatting in this file is not great, but we should change it in separate work because this change is a bit scary and keeping it focused will be much easier for review, and for debuging later, if necessary, if this ends up as someone's `git bisect` endpoint :)

aantron (commented):
Thanks. This looks correct up to some nits.

Could you apply this patch to get opam pin/development installation working again? Related: https://github.com/aantron/bisect_ppx/pull/327. This should also (help) fix the CI.

```diff
diff --git a/lwt.opam b/lwt.opam
index 277dea245..0fe007bfd 100644
--- a/lwt.opam
+++ b/lwt.opam
@@ -28,7 +28,8 @@ depends: [
   "result" # result is needed as long as Lwt supports OCaml 4.02.
   "seq" # seq is needed as long as Lwt supports OCaml < 4.07.0.

-  "bisect_ppx" {dev & >= "2.0.0"}
+  # Until https://github.com/aantron/bisect_ppx/pull/327.
+  # "bisect_ppx" {dev & >= "2.0.0"}
   "ocamlfind" {dev & >= "1.7.3-1"}
 ]

```

src/ppx/dune
pveber: done

src/ppx/ppx_lwt.ml
pveber: Yes I understand it very well, sorry about that. I hope I fixed all superfluous changes.

pveber:
I tried to fix the various problems and I rebased the PR on master.

aantron:
Thanks!

Note that `lwt_ppx` now implicitly requires OCaml 4.04. It previously required only 4.02.

I've adjusted the CI accordingly.

pveber:
No problem, thanks for the careful review!

Issue #803, opened 2020-09-05
can `getgrgid` and `getgrnam` handle ERANGE by retrying with a bigger buffer?


dwang20151005:
I saw the following test failure:

```
[2020-09-04 12:53:37.834749-04:00] Testing library 'unix'...

[2020-09-04 12:53:37.834749-04:00] .................................SS.....................................................................................................EE.............................................................................
[2020-09-04 12:53:37.834749-04:00] Test 'getgrgid and Unix.getgrgid' in suite 'lwt_unix' raised 'Unix.Unix_error(Unix.ERANGE, "getgrgid", "")'

[2020-09-04 12:53:37.834749-04:00] Test 'getgrnam and Unix.getgrnam' in suite 'lwt_unix' raised 'Unix.Unix_error(Unix.ERANGE, "getgrnam", "tech")'
```

I think this is because lwt trusts `sysconf(_SC_GETGR_R_SIZE_MAX)` to
predict the buffer required, and does not handle `ERANGE`:

https://github.com/ocsigen/lwt/blob/master/src/unix/unix_c/unix_get_pw_gr_nam_id_job.c#L82

`man 3 getgrgid` suggests retrying with a bigger buffer:

```
The call

    sysconf(_SC_GETGR_R_SIZE_MAX)

returns  either -1, without changing errno, or an initial suggested size for buf.  (If
this size is too small, the call fails with ERANGE, in which case the caller can retry
with a larger buffer.)
```

This is what the corresponding `Core` code does:

https://github.com/janestreet/core/blob/master/src/core_unix.ml#L1880

I wish I could offer you a PR, but `unix_get_pw_gr_nam_id_job.c` is
more C than I can handle.  I patched my lwt locally with a minimum
value I know to be sufficient for now:

```
diff --git a/external/lwt/src/unix/unix_get_pw_gr_nam_id_job.c b/external/lwt/src/unix/unix_get_pw_gr_nam_id_job.c
--- a/external/lwt/src/unix/unix_get_pw_gr_nam_id_job.c
+++ b/external/lwt/src/unix/unix_get_pw_gr_nam_id_job.c
@@ -81,6 +81,7 @@ static value alloc_group_entry(struct gr
     {                                                                  \
         size_t buffer_size = sysconf(_SC_##CONF##_R_SIZE_MAX);         \
         if (buffer_size == (size_t)-1) buffer_size = 16384;            \
+        if (buffer_size < 16 * 1024) buffer_size = 1024 * 16;          \
[...]

PR #808, opened 2020-09-22
unix_get_pw_gr_nam_id_job: retry with a bigger buffer on ERANGE

raphael-proust:
Fixes https://github.com/ocsigen/lwt/issues/803

### DIFF
--- a/src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
+++ b/src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
@@ -16,6 +16,7 @@
 #include <pwd.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <errno.h>
 
 #include "lwt_unix.h"
 
@@ -81,9 +82,16 @@ static value alloc_group_entry(struct group *entry)
     {                                                                  \
         size_t buffer_size = sysconf(_SC_##CONF##_R_SIZE_MAX);         \
         if (buffer_size == (size_t)-1) buffer_size = 16384;            \
-        job->buffer = (char *)lwt_unix_malloc(buffer_size);            \
-        job->result = FUNC##_r(job->ARG, &job->entry, job->buffer,     \
-                               buffer_size, &job->ptr);                \
+        while (1)                                                   \
+        {                                                              \
+            job->buffer = (char *)lwt_unix_malloc(buffer_size);        \
+            job->result = FUNC##_r(job->ARG, &job->entry, job->buffer, \
+                                   buffer_size, &job->ptr);            \
+            if (job->result != ERANGE) break;                          \
+            buffer_size = buffer_size * 2;                             \
+            if (buffer_size > 1 << 20) break;                          \
+            free(job->buffer);                                         \
+        }                                                              \
     }                                                                  \
                                                                        \
     static value result_##FUNC(struct job_##FUNC *job)                 \


raphael-proust:
I'm not sure how to test this. Is there a way to set some environment so that the while loop is actually entered?

src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
```diff
@@ -84,6 +85,15 @@ static value alloc_group_entry(struct group *entry)
         job->buffer = (char *)lwt_unix_malloc(buffer_size);            \
         job->result = FUNC##_r(job->ARG, &job->entry, job->buffer,     \
                                buffer_size, &job->ptr);                \
+        while (job->result == -1 && errno == ERANGE)                   \
+        {                                                              \
+            free(job->buffer);                                         \
+            buffer_size = buffer_size * 2;                             \
+            if (buffer_size < 0) break;                                \
```
aantron, line 92: This condition is always false, because `buffer_size` is a `size_t`, which is unsigned, and so is never less than zero. I suppose this should be a bound check against some large buffer size instead.

src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
```diff
@@ -84,6 +85,15 @@ static value alloc_group_entry(struct group *entry)
         job->buffer = (char *)lwt_unix_malloc(buffer_size);            \
         job->result = FUNC##_r(job->ARG, &job->entry, job->buffer,     \
                                buffer_size, &job->ptr);                \
+        while (job->result == -1 && errno == ERANGE)                   \
```
aantron, line 88: I think this comparison with `-1` is incorrect. From the [man page](https://www.man7.org/linux/man-pages/man3/getpwnam.3.html):

> On success, `getpwnam_r()` and `getpwuid_r()` return zero, and set
> `*result` to `pwd`.  If no matching password record was found, these
> functions return `0` and store `NULL` in `*result`.  In case of error, an
> error number is returned, and `NULL` is stored in `*result`.

So it should probably be `job->result != 0`. Compare that also with `if (result)` down in `result##FUNC` later on in this file (although it would be clearer if it was `if (result != 0)`).

aantron: As for `errno`, it's possible that `job->result` itself will be `ERANGE`. Otherwise, if still checking `errno`, there is this in the man page:

> If an error occurs, `errno` is set appropriately.  If one
> wants to check `errno` after the call, it should be set to zero before
> the call.

However, that's in the paragraph for the non-`_r` versions of these functions, so I wonder if the `_r` versions set `errno` at all, in which case `job->result` should be checked for `ERANGE`. I would check this locally in a small C program by calling these functions and triggering any failure, and examining the return value and `errno`.

aantron (changes requested):
We probably won't be able to test this reasonably in the test suite. I'd test it locally by inserting `ERANGE` artificially (i.e. temporarily hardcoding it in response to the initial buffer size, and running the code, with maybe some prints inserted). Perhaps @dwang20151005 would also be willing to test it once under real conditions once the PR is finalized.

aantron:
I also suggest to, optionally, adjust the code so that the `malloc` and the system call happen in only one place, i.e. have the loop body do allocation, system call, and freeing, and `break` out on success or very large buffer size. Leaving in two system call sites invites future contributors to overlook one of them while editing the other :)

src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
raphael-proust: I'm checking if it's bigger than 2^30. I picked that constant sort of randomly: I wanted something less than 2^32 for compatibility with 32-bit architectures, but otherwise not real reason.

src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
raphael-proust: I have a new version (pushed) that should handle correctly:
- success case: result is `0` and buffer has not `NULL` (breaks and post-processing is handled by `result_##FUNC`)
- not found: result is `0` and buffer has `NULL` (same)
- error: result is non `0` and buffer has `NULL` (result is checked (rather than errno, as suffested by manpage) for ERGANE and either (a) it is post-processed by `result_##FUNC` of (b) buffer is freed, size is increased, buffer-size is checked, and we just retry)

It uses a lot more breaks than before. So I'm even less confident in the code.

raphael-proust:
> I also suggest to, optionally, adjust the code so that the `malloc` and the system call happen in only one place, i.e. have the loop body do allocation, system call, and freeing, and `break` out on success or very large buffer size.

I've done that too.

src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
```diff
@@ -81,9 +82,17 @@ static value alloc_group_entry(struct group *entry)
...
+        while (1)                                                   \
+        {                                                              \
+            job->buffer = (char *)lwt_unix_malloc(buffer_size);        \
+            job->result = FUNC##_r(job->ARG, &job->entry, job->buffer, \
+                                   buffer_size, &job->ptr);            \
+            if (job->result == 0) break;                               \
+            if (job->result != ERANGE) break;                          \
+            free(job->buffer);                                         \
+            buffer_size = buffer_size * 2;                             \
+            if (buffer_size > 1 << 30) break;                          \
```
aantron, line 94: > I'm checking if it's bigger than 2^30. I picked that constant sort of randomly: I wanted something less than 2^32 for compatibility with 32-bit architectures, but otherwise not real reason.

That actually wouldn't be compatible with 32-bit architectures if it ever reached anywhere near that limit, since most 32-bit OSs probably wouldn't let you allocate a 1GB buffer. I don't think we would ever keep failing with `ERANGE` until reaching such a size, but I think, for defensive programming purposes, it's best to choose something smaller (especially since we are building up to this in powers of 2, and in case of some kind of heap fragmentation, we are risking allocating almost 2GB of memory in a geometric progression, say, on a buggy system). I suggest picking 1MB (`1 << 20`). This should be almost risk-free on most systems, and it's plenty enough for the 16K that seemed to solve the original issue.

src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
```diff
@@ -81,9 +82,17 @@ static value alloc_group_entry(struct group *entry)
...
         if (buffer_size == (size_t)-1) buffer_size = 16384;            \
-        job->buffer = (char *)lwt_unix_malloc(buffer_size);            \
-        job->result = FUNC##_r(job->ARG, &job->entry, job->buffer,     \
-                               buffer_size, &job->ptr);                \
+        while (1)                                                   \
+        {                                                              \
+            job->buffer = (char *)lwt_unix_malloc(buffer_size);        \
+            job->result = FUNC##_r(job->ARG, &job->entry, job->buffer, \
+                                   buffer_size, &job->ptr);            \
+            if (job->result == 0) break;                               \
```
aantron, line 90: This check is redundant with the next one, and it's probably best to delete it because the `!= ERANGE` is better for clarity.

src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
```diff
@@ -81,9 +82,17 @@ static value alloc_group_entry(struct group *entry)
...
+            if (job->result != ERANGE) break;                          \
+            free(job->buffer);                                         \
```
aantron, line 92: I think we risk a double `free` here. If the result is `ERANGE` but the buffer size reaches the limit, we will have freed the buffer and not allocated a new one. `result_##FUNC` will then attempt to free the buffer again. We should probably only `free` the buffer in this worker function if the next buffer is still not too large, i.e. move this line to be last in the loop body.

aantron: I guess I forgot what C memory management is like when I said this loop is simple :P

src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
raphael-proust: Fixed.

src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
raphael-proust: Fixed.

src/unix/unix_c/unix_get_pw_gr_nam_id_job.c
raphael-proust: Fixed. Good catch!

Yeah, I haven't done C in a long time and I'd forgotten the hassle… I'm leaving this one unresolved because of that.

aantron (approved):
LGTM.

PR #809, opened 2020-09-26
Handle uncaught exceptions in Lwt_react.of_stream.

paurkedal:
Currently exceptions raised during reactive updates set up by `Lwt_react.of_stream` will cause the event to stop without notice. My suggestion is to call `!Lwt.async_exception_hook`.

I ran into this while hunting down a bug in an Eliom application using `Eliom_react.S.Down.of_react` and with this PR I get the needed feedback in the JS console.

### DIFF
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -872,8 +872,8 @@ val async_exception_hook : (exn -> unit) ref
 (** Reference to a function, to be called on an "unhandled" exception.
 
     This reference is used by {!Lwt.async}, {!Lwt.on_cancel}, {!Lwt.on_success},
-    {!Lwt.on_failure}, {!Lwt.on_termination}, {!Lwt.on_any}, and the deprecated
-    {!Lwt.ignore_result}.
+    {!Lwt.on_failure}, {!Lwt.on_termination}, {!Lwt.on_any},
+    {!Lwt_react.of_stream}, and the deprecated {!Lwt.ignore_result}.
 
     The initial, default implementation prints the exception, then terminates
     the process with non-zero exit status, as if the exception had reached the
--- a/src/react/lwt_react.cppo.ml
+++ b/src/react/lwt_react.cppo.ml
@@ -101,7 +101,8 @@ module E = struct
   let of_stream stream =
     let event, push = create () in
     let t =
-      Lwt.pause () >>= fun () -> Lwt_stream.iter (fun v -> push v) stream in
+      Lwt.pause () >>= fun () ->
+      Lwt_stream.iter (fun v -> try push v with exn -> !Lwt.async_exception_hook exn) stream in
     with_finaliser (cancel_thread t) event
 
   let delay thread =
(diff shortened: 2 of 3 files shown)


raphael-proust:
Hiya,  
Sorry for the long time it took for me to reply.

I'm guessing you stumble onto this issue because the `stream` you passed to `of_stream` got something pushed to it during the step that was triggered by a previous push. Can you share an example of this? Or an example of whatever caused the exception on your side?


I have the following reservations:

- Major: the documentation of `of_stream` should explain what can lead to an exception being raised and possibly even give advice about how to avoid situations that lead to an exception being raised. This is why the example I'm asking about above is important. On this note, do we want to simply put a layer around the stream so that it delays popping elements until we are out of the current react step if any? And, would this prevent the exception you observed altogether?
- Minor: Are there other places that this one where we lose exceptions? How can we find them?
- Minor: `async_exception_hook` is supposed to be for handling exceptions that are raised in promises that have been ignored. It's not an exact fit in terms of intended use. OTOH, there isn't anything else and it's not too ugly of a design.

aantron:
@raphael-proust, regarding the last point, there are a few other places where we use `Lwt.async_exception_hook` when it's not directly related to promises. I believe one set is the `Lwt_io.establish_server` family of functions.

paurkedal:
@raphael-proust It's simpler than that. When triggering a reactive update due to setting a signal, event, or running `Step.execute`, the caller will receive exceptions raised during the reactive update. So, if we give a buggy function to `React.E.map`:
```ocaml
let () =
  let counter =
    let n = ref 10 in
    Lwt_stream.from
      (fun () -> Lwt.map (fun () -> decr n; Some !n) (Lwt_unix.sleep 0.2))
  in
  let e1 = Lwt_react.E.of_stream counter in
  let e2 = Lwt_react.E.map (fun i -> 1260 / i) e1 in
  let _e3 = Lwt_react.E.trace (Printf.printf "%d\n%!") e2 in
  let waiter, _ = Lwt.wait () in
  Lwt_main.run waiter
```
the current version will silently stop after printing 1260, while this PR we're told what is wrong:
```
$ dune exec ./test.exe
140                  
157
180
210
252
315
420
630
1260
Fatal error: exception Division_by_zero
```
Dune file:
```
(executable
 (name test)
 (modules test)
 (libraries lwt lwt_react lwt.unix))
```

paurkedal:
Do you still think the documentation is unclear?  I think it hinges on the meaning of "updating the event", which refers to the execution of the reactive update step, which involve whatever the user has attached to the result of `Lwt_react.E.of_stream`.

raphael-proust:
Oh I see, I wasn't familiar with React's policy of sending exceptions back to whichever called triggered the step.

I guess the documentation is ok. I'd maybe add something along the lines of the chunk below. But being unfamiliar with the exception-handling of React I might have written something that is immediately evident to React users. So if you feel like the added details are unnecessary just say so and we can go with your version.

```
        If updating the event causes an exception at any point during the
        update step, the excpetion is passed to
        [!]{!Lwt.async_exception_hook}, which terminates the process by
        default.
```

paurkedal:
I agree, your phrasing is better. I pushed an update.

raphael-proust:
LGTM. I'll merge after a bit in case someone else wants to review / comment.

PR #810, opened 2020-09-28
lwt: fix typo (immedaitely -> immediately)

hannesm:


raphael-proust:
Thanks for taking the time to open the MR!

PR #811, opened 2020-10-06
Add a initial engine based on luv

ulrikstrid:
This is meant to be a conversation starter above all else. The code is pretty naive at this point.

It creates a new engine that is based on `luv` (lovingly called the luv engine).
This could be a nice migration path to move the whole unix module to be using `libuv` which is very well maintained and tested since it's been used by node for years.

Locally I have the same test failures I have with the `libev` engine.


src/unix/lwt_engine.ml
```diff
@@ -197,6 +197,53 @@ end
 
 class libev_deprecated = libev ()
 
+class libuv () = object
+  inherit abstract
+
+  val loop = Luv.Loop.init () |> Result.get_ok
```
aantron, line 203: Is it a deliberate choice not to use libuv's default loop?

src/unix/lwt_engine.ml
```diff
@@ -197,6 +197,53 @@ end
...
+
+  method loop = loop
+
+  method private cleanup = Luv.Loop.stop loop
+
+  method iter block =
+    try
+      match (block) with
+      | true -> Luv.Loop.run ~loop ~mode:`ONCE () |> ignore
+      | false -> Luv.Loop.run ~loop () |> ignore
```
aantron, line 213: I believe this needs ``~mode:`NOWAIT``.

src/unix/lwt_engine.ml
ulrikstrid: That is probably better yes

src/unix/lwt_engine.ml
```diff
@@ -197,6 +197,55 @@ end
...
+    try
+      match (block) with
+      | true -> Luv.Loop.run ~loop ~mode:`ONCE () |> ignore
+      | false -> Luv.Loop.run ~loop ~mode:`NOWAIT () |> ignore
+    with exn ->
+      Luv.Loop.stop loop;
+      raise exn
+
+  method private register_readable fd f =
+    let p = Luv.Poll.init ~loop (Obj.magic fd) in
```
aantron, line 219: This version still has the "naked" `Obj.magic`; did you mean to include factoring it out in this commit?

ulrikstrid: Fixed in a new commit

src/unix/lwt_engine.ml
```diff
@@ -197,6 +197,57 @@ end
...
+      Luv.Loop.stop loop;
+      raise exn
+
+  method private register_readable fd f =
+    let p = Luv.Poll.init ~loop (unix_fd_to_fd fd) in
+    match p with
+    | Ok poll ->
+        let () = Luv.Poll.start poll [`READABLE] (fun _ -> f ()) in
+        lazy(Luv.Poll.stop poll |> ignore)
+    | Error e -> failwith (Printf.sprintf "This is probably a error in Lwt, please open a issue on the repo. \nError message: %s" (Luv.Error.strerror e))
```
aantron, line 226: I suggest adding a bit to each message what the error is, e.g. "Could not register fd for read polling, ......."

src/unix/lwt_engine.ml
```diff
@@ -197,6 +197,57 @@ end
...
+
+  method private cleanup = Luv.Loop.stop loop
+
+  method iter block =
+    try
+      match (block) with
+      | true -> Luv.Loop.run ~loop ~mode:`ONCE () |> ignore
+      | false -> Luv.Loop.run ~loop ~mode:`NOWAIT () |> ignore
+    with exn ->
+      Luv.Loop.stop loop;
```
aantron, line 217: I would omit this, as we are now using the default loop. I guess we don't need the exception handler at all here, then.

src/unix/lwt_engine.ml
```diff
@@ -197,6 +197,57 @@ end
...
+let unix_fd_to_fd (fd: Unix.file_descr) : int = Obj.magic fd
+
+class libuv () = object
+  inherit abstract
+
+  val loop = Luv.Loop.default ()
+
+  method loop = loop
+
+  method private cleanup = Luv.Loop.stop loop
```
aantron, line 209: This method isn't called anywhere.

src/unix/lwt_engine.ml
```diff
@@ -197,6 +197,57 @@ end
...
+    match t with
+    | Error _ -> lazy(())
+    | Ok timer ->
+      let timer_fn = match repeat with
+      | true -> Luv.Timer.start ~repeat:delay_ms timer
+      | false -> Luv.Timer.start timer
+      in
+      match timer_fn delay_ms f with
+      | Ok () -> lazy(Luv.Timer.stop timer |> ignore)
+      | Error _ -> lazy(())
```
aantron, line 248: And `failwith` here.

test/test.ml
```diff
@@ -264,6 +264,8 @@ let run library_name suites =
             ((suite.suite_name, outcomes)::aggregated_outcomes) rest
       end
   in
+  
+  Lwt_engine.set (new Lwt_engine.libuv ());
```
aantron, line 268: It's ok to leave this for now, but we will probably need to remove it before merging (note to self). We might do something like factor out the Luv engine into a separate package, and give it a tester that does this call, then calls back into the ordinary lwt.unix tester.

src/unix/lwt_engine.ml
ulrikstrid: This method is present in the libev code, but is it okay to remove?

ulrikstrid:
I'm outside of my comfort zone with the `file_descr` stuff but I think we might be able to pick the solution from here: https://github.com/aantron/luv/commit/705ec144a9f047260de8140583e88d016e33e061

aantron:
I should probably restore that code to Luv, as it has use outside Lwt for those that want to use Luv together with some `Unix` code. In the meantime, though, you're welcome to borrow from it as needed. We can adjust Lwt later, if Luv gets that code again.

ulrikstrid:
It looks like the code is using some internal code that is not exposed outside of `Luv`, specifically `C.Types.Os_fd.t`. Not sure how to handle that.

src/unix/lwt_engine.ml
aantron: Also, IIRC, `Luv.Loop.run` can't raise any exceptions, so another argument in favor of not trying to handle any here.

src/unix/lwt_engine.ml
aantron: Ah yes, it's required by the base class. Sorry.

src/unix/lwt_engine.mli
```diff
@@ -140,6 +140,16 @@ class libev : ?backend:Ev_backend.t -> unit -> object
     (** Returns [loop]. *)
 end
 
+class libuv : unit -> object
+  inherit t
+
+  val loop : Luv.Loop.t
+    (** The libuv loop used for this engine. *)
+
+  method loop : Luv.Loop.t
```
aantron, line 149: These don't need to be exposed for now, however. I assume this is the result of basing the code on `libev`.

aantron:
> It looks like the code is using some internal code that is not exposed outside of `Luv`, specifically `C.Types.Os_fd.t`. Not sure how to handle that.

`C.Types.Os_fd.t` is a Ctypes binding to libuv's `os_fd_t`. Since Lwt is not using Ctypes (at the moment), it can just be replaced by `int`.

It might be more work-efficient, however, to simply not support Windows during this PR. We will merge the PR in a non-intrusive way (as an optional package in this repo), so the lack of Windows support won't affect anyone. Then, I'll restore the `Unix`-libuv fd conversions in Luv (probably in a helper package), and we can simply call those from Lwt to help get the Windows support working. As the code matures more and more, we can consider moving it from the optional package into main Lwt. I'll post a suggested roadmap for doing so and follow-on work that is enabled by the libuv integration, and we will discuss with @raphael-proust and @avsm.

It looks like the Luv test cases will also [benefit](https://github.com/aantron/luv/commit/705ec144a9f047260de8140583e88d016e33e061#diff-5fdb98dcbf975772ed6e0646dd91fedfR52-R55) from the conversions:

```ocaml
(* Until https://github.com/libuv/libuv/pull/1498. This implementation will not
   work on Windows. One can be provided, but hopefully the PR lands first. *)
let unix_fd_to_file : Unix.file_descr -> Luv.File.t =
  Obj.magic
```

The Luv tester also shows that there is precedent for getting away with `Obj.magic` on Unix, at least for non-production code, which is what this integration will be in its first days after merging if following my suggestion above :)

aantron:
So focusing on Unix only, at the moment, for me locally, the Unix tester passes, but the React and PPX testers hang.

In Travis, I see errors like [this](https://travis-ci.org/github/ocsigen/lwt/jobs/733968318#L339):

```
(cd _build/default/test/ppx && ./main.exe)
Testing library 'ppx'...
Fatal error: exception Failure("Could not register fd for read polling, this is probably a error in Lwt, please open a issue on the repo. \nError message: operation not permitted")
Raised at file "stdlib.ml", line 29, characters 17-33
Called from file "src/unix/lwt_engine.ml", line 88, characters 15-37
Called from file "src/unix/lwt_engine.ml", line 73, characters 29-47
Called from file "src/core/lwt_sequence.ml", line 132, characters 31-47
Called from file "src/unix/lwt_engine.ml", line 72, characters 4-112
Called from file "src/unix/lwt_engine.ml", line 472, characters 19-36
Called from file "test/test.ml", line 268, characters 2-42
Called from file "test/ppx/main.ml", line 165, characters 8-32
        main alias test/core/runtest
```

and similar errors for React.

src/unix/lwt_engine.ml
```diff
@@ -197,6 +197,57 @@ end
...
+      Luv.Loop.stop loop;
+      raise exn
+
+  method private register_readable fd f =
+    let p = Luv.Poll.init ~loop (unix_fd_to_fd fd) in
+    match p with
+    | Ok poll ->
+        let () = Luv.Poll.start poll [`READABLE] (fun _ -> f ()) in
+        lazy(Luv.Poll.stop poll |> ignore)
+    | Error e -> failwith (Printf.sprintf "Could not register fd for read polling, this is probably a error in Lwt, please open a issue on the repo. \nError message: %s" (Luv.Error.strerror e))
```
aantron, line 226: I suggest replacing `strerror` by [`err_name`](https://aantron.github.io/luv/luv/Luv/Error/index.html#val-err_name) (so it prints `EPERM`, etc.), because, in practice, when we are debugging this, we will be looking either at man pages that list e.g. `EPERM`, or in the libuv source code, which also uses these constants. The full error descriptions aren't that useful, since we are not giving "actionable" feedback to "end" users, and meanwhile we end up having to mentally translate them back into error codes.

ulrikstrid:
So I think I addressed your comments now.

The next step is to break it out of lwt-unix so that not everyone needs to depend on Luv at this point (even if I think they should 😉). How do you want me to do that?

aantron:
> So I think I addressed your comments now.

👍 



> The next step is to break it out of lwt-unix so that not everyone needs to depend on Luv at this point (even if I think they should 😉). How do you want me to do that?

Create a new directory, say, `src/unix/luv`, with its own `dune` file that declares a library called `lwt_luv` (or any other suitable name) that depends on luv and lwt.unix. Then move the `libuv` engine class into a new `.ml` file in that directory (probably with an `.mli` file for it, too). Then add `lwt_luv.opam` with some metadata for that library.

Then create `test/unix/luv_main.ml` as a copy of `main.ml`, but insert the call to set the engine to libuv into it. Remove it from `main.ml`. Declare the `luv_main` executable `test/unix/dune`, and make it depend on `lwt_luv`. Don't add it to the `runtest` alias as we don't want Dune to run it on Windows as part of our AppVeyor build. Instead, we will run it manually only in Travis. After this line in `.travis.yml`:

https://github.com/ocsigen/lwt/blob/1157930612348286fcb5b675b3309f04e0e5a2b8/.travis.yml#L100

...insert something like `dune exec test/unix/luv_main.exe`.

Copying `main.ml` is probably not ideal in the long term; we'd want to list out the test suites in one place instead. However, I hope this separate packaging is only temporary, and eventually `main.ml` will itself become the libuv tester if Lwt starts using libuv by default.

aantron:
Also we probably won't ever release `lwt_luv.opam`, as we'd rather see if we will merge the libuv support into the main Lwt instead. So it doesn't have to be neat. It just has to have the dependency metadata to make Travis and/or contributors install Luv.

ulrikstrid:
I think I did the breakout correctly, I don't have my opam drivers license yet so might have missed something there.

aantron:
I think the most important missing thing is triggering the new `luv_main.exe` tester from `.travis.yml`.

ulrikstrid:
I did that change locally but forgot to add it 🙈

src/unix/luv/dune
```diff
@@ -0,0 +1,18 @@
...
+let preprocess =
+   match Sys.getenv "BISECT_ENABLE" with
+   | "yes" -> "(preprocess (pps bisect_ppx))"
+   | _ -> ""
+   | exception _ -> ""
+
+let () = Jbuild_plugin.V1.send @@ {|
+
+(library
+ (name lwt_luv)
```
aantron, line 12: IIRC `(name ...)` is redundant if `(public_name ...)` is defined nowadays. We only have both fields in other `dune` files because it's older code.

src/unix/luv/dune
```diff
@@ -0,0 +1,18 @@
...
+ (public_name lwt_luv)
+ (libraries luv lwt lwt_unix)
```
aantron, line 14: I think it's safer to depend on the public name `lwt.unix`.

aantron (commented):
The libuv tests are now failing. They seem to be not terminating on Mac, and [failing on Linux](https://travis-ci.com/github/ocsigen/lwt/jobs/401733667#L364) with

```
Unix.Unix_error(Unix.EAGAIN, "recv_notifications", "")
```

which strongly suggests some kind of polling error.

I am concerned that they appeared to be woking before.

raphael-proust:
The tests also fail for me (Archlinux, x86_64, libuv 1.40, luv 0.5.6). There are multiple failures and the test even gets stuck with some leftover unterminated processes.

I'll have a look at the code this week.

ulrikstrid:
@raphael-proust I and @aantron has tried to debug a bit and it is something with `Lwt_unix.fork` that breaks.
There's some assumption in my code about lwt and libuv that doesn't hold.

lwt_luv.opam
```diff
@@ -0,0 +1,28 @@
...
+bug-reports: "https://github.com/ocsigen/lwt/issues"
+
+maintainer: [
+  "Anton Bachin <antonbachin@yahoo.com>"
+  "Ulrik Strid <ulrik.strid@outlook.com>"
+]
+dev-repo: "git+https://github.com/ocsigen/lwt.git"
+
+depends: [
+  "lwt"
```
avsm, line 18:
```suggestion
  "lwt" {=version}
```

src/unix/luv/dune
```diff
@@ -0,0 +1,17 @@
+(* -*- tuareg -*- *)
+
+let preprocess =
+   match Sys.getenv "BISECT_ENABLE" with
```
avsm, line 4: This can be separate from this PR, but dune now has direct support for bisect_ppx: https://dune.readthedocs.io/en/latest/instrumentation.html

So the need for an environment variable can be removed from all the dune files and replaced with the direct instrumentation support.

src/unix/luv/lwt_luv.ml
```diff
@@ -0,0 +1,42 @@
+let unix_fd_to_fd : Unix.file_descr -> int = Obj.magic
```
avsm, line 1: This will break on Windows unfortunately, where `Unix.file_descr` is a struct. See https://github.com/ocaml/ocaml/pull/1990 for some ongoing work by @dra27 to expose this function more systematically. It may be worth picking out C binding from that PR directly into Lwt to facilitate the conversion.

avsm (changes requested):
It's pretty exciting that you've hooked in libuv, since this is the library I'd like to use when adding multicore IO support in a future branch.  Just a few minor comments: my main concern is that this won't be stable on Windows -- have you had a chance to try it there?

Also copying @djs55, who is a user of libuv (via uwt) in https://github.com/moby/vpnkit, and may be interested in shifting to this Lwt_engine approach rather than directly juggling uwt and lwt simultaneously.

src/unix/luv/lwt_luv.ml
dra27: Alternatively, if I understand the PR correctly, it would be disable this engine for Windows and bat the issue back to luv. I don't know how libev does this (because I haven't looked 😊), but my instinct is that the API in luv is wrong and shouldn't expose int, but an abstract type along with a `luv_abstract_file_descr_of_unix_file_descr` function which would deal with this. Surely it affects any Windows users of luv?

src/unix/luv/dune
psafont: This would mean lifting the version required to dune 2.7 (from 1.8), which means dropping ocaml versions from 4.02 to 4.08.

A new PR would be better do discuss the tradeoffs.

avsm: Earlier versions of OCaml should work just fine with modern dune. There is a ocaml-secondary-compiler package in opam which builds the `dune` binary, and the dune-* libraries are 4.03+

aantron: The issue is that we force users of Lwt to also raise their Dune version, which might be too aggressive at this point. I agree that this should be discussed separately.

src/unix/luv/lwt_luv.ml
aantron: @dra27 Luv exposes an abstract type. I'm not sure what API you are referring to. Luv works on Windows and is tested on it in CI all the time. I understand it is also being used in production on Windows.

As @ulrikstrid already mentioned, and as we discussed earlier in this PR, all this `Obj.magic` and fd casting business is related only to getting this integration working on Unix "by any means necessary." Once that's done, we will clean up the Unix implementation and get this PR working the right way on Windows, too.

aantron: Ah well I just realized you mean the polling API specifically, like [here](http://docs.libuv.org/en/v1.x/poll.html#api). libuv exposes it as an `int` &mdash; I understand on Windows this should be a CRT fd, and I don't know which kinds of files have a CRT fd on Windows (but at least some do). In the long term, we may have to do some PRs to libuv itself to get this to work fully on Windows at the libuv level, and take a look at what is already available in the libuv 2.x branch. In the even longer term, most of Lwt would be rewritten to bypass libuv poll handles and use libuv directly, if we went with a thorough conversion. That would make the issue of any polling problems on the libuv level moot for most purposes.

aantron: (In general, by the way, the libuv API is somewhat sloppy and inconsistent in how it represents fds across its various modules, but I am sure we can get the right dynamics for getting everything working)

ulrikstrid:
The latest commit is obviously not OK since it adds `luv` as a dependency to `lwt` but it seems to fix Linux if we run a `Luv.Loop.fork` after forking. If I run the test without concurrency on my mac they seem to be fine here as well.

I guess we need to move the `Luv.Loop.fork` call into the luv-engine somehow. I'll try to find the issue that makes it run forever but this is at least a step forward.

mseri:
If there is no alternative, how bad would it be to allow an optional external `unit->unit` callback to run in the Lwt `fork`?

ulrikstrid:
> If there is no alternative, how bad would it be to allow an optional external `unit->unit` callback to run in the Lwt `fork`?

A solution would be to add a `fork` method to the engine class that is a noop (`unit -> unit`) on all engines except libuv. I guess this is technically a breaking change though which is unwanted. But maybe there is a better way to detect a fork from within the  engine?

src/unix/luv/lwt_luv.mli
```diff
@@ -0,0 +1,2 @@
+(** Engine based on libev. *)
```
ulrikstrid, line 1:
```suggestion
(** Engine based on libuv. *)
```

src/unix/luv/lwt_luv.ml
ulrikstrid: I added the code that I got from the luv history so there's no `Obj.magic` here anymore. Hopefully that code will be added back into `Luv` at some point so that we don't have to maintain it here since it might be flaky.

lwt_luv.opam
ulrikstrid: We (me and @aantron) decided to make the version of the engine package `1.0.0`, and since it should work with any `Lwt` verison and it probably won't ever get to opam in this form I don't think this change make sense.

aantron:
> I guess this is technically a breaking change though which is unwanted.

It shouldn't be a breaking change. We will add it as a method to the base class and override it in engines that need to do so. Existing custom engines outside Lwt won't need to do anything. OOP! :)

ulrikstrid:
I think there are just 2 tests failing now. I need to move the luv fork call into the engine as well. But this is getting closer.

ulrikstrid:
So the last main blocker is to understand why it runs forever on macOS in CI when it doesn't on my local machine.
Another open question is why it works better with multiple loops when forking.

Any ideas @aantron @raphael-proust @avsm

aantron:
> So the last main blocker is to understand why it runs forever on macOS in CI when it doesn't on my local machine.

Do you already know which specific test (or set of tests) is running forever? Looking at the Mac output, it looks like most tests completed. It seems you should be able to replace the dot output in the test runner with the name of the completed test case and suite. Once we know which test(s) is(are) hanging, we might be able to guess the problem. If not, I suggest to continue by removing other tests to rule out an interaction between tests as the problem.

And so on, normal debugging... but more painful because of the need to wait for CI.

There is also some way to get Travis to SSH into a machine, that allows you to then run commands inside the Travis worker, but it seems involved to set up and I've never tried it.

ulrikstrid:
Just letting you know I'm still working on this. I'm currently trying to debug the Linux failure/running forever.
Sometimes `Lwt_main.run` gets called in the child process when it should quit and then it hangs forever.

ulrikstrid:
There are still failing tests but I think the major stuff works now and I'm not sure it's worth it to fix the rest until we swap the rest out

aantron:
This is in my queue for reviewing for merge. Should have time soon.

src/unix/luv/lwt_luv.ml
```diff
@@ -0,0 +1,72 @@
+let l f = function
+| Result.Error e -> failwith (Luv.Error.err_name e)
+| Ok l -> List.iter (function
+| `DISCONNECT -> ()
+| `PRIORITIZED -> ()
+| `READABLE -> f ()
+| `WRITABLE -> f ()
```
aantron, line 7: This looks like it will call `f` twice if both `` `READABLE`` and `` `WRITABLE`` are returned. Is that intentional? Have you tried calling `f` only once, if either or both of those values are present? Is returning both even possible with the way the code is written? I suspect not, but I also suspect you have much more direct and recent experience with this now than I do :)

Also, if we always ignore `` `DISCONNECT`` and `` `PRIORITIZED`` (at least for now), why register for them later on down in this code?

aantron (commented):
Want to link again to this failure :) https://travis-ci.com/github/ocsigen/lwt/jobs/439876055#L414

src/unix/luv/luv_unix_stubs.c
```diff
@@ -0,0 +1,21 @@
+#include <caml/alloc.h>
+#include <caml/mlvalues.h>
+#include <caml/unixsupport.h>
+
+#include <uv.h>
+
+CAMLprim value luv_unix_fd_to_os_fd(value unix_fd, value os_fd_storage)
+{
+    uv_os_fd_t *os_fd = (uv_os_fd_t *)Nativeint_val(os_fd_storage);
```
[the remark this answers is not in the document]
ulrikstrid: As we talked about "offline" I'll just revert to use `Obj.magic` for now, this currently doesn't work on Windows anyway so we don't need to have that fragile code here.

src/unix/luv/lwt_luv.ml
ulrikstrid: This is  a  leftover from  when I was debugging. I'll change it to just listen for the event it needs (`READABLE and `WRITABLE depending on function) and then just call the callback.

ulrikstrid: And we should never get both `` `READABLE `` and `` `WRITABLE `` at the same  time since  we're explicitly passing in the mask we want.

src/unix/luv/luv_unix_stubs.c
```diff
@@ -0,0 +1,21 @@
...
+
+#ifndef _WIN32
+    *os_fd = Int_val(unix_fd);
+#else
+    if (Descr_kind_val(unix_fd) == KIND_HANDLE)
+        *os_fd = Handle_val(unix_fd);
+    else
+        *os_fd = -1;
```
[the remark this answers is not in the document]
ulrikstrid: Will remove for now according to the above comment

.travis.yml
```diff
@@ -136,7 +138,7 @@ scripts:
         - ./_cache
 
   - &esy
-    install: npm install -g esy
```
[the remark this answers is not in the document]
ulrikstrid: I think I wanted to fix the esy tests but some change didn't get reverted as it should have I think.

aantron:
Merging this, despite Travis CI not running at all. It (mostly) works locally on my system.

1. This doesn't affect "base" Lwt &mdash; almost all the new code is in `Luv_engine.ml`, in the separate package `lwt_luv.opam`, which we don't intend to release for the time being (but people can pin). The only exception to this is the addition of `method fork` to the base `Lwt_engine` class, which does nothing in that base class &mdash; though it should be mentioned in the changelog of 5.4.0, cc @raphael-proust.
2. The tests fail sporadically. We can stabilize this in and during follow-on work. Some of the occasional failures are due to too-strict timing tests. It's better to look at all that later, rather than risk excessive bit rot in this PR.

We will go from here as described in #813 "A Roadmap for Porting lwt.unix to libuv."

Thank @ulrikstrid and reviewers!

PR #814, opened 2020-11-04
small optimizations in lwt.ml

c-cube:


raphael-proust:
Thanks for the patches. I read through them and found it strange that your second patch for `all` worked (specifically that it would resolve at all considering `wake` is only ever used to resolve `waitall` in case of exception). And I discovered that the test suite only has tests for `all` with lists of size 2.

I've added multiple tests and indeed there's an issue. You can find the tests at https://github.com/raphael-proust/lwt/tree/add-more-tests-to-all

They pass if you drop 1f7bc7b30f6551908f0d63839387dab927a6be53 but not otherwise. Do you want to amend that commit? Or do you just want to drop it?

Additionally, did you by any chance micro-benchmark the changes?

c-cube:
just reverted that, as it's indeed not worth it, and added some tests. can you add  your own tests to the same branch? not sure it's doable with github on two distinct repos.

Didn't microbenchmark but it should definitely allocate less (no intermediate array that is immediately turned into a list).

raphael-proust:
I'm not sure if github will allow me to push onto your repo without having to setup some kind of rights before. A simple way to do something equivalent is if you fetch my branch and then cherry-pick the commit onto your branch and then you should be able to push onto your MR.

aantron:
GitHub does allow pushing into contributor branches, unless @c-cube has deliberately disabled that by unchecking it when opening the PR.

c-cube:
just rebased anyway :)

raphael-proust:
I think the code is ok now. I don't expect a big difference from the change to `both` but I think the change for `all` is good.

@aantron is there some kind of policy about commit-history? Should we squash a little bit to avoid a commit followed by a revert? Or is it fine to leave it as is for the merge?

aantron:
> @aantron is there some kind of policy about commit-history? Should we squash a little bit to avoid a commit followed by a revert? Or is it fine to leave it as is for the merge?

There is no policy. However, I've kept the history a straight line with squashes, because IMO most merge commits are pointless. I do look in history a lot and merges make the history difficult to follow &mdash; the detail of most work branches is noise after merge. Exceptions are very large branches that do have their own relevant internal history, but I merged such a branch only once in four years with Lwt (as I recall), only after cleaning up its history, and because it made more sense than rebasing in that particular instance.

I would just squash-merge this whole PR (and practically every other PR) with GitHub's squash button. It eases the burden on maintainer and contributor of trying to keep the PR's internal history neat, allowing more room for messy work :)

raphael-proust:
@c-cube I just pushed a little æsthetics commit so that the new code has a similar layout as the existing code. It's mostly adding space here and there.

It's ready for merge AFAIAC. Let me know if it's good for you and I'll push the button.

c-cube:
I think it's good to go, yes! Glad we added some more tests :-)

PR #815, opened 2020-11-14
add ppx_let support for Lwt_result

anuragsoni:
This adds the module needed for ppx_let to work for Lwt_result.
Similar module already exists for the regular `Lwt` module.
In addition, this change also enables support for using `ppx_let`
while using module qualifier to the `bind/map` operation, ex:
`let%bind.Lwt` or `let%map.Lwt_result`.

### DIFF
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -3133,6 +3133,17 @@ struct
   let (<&>) p p' = join [p; p']
   let (<?>) p p' = choose [p; p']
 
+  module Let_syntax =
+  struct
+    let return = return
+    let map t ~f = map f t
+    let bind t ~f = bind t f
+    let both = both
+
+    module Open_on_rhs =
+    struct
+    end
+  end
 end
 include Infix
 
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -3133,6 +3133,11 @@ struct
   let (<&>) p p' = join [p; p']
   let (<?>) p p' = choose [p; p']
 
+end
+include Infix
+
+module Let_syntax =
+struct
   module Let_syntax =
   struct
     let return = return
@@ -3145,7 +3150,6 @@ struct
     end
   end
 end
-include Infix
 
 module Syntax =
 struct
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -3122,31 +3122,6 @@ struct
 end
 include Miscellaneous
 
-
-
-module Infix =
-struct
-  let (>>=) = bind
-  let (=<<) f p = bind p f
-  let (>|=) p f = map f p
-  let (=|<) = map
-  let (<&>) p p' = join [p; p']
-  let (<?>) p p' = choose [p; p']
-
-  module Let_syntax =
-  struct
-    let return = return
-    let map t ~f = map f t
-    let bind t ~f = bind t f
-    let both = both
-
-    module Open_on_rhs =
-    struct
-    end
-  end
-end
-include Infix
-
 module Let_syntax =
 struct
   module Let_syntax =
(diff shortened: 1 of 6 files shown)


anuragsoni:
This is a breaking change. There was ppx_let support for `Lwt` module, but it was missing from `Lwt_result`. I also took the opportunity to align the module structure with what's done in Base/Async etc so we can use `let%bind.Lwt` etc as module qualifier for the ppx.

That said, if you'd prefer not to support this in Lwt that'd be fine with me too, as I don't mind adding a similar module on my end :)

src/core/lwt.ml
```diff
@@ -3133,6 +3133,11 @@ struct
   let (<&>) p p' = join [p; p']
   let (<?>) p p' = choose [p; p']
 
+end
+include Infix
+
+module Let_syntax =
```
anuragsoni, line 3139: I made this change as when i try to use the ppx in the form `let%bind.Lwt`, it looks for a module `Lwt.Let_syntax.Let_syntax`. Allowing for `let%bind.<module>` helps when mixing/matching different monads.

aantron: Could you briefly explain/link us to why ppx_let expects duplicate nesting like this? I understand it's likely intentional, but it seems like some sort of bug.

src/core/lwt.mli
```diff
@@ -1377,10 +1377,14 @@ let () =
...
-  (** This module provides support for {{:https://github.com/janestreet/ppx_let}
-      ppx_let}.
+end
+
+(** This module provides support for {{:https://github.com/janestreet/ppx_let}
+    ppx_let}.
 
-      @since 4.2.0 *)
+    @since 4.2.0 *)
+module Let_syntax :
```
aantron, line 1386: Could you avoid a breaking change by just creating a new module `Let_syntax.Let_syntax`, without *moving* the existing `Syntax.Let_syntax`? That is, keep `Syntax.Let_syntax` around.

If it has to be a breaking change, I recommend splitting this PR into the `Lwt_result` part and the breaking part, because the breaking change will likely take much longer to merge (for a major release).

anuragsoni: > Could you avoid a breaking change by just creating a new module Let_syntax.Let_syntax, without moving the existing Syntax.Let_syntax? That is, keep Syntax.Let_syntax around.

That's a good point. I've updated the PR keeping the old `Let_syntax` module in place.

src/core/lwt.ml
anuragsoni: > Could you briefly explain/link us to why ppx_let expects duplicate nesting like this? I understand it's likely intentional, but it seems like some sort of bug.

I tried to match the monad interface defined in the base library (which is what i believe `ppx_let` targets).

https://github.com/janestreet/base/blob/7fb12eef1e8eb23856f3337e47dbe7522c1955dd/src/monad_intf.ml#L44-L59

I'm not too familiar with the ppx implementation, but when I tried to use the `ppx_let` with module resolution via `let%bind.Lwt`, it complained about not finding a module `Let_syntax.Let_syntax` when i didn't have this nesting. I do agree it looks a little strange. I think i'll be fine with not having this nesting as the only thing that'll not work is mixing monads and using module resolution to tell the ppx about which one to use:

```ocaml
let%bind.Lwt x = x in
let%bind.Lwt_result y = y
...
```

src/core/lwt.mli
anuragsoni: I'll need to read up on how to achive this without causing an error on ocaml < 4.08

```
Error: Multiple definition of the module name Let_syntax.

       Names must be unique in a given structure or signature.
```

anuragsoni: Looks like my newest commit fixes the build error on older compiler versions

aantron:
I want to wait a bit for comments on https://github.com/janestreet/ppx_let/issues/11 before merging this.

anuragsoni:
> I want to wait a bit for comments on [janestreet/ppx_let#11](https://github.com/janestreet/ppx_let/issues/11) before merging this.

Sounds good! Thanks for opening the upstream issue.

aantron:
Thank you!

PR #816, opened 2020-11-14
Skip "bind: Unix domain" on Hurd

### LINKED ISSUE #802: Failing "bind: Unix domain" test on GNU/Hurd
After the build fix done with #801, now the left issue to compile and test lwt successfully on GNU/Hurd is the "bind: Unix domain" test in `test/unix/test_lwt_unix.cppo.ml`.
The problem is that, because of the way the socket server (i.e. the userspace service providing Unix sockets) and the glibc implementation of sockets are, at the moment `getsockname()` on Unix sockets cannot provide the path of the bound sockets. Because of this, the end of the test
```ocaml
      try
        Lwt.return
          (chosen_path = String.sub actual_path 0 (String.length chosen_path))
      with Invalid_argument _ -> Lwt.return_false);
```
fails because `actual_path` is an empty string.
I can open a PR to fix/workaround it; however, since there are different ways to do that, I wanted to know what to do in a way acceptable for the project. Possible options are:
1. do nothing, GNU/Hurd is too niche to handle it -- understandable
2. skip the whole test on GNU/Hurd -- considering that half of the test is about checking the bound socket path, can be an option
3. skip the path comparison on GNU/hurd only
4. `Lwt.return_true` if the path is empty, on any OS
5. `Lwt.return_true` if the path is empty, on GNU/Hurd only
6. something else I'm missing

Also, in case the chosen option is to do something on GNU/Hurd only, how should the OS be checked? Neither `Sys` or `Unix` provide `uname` or an equivalent, `ExtUnix` is not used, and there is no local `uname` binding/equivalent. Simply run `uname` via `Lwt_process`?

pinotree:
At the moment the Hurd cannot provide the name (i.e. the path) of a connected Unix socket, and thus getsockname always returns an empty string. Hence skip this test on Hurd for now, as the path comparison does not work, while the test (and the APIs underneath) works fine otherwise.

Fixes #802

### DIFF
--- a/test/unix/test_lwt_unix.cppo.ml
+++ b/test/unix/test_lwt_unix.cppo.ml
@@ -929,7 +929,13 @@ let send_recv_msg_tests = [
 
 let bind_tests_address = Unix.(ADDR_INET (inet_addr_loopback, 56100))
 
-let bind_tests = [
+let bind_tests =
+  let directory_exists dir =
+    try Sys.is_directory dir
+    with Sys_error _ -> false
+  in
+
+  [
   test "bind: basic"
     (fun () ->
       let socket = Lwt_unix.(socket PF_INET SOCK_STREAM 0) in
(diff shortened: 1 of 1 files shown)


raphael-proust:
The code looks good. I'll run some local tests and such and merge it.

Issue #819, opened 2020-11-20
Dont_wait instead of `ignore_result`


francoisthire:
In the Tezos codebase, we are using the function `dont_wait` instead of `ignore_result` which forces the user to handle errors:

```ocaml
let dont_wait handler f =
  let open Lwt in
  let p = apply f () in
  on_failure p handler
```

Is it welcome to have such function in `lwt.ml`? And such function could it be recommended as an alternative to `Lwt.ignore_result`? This alternative was implemented by @raphael-proust .

aantron:
This seems like a nice and less error-prone alternative to `Lwt.async`. Note that `Lwt.ignore_result` has been not recommended for a long time, in favor of `Lwt.async`, but `dont_wait` may be even better.

PR #820, opened 2020-11-21
Lwt.dont_wait: a more local, more explicit Lwt.async

raphael-proust:
Fixes https://github.com/ocsigen/lwt/issues/819

This is an alternative to `async`/`ignore_result` with a more explicit exception handler.

### DIFF
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -824,7 +824,8 @@ val async : (unit -> unit t) -> unit
     [!]{!Lwt.async_exception_hook}.
 
     [!]{!Lwt.async_exception_hook} typically prints an error message and
-    terminates the program.
+    terminates the program. If you need a similar behaviour with a different
+    exception handler, you can use {!Lwt.dont_wait}.
 
     [Lwt.async] is misleadingly named. Itself, it has nothing to do with
     asynchronous execution. It's actually a safety function for making Lwt
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -796,6 +796,25 @@ val try_bind : (unit -> 'a t) -> ('a -> 'b t) -> (exn -> 'b t) -> 'b t
       performing any operation on one is equivalent to performing it on the
       other. *)
 
+val dont_wait : (unit -> unit t) -> (exn -> unit) -> unit
+(** [Lwt.dont_wait f handler] applies [f ()], which returns a promise, and then
+    makes it so that if the promise is {{: #TYPEt} {e rejected}}, the exception
+    is passed to [handler].
+
+    In addition, if [f ()] raises an exception, it is also passed to [handler].
+
+    As the name implies, [dont_wait (fun () -> <e>) handler] is a way to
+    evaluate the expression [<e>] (which typically has asynchronous
+    side-effects) {e without waiting} for the resolution of the promise [<e>]
+    evaluates to.
+
+    [dont_wait] is meant as an alternative to {!async} with a local, explicit,
+    predictable exception handler.
+
+    Note that [dont_wait f h] causes [f ()] to be evaluated immediately.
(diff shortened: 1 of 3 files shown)


test/core/test_lwt.ml
```diff
@@ -1801,7 +1801,56 @@ let async_tests = suite "async" [
...
+    let p, r = Lwt.wait () in
+    Lwt.dont_wait
+      (fun () -> p)
+      (fun exn -> saw := Some exn) ;
+    Lwt.wakeup_exn r Exception;
+    later (fun () ->
+      !saw = Some Exception)
+  end;
+]
+let suites = suites @ [async_tests ; dont_wait_tests]
```
aantron, line 1853: I recommend not accumulating two suites on one line, i.e. leave old line 1804 alone. The idea is to do the concatenation "locally" to each test suite, so that the concatenations don't get lost, and we don't accidentally stop running some tests when code is edited.

test/core/test_lwt.ml
```diff
@@ -1801,7 +1801,56 @@ let async_tests = suite "async" [
...
 ]
-let suites = suites @ [async_tests]
+let dont_wait_tests = suite "dont_wait" [
+  test "fulfilled" begin fun () ->
+    let f_ran = ref false in
+    Lwt.dont_wait (fun () -> f_ran := true; Lwt.return ()) (fun _ -> ());
+    later (fun () -> !f_ran = true)
+  end;
+
+  test ~sequential:true "f raises" begin fun () ->
```
aantron, line 1811: This test doesn't need to be sequential. That's probably a leftover from using `async` tests as a template, where the test corresponding to this one mutates the global `async_exception_hook` ref. This doesn't happen in `dont_wait` tests.

aantron: The same goes for the other tests in this suite with `~sequential:true`.

aantron:
> #819

Change this to "Resolves #819" or "Fixes #819" to automatically close #819 when this PR is merged. Just listing the issue number only links to the issue. Furthermore, "Fixes #819" or similar changes the message GitHub prints in #819 for the back-link, making it clearer that this PR did close that issue to readers of #819.

aantron:
LGTM modulo test suite nits.

src/core/lwt.mli
```diff
@@ -796,6 +796,25 @@ val try_bind : (unit -> 'a t) -> ('a -> 'b t) -> (exn -> 'b t) -> 'b t
...
+    side-effects) {e without waiting} for the resolution of the promise [<e>]
+    evaluates to.
+
+    [dont_wait] is meant as an alternative to {!async} with a local, explicit,
+    predictable exception handler.
+
+    Note that [dont_wait f h] causes [f ()] to be evaluated immediately.
+    Consequently, the non-yielding/non-pausing prefix of the body of [f] is
+    evaluated immediately. *)
+
```
francoisthire, line 817: Would it be nice to modify documentation of `async` and/ore `ignore_result` to propose `dont_wait` as a safer alternative to use?

raphael-proust: I added a mention of `dont_wait` in the documentation of `async`.

Because `async` and `ignore_result` are very similar (and thus it makes sense to point `ignore_result` specifically to `async`) I haven't added the mention in the documentation of `ignore_result`.

I also don't necessarily want to say that `dont_wait` is "safer" than `async`. There are very valid uses of `async`. (We departed from it in the Tezos code-base because the code-base and development team are large and we are trying to go for more explicit error/exception management. But for smaller projects it can make sense to use `async`.)

francoisthire: I am fine with that, thanks!

aantron (approved):

raphael-proust:
Rebased on master to check the CI. I'll squash-and-merge if it all goes well.

PR #824, opened 2020-11-26
allow reading/writing from Lwt_bytes using Lwt_io

anuragsoni:
The name for these functions could be better (open to suggestions). 
This can be useful when working with libraries that perform reads/writes using bigstrings


aantron:
I suggest `_bigstring` rather than `_lwt_bytes` as the suffix for these functions.

aantron:
Thank you!

PR #826, opened 2020-12-09
Fix OCaml 4.12 support (port breaking change https://github.com/ocaml/ocaml/pull/10073)

kit-ty-kate:
Port https://github.com/ocaml/ocaml/pull/10073


aantron:
Thanks!

PR #827, opened 2020-12-10
Tiniest fix in documentation

raphael-proust:
A tiny fix, a typo fix really, for the documentation of `finalize`.

And because I was there I also reworded a bit for uniformity.


aantron (approved):

aantron:
For the remaining CI, you can probably just remove `bisect_ppx` as a dev dependency (so, remove it completely) from all the `.opam` files. We converted the PPX to ppxlib in https://github.com/ocsigen/lwt/pull/807, but Bisect's conversion is still going on in https://github.com/aantron/bisect_ppx/pull/327.

aantron:
I would suggest a different PR for that. or just a push to `master`. Also, to add to an earlier comment of mine about squash-merging, this PR is a good candidate for a rebase-merge, since the individual commits can stand on their own in a clean history.

PR #831, opened 2020-12-17
Retry on ECONNABORTED in Lwt_unix.accept

### LINKED ISSUE #829: Lwt_io.establish_server_generic doesn't handle ECONNABORTED

reynir:
This was the change suggested by @hannesm in https://github.com/ocsigen/lwt/pull/830

It is unclear if anyone would want to handle this error differently.

In `Lwt_unix.accept_n` handles `Retry` exception and a comment there says `(* Ignore blocking errors if we have at least one file-descriptor: *)`. I am not sure `ECONNABORTED` is a blocking error. It's probably not a big deal though(?)

This fixed #829 as well.

### DIFF
--- a/src/unix/lwt_unix.cppo.ml
+++ b/src/unix/lwt_unix.cppo.ml
@@ -1694,12 +1694,15 @@ external accept4 :
     Unix.file_descr * Unix.sockaddr = "lwt_unix_accept4"
 
 let accept_and_set_nonblock ch_fd =
-  if Lwt_config._HAVE_ACCEPT4 then
-    let (fd, addr) = accept4 ~close_on_exec:false ~nonblock:true ch_fd in
-    (mk_ch ~blocking:false ~set_flags:false fd, addr)
-  else
-    let (fd, addr) = Unix.accept ch_fd in
-    (mk_ch ~blocking:false fd, addr)
+  try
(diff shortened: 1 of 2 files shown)


aantron:
The "blocking errors" mentioned in this comment are errors that mean non-blocking I/O was attempted, but could not be done at the moment (so I/O can only be done "immediately" from the point of view of the current thread by *blocking* the thread, hence `EWOULDBLOCK` as one of the error codes).

`ECONNABORTED` has nothing to do with blocking or non-blocking I/O, and is an error related to real external events, so an error that the user should be able to know about. It's not the result of buffer fill levels in the kernel or the like, that Lwt itself should respond to intenrally as part of how it does I/O. `ECONNABORTED` should not be handled here, and perhaps not anywhere in Lwt, though maybe it can be handled in `establish_server`. See https://github.com/ocsigen/lwt/pull/830#issuecomment-751886006.

PR #832, opened 2020-12-21
Use Github Actions for testing

patricoferris:
This PR is a simple testing workflow using Github Actions -- it is meant as a starting point, for example you may wish to add more OCaml versions, not test ubuntu, add caching etc.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,27 @@
+name: Tests for lwt_react, lwt_ppx, lwt_luv and lwt
+on: [push, pull_request]
+jobs:
+  test:
+    strategy:
+      matrix:
+        operating-system: [macos-latest, ubuntu-latest, windows-latest]
+        ocaml-version: [4.11.0, 4.10.0, 4.09.1]
+    runs-on: ${{ matrix.operating-system }}
+    steps:
+    - uses: actions/checkout@v2
+    - uses: avsm/setup-ocaml@v1
+      with:
+        ocaml-version: ${{ matrix.ocaml-version }}
+    - name: Pinning Package
+      run: opam pin add -yn lwt_react.dev './' && opam pin add -yn lwt_ppx.dev './'
+        && opam pin add -yn lwt_luv.dev './' && opam pin add -yn lwt.dev './'
+    - name: Packages
+      run: opam depext -yt lwt_react.dev lwt_ppx.dev lwt_luv.dev lwt.dev
+    - name: Dependencies
+      run: opam install -t -y . --deps-only
+    - name: Building
+      run: opam exec -- dune build
+    - name: Installing
+      run: opam exec -- dune install
+    - name: Testing
+      run: opam exec -- dune runtest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,10 +18,6 @@ jobs:
     - name: Packages
       run: opam depext -yt lwt_react.dev lwt_ppx.dev lwt.dev
     - name: Dependencies
-      run: opam install -t -y . --deps-only
-    - name: Building
-      run: opam exec -- dune build
-    - name: Installing
-      run: opam exec -- dune install
-    - name: Testing
-      run: opam exec -- dune runtest
+      run: opam install ./lwt.opam ./lwt_react.opam ./lwt_ppx.opam -yt --deps-only
+    - name: Build, Install and Test
+      run: opam exec -- dune build --only-packages lwt,lwt_react,lwt_ppx @install @runtest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,11 +1,11 @@
-name: Tests for lwt_react, lwt_ppx, lwt_luv and lwt
+name: Tests for lwt_react, lwt_ppx and lwt
 on: [push, pull_request]
 jobs:
   test:
     strategy:
       matrix:
         operating-system: [macos-latest, ubuntu-latest, windows-latest]
-        ocaml-version: [4.11.0, 4.10.0, 4.09.1]
+        ocaml-version: ['4.11.1', '4.10.2', '4.09.1']
     runs-on: ${{ matrix.operating-system }}
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


raphael-proust:
Hi,

Thank you for the PR. I'll read up on actions to get enough background to meaningfully comment on the PR. I've had a quick look and it looks good! But I need to do a little bit more reading before I merge this.

(If another reviewer knows more about this, don't hesitate to take over: I'm not acquiring a lock or anything.)

patricoferris:
Hi @raphael-proust -- sorry for the complete lack of context here! This was in collaboration with @ulrikstrid who suggested I start the PR (should have mentioned that) :)) 

Thanks both for maintaining and improving such a great library!

ulrikstrid:
I think this is a great start, the only thing we might want to do is to remove lwt_luv for now since the tests won't run on Windows currently.
This allows us to move forward and consolidate the appveyor/travis

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
+name: Tests for lwt_react, lwt_ppx, lwt_luv and lwt
+on: [push, pull_request]
+jobs:
+  test:
+    strategy:
+      matrix:
+        operating-system: [macos-latest, ubuntu-latest, windows-latest]
+        ocaml-version: [4.11.0, 4.10.0, 4.09.1]
```
avsm, line 8:
```suggestion
        ocaml-version: [4.11.1, 4.10.2, 4.09.1]
```

patricoferris: Thanks! I should be using OCaml Version in `opam-github-workflow` https://github.com/patricoferris/opam-github-workflow/issues/8

undu: I recommend quoting the versions, in some cases they might interpreted as float instead of strings.

raphael-proust:
I've read up on the actions things and I now understand that file better. I'll merge this once the comments above have been addressed.

.github/workflows/test.yml
patricoferris: Thanks for the good suggestion, have added both in latest commit and removed `lwt_luv` 👍  I'll have to work out how to force quotation in the afore-mentioned tool for generating these workflow files.

raphael-proust:
I'll merge this later this week unless someone has complains.

By the way, do we still need the travis CI too? I think it is not necessary with the new actions, right?

aantron:
It's not a question of necessary &mdash; CI is not *necessary* at all :) Here is what Travis CI currently does, that I don't think this GitHub Actions setup does.

1. Scheduled runs to proactively catch upstream changes:

    https://github.com/ocsigen/lwt/blob/d901388988c64fd637e7f2d1dc38a278f0f880fa/.travis.yml#L7

2. Rows with and without libev:

    https://github.com/ocsigen/lwt/blob/d901388988c64fd637e7f2d1dc38a278f0f880fa/.travis.yml#L19
    https://github.com/ocsigen/lwt/blob/d901388988c64fd637e7f2d1dc38a278f0f880fa/.travis.yml#L175-L180

3. Rows with musl instead of libc:

    https://github.com/ocsigen/lwt/blob/d901388988c64fd637e7f2d1dc38a278f0f880fa/.travis.yml#L21-L29
    https://github.com/ocsigen/lwt/blob/d901388988c64fd637e7f2d1dc38a278f0f880fa/.travis.yml#L189

4. Automatic update of the dev docs:

    https://github.com/ocsigen/lwt/blob/d901388988c64fd637e7f2d1dc38a278f0f880fa/.travis.yml#L129-L132

    These are then published, IIRC, by an additional, separate `.travis.yml` in a branch.

5. Some revdeps and integration tests:

    https://github.com/ocsigen/lwt/blob/d901388988c64fd637e7f2d1dc38a278f0f880fa/.travis.yml#L133-L142

Does this GitHub Actions setup do caching? If not, and there is no faster setup that GHA can be ignored in favor of, the CI will be a bottleneck if one wishes to develop something quickly and use the CI to test on all systems along the way. The Travis setup does have caching and is quite quick.

This GitHub Actions setup appears not to test on compilers older than 4.09.1, which will become a problem when we inevitably add code that assumes newer syntax or fresher library structures.

I would say it's early to disable the Travis setup. However, one could do so, at the risk of not noticing the various problems all these additional tests tend to catch.

aantron:
See also blame and issue search, which should show, for at least some of the extra tests, why they were added.

aantron:
I also missed testing bytecode-only builds, and some stress test:

https://github.com/ocsigen/lwt/blob/d901388988c64fd637e7f2d1dc38a278f0f880fa/.travis.yml#L193-L198

Having a bytecode-only build on Windows would have caught https://github.com/ocsigen/lwt/pull/833.

ulrikstrid:
The goal of this PR was to introduce GitHub Actions and then we can do a separate PR to move over more things from Travis when this is done.

patricoferris:
Hi @aantron thanks for the great feedback -- as mentioned this was just to get the ball rolling. Looking at your points GHA should be able to do most of that indeed, I thought I would add some points that might help if people want to take the GHA further. I agree it makes sense to keep the travis build whilst the transition takes place. 

#### Different Compilers

For the basic compilers (i.e. not variants) this should be fairly straight-forward by extending the `ocaml-version` variable in the matrix to however many you need. 

#### Caching and Speed

Yes, there's an action for that :)) See [odoc's example of using it](https://github.com/ocaml/odoc/blob/master/.github/workflows/build.yml#L31) -- this could be added as I'm sure things like lwt's dependencies don't change that frequently so instead of running `opam install --deps-only` it could just restore the cache between runs.

An upcoming version of `setup-ocaml` (v2) looks [very promising](https://github.com/smorimoto/irmin/runs/1635514457) at bringing the speed of compiler setup down as well and maybe even for the more obscure variants too. 

#### Musl

Should be doable too along with the byte-code version and the stress test (I'm not sure these will get the speed up from setup-ocaml@v2 so those compilers will need to be built from scratch as it currently does) -- these would just be separate concurrent jobs in the workflow.

aantron:
I understand introducing GHA gradually. I was only responding to the question about removing Travis at this point.

PR #833, opened 2021-01-07
fix LWT_NOT_AVAILABLE stub code for Lwt_unix.send_msgto

fdopen:
`lwt_unix_send_msg_byte` is missing in windows_not_available.c which causes linking errors on Windows.
While fixinig it, I've spotted wrong dummy functions for two unrelated functions.
I didn't check, if there are other functions that are not longer in sync with the unix code.


raphael-proust:
Thanks for the patch. It looks good. I'm not familiar with the Windows not-available subsystem so I'll have a bit of a read on this part of the code and come back here with (probably) an approval or (less likely) comments.

aantron (approved):

PR #834, opened 2021-01-12
lwt_ppx: carry explicit dependency from dune to opam

raphael-proust:
Thanks `opam-dune-lint` for the fix

### DIFF
--- a/lwt_ppx.opam
+++ b/lwt_ppx.opam
@@ -21,6 +21,7 @@ depends: [
   "lwt"
   "ocaml" {>= "4.02.0"}
   "ppxlib" {>= "0.16.0"}
+  "ocaml-compiler-libs" {>= "v0.12.3"}
 ]
 
 build: [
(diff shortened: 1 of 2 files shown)


aantron:
Does the code of lwt_ppx itself have this exact constraint on ocaml-compiler-libs? I assume we do directly use ocaml-compiler-libs.

raphael-proust:
The file `src/ppx/dune` lists `ocaml-compiler-libs.common` as a dependency.

I removed the dependency and it still builds. It seems the recent port to ppxlib actually removed the direct use of this library.

I've just pushed a version with the dependency removed from the dune file.

aantron:
It might still be pulled in transitively by ppxlib. Does the code of the PPX use any module from `ocaml-compiler-libs.common`? If so, it should be listed in both `dune` and `opam`, and the constraint in `opam` should be set according to what the code in the PPX requires.

avsm:
Fixed in #838

PR #836, opened 2021-01-22
Add Lwt_seq, adapted from Stdlib

zshipko:
Adds `Lwt_seq`, which mirrors `Seq` from the standard library but the callbacks return deferred values.

Let me know if there's anything missing!

### DIFF
--- a/src/core/lwt_seq.ml
+++ b/src/core/lwt_seq.ml
@@ -119,15 +119,13 @@ let rec of_seq seq =
 
 let rec to_seq seq =
   Lwt.catch
-    (fun () ->
-      seq () >|= fun x ->
-      `Result x )
+    (fun () -> seq () >>= Lwt.return_ok)
     (fun exn ->
-      Lwt.return (`Exn exn)
+      Lwt.return_error exn
   ) >>= function
-  | `Result Nil -> Lwt.return Seq.empty
-  | `Result (Cons (x, next)) ->
+  | Ok Nil -> Lwt.return Seq.empty
+  | Ok (Cons (x, next)) ->
       let+ next = to_seq next in
       fun () -> Seq.Cons (x, next)
-  | `Exn exn ->
+  | Error exn ->
       Lwt.return (fun () -> raise exn)
--- a/src/core/lwt_seq.ml
+++ b/src/core/lwt_seq.ml
@@ -109,14 +109,25 @@ let rec to_list seq =
     x :: l
 
 let rec of_seq seq =
-  match seq () with
-  | Seq.Nil -> empty
-  | Seq.Cons (x, next) ->
+  try
+    match seq () with
+    | Seq.Nil -> empty
+    | Seq.Cons (x, next) ->
       cons x (of_seq next)
+  with exn ->
+    fun () -> raise exn
 
 let rec to_seq seq =
-  seq () >>= function
-  |Nil -> Lwt.return Seq.empty
-  | Cons (x, next) ->
+  Lwt.catch
+    (fun () ->
+      seq () >|= fun x ->
+      `Result x )
+    (fun exn ->
+      Lwt.return (`Exn exn)
+  ) >>= function
+  | `Result Nil -> Lwt.return Seq.empty
+  | `Result (Cons (x, next)) ->
       let+ next = to_seq next in
       fun () -> Seq.Cons (x, next)
+  | `Exn exn ->
+      Lwt.return (fun () -> raise exn)
--- a/src/core/lwt_seq.ml
+++ b/src/core/lwt_seq.ml
@@ -119,4 +119,4 @@ let rec to_seq seq =
   |Nil -> Lwt.return Seq.empty
   | Cons (x, next) ->
       let+ next = to_seq next in
-      Seq.cons x next
+      fun () -> Seq.Cons (x, next)
--- a/src/core/lwt_seq.ml
+++ b/src/core/lwt_seq.ml
@@ -105,15 +105,12 @@ let rec of_seq seq =
   with exn ->
     fun () -> raise exn
 
-let rec to_seq seq =
-  Lwt.catch
-    (fun () -> seq () >>= Lwt.return_ok)
-    (fun exn ->
-      Lwt.return_error exn
-  ) >>= function
-  | Ok Nil -> Lwt.return Seq.empty
-  | Ok (Cons (x, next)) ->
-      let+ next = to_seq next in
-      fun () -> Seq.Cons (x, next)
-  | Error exn ->
-      Lwt.return (fun () -> raise exn)
+let rec of_seq_lwt (seq: 'a Lwt.t Seq.t): 'a t Lwt.t =
+    match seq () with
+    | Seq.Nil -> Lwt.return empty
+    | Seq.Cons (x, next) ->
+        Lwt.catch (fun () ->
+          let* x = x in
+          let+ next = of_seq_lwt next in
+          cons x next)
+        (fun exc -> Lwt.return (fun () -> raise exc))
--- a/src/core/lwt_seq.ml
+++ b/src/core/lwt_seq.ml
@@ -1,17 +1,5 @@
-(**************************************************************************)
-(*                                                                        *)
-(*                                 OCaml                                  *)
-(*                                                                        *)
-(*                 Simon Cruanes                                          *)
-(*                                                                        *)
-(*   Copyright 2017 Institut National de Recherche en Informatique et     *)
-(*     en Automatique.                                                    *)
-(*                                                                        *)
-(*   All rights reserved.  This file is distributed under the terms of    *)
-(*   the GNU Lesser General Public License version 2.1, with the          *)
-(*   special exception on linking described in the file LICENSE.          *)
-(*                                                                        *)
-(**************************************************************************)
+(* This file is part of Lwt, released under the MIT license. See LICENSE.md for
+   details, or visit https://github.com/ocsigen/lwt/blob/master/LICENSE.md. *)
 
 
 
--- a/src/core/lwt_seq.ml
+++ b/src/core/lwt_seq.ml
@@ -16,35 +16,35 @@
 
 
 open Lwt.Syntax
+open Lwt.Infix
 
 type +'a node = Nil | Cons of 'a * 'a t
 
 and 'a t = unit -> 'a node Lwt.t
 
-let empty : 'a t = fun () -> Lwt.return Nil
+let return_nil = Lwt.return Nil
+
+let empty : 'a t = fun () -> return_nil
 
 let return x : 'a t = fun () -> Lwt.return (Cons (x, empty))
 
 let cons x t () = Lwt.return (Cons (x, t))
 
 let rec append seq1 seq2 () =
-  let* seq1 = seq1 () in
-  match seq1 with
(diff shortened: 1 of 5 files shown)


raphael-proust:
Nice! I wanted to create something like that some time ago but I put it aside.

You could add `return_nil` to avoid allocations but also for completeness (see `Lwt.return_nil`, `Lwt.return_none`, etc.).

An then, the main thing that could be added is a way to map from `Seq` into `Lwt_seq`. Right now, if I'm given an `'a Stdlib.Seq.t`, how do I make an `'a Lwt_seq.t`? I think it'd involve a manual fold of the original seq. I'm not sure what the best interface would be for that.

Should it be a simple `of_seq : 'a Stdlib,Seq.t -> 'a t`?  
Or maybe it should be `of_seq_lwt : 'a Lwt.t Stdlib.Seq.t -> 'a t`?

The first approach looks straightforward to use, but the second plays well with the lazyness of `Seq.map`: `Seq.map files_in_a_dir read_a_file |> of_seq_lwt`

Ideas? Comments?

avsm:
Thanks @zshipko. Agree with @raphael-proust that using the pre-allocated `return_nil` and friends would shave off some allocations.  Aside from the Seq to Lwt_seq conversions, I wonder if we need to be careful about exception propagation as well. If an exception is raised in the middle of an Lwt_seq evaluation, it should be threaded through as an Lwt exception in the return value. This will slow it down a bit _vs_ the native one, but at least have unsurprising semantics.

src/core/lwt_seq.mli
```diff
@@ -0,0 +1,95 @@
...
+  This transformation is lazy, it only applies when the result is
+  traversed. *)
+
+val fold_left : ('a -> 'b -> 'a Lwt.t) -> 'a -> 'b t -> 'a Lwt.t
+(** Traverse the sequence from left to right, combining each element with the
+  accumulator using the given function.
+  The traversal happens immediately and will not terminate on infinite
+  sequences.
+
+  Also see {!List.fold_left} *)
```
CraigFe, line 74:
```suggestion
```

This probably works better when it's an internal stdlib reference.

src/core/lwt_seq.ml
```diff
@@ -0,0 +1,115 @@
...
+
+let empty : 'a t = fun () -> Lwt.return Nil
+
+let return x : 'a t = fun () -> Lwt.return (Cons (x, empty))
+
+let cons x t () = Lwt.return (Cons (x, t))
+
+let rec append seq1 seq2 () =
+  let* seq1 = seq1 () in
+  match seq1 with
```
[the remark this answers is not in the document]
raphael-proust: (Not that it should impact this PR, but is there a plan to have `match*` in OCaml soon?)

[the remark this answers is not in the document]
CraigFe: Relevant PR: https://github.com/ocaml/ocaml/pull/1955

zshipko:
I just pushed some new changes with `to_seq` and `of_seq` added (in addition to fixing some of the other suggestions mentioned)

I would still like to add some tests to help understand the exception situation as @avsm mentioned

raphael-proust:
Looking into it more closely, I think that the documentation should probably mention `Lwt_stream` and explain the differences. `Lwt_seq` and `Lwt_stream` provide similar abstractions. Or more precisely, it is possible to use `Lwt_stream` carefully enough that it is indistinguishable from a Seq. Failure to acknowledge so in the doc, to point out the difference, etc. will result in confusion.

I'll draft some comment block that explains it.

src/core/lwt_seq.ml
```diff
@@ -0,0 +1,131 @@
...
+(*                 Simon Cruanes                                          *)
+(*                                                                        *)
+(*   Copyright 2017 Institut National de Recherche en Informatique et     *)
+(*     en Automatique.                                                    *)
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
```
aantron, line 14: Even though nobody will bother with this realistically, we can't have a part of the code claiming to be distributed under a (modified) LGPL license in an otherwise MIT-licensed Lwt, which doesn't (and shouldn't) make an exception for it &mdash; especially if it sends people to a `LICENSE` file that isn't present in Lwt.

I think the detailed authorship of this code needs to be addressed. My (uninformed) guess is that this code was originally written by @c-cube, and modified by @zshipko for Lwt. So, we should ask @c-cube for permission to distribute this variant of it in Lwt under the MIT license, in the same way as all the other Lwt code. In particular, we don't include the names of authors in each file, but rely on blame for detailed authorship history.

If @c-cube agrees, @zshipko, please change these license headers to match the ones in the rest of Lwt. If this code is indeed substantially @c-cube's, we can also express that using a Co-authored-by field in the commit, and including a note about where the code comes form in the commit's detailed message. @raphael-proust can then make sure that is preserved during the squash.

aantron: You can probably just write all that in the message of a future commit in this PR, probably thet same one that would change the license headers, and, again, @raphael-proust can move the Co-authored-by and the history message to the right place when squashing.

raphael-proust: I'll make sure that the squashed commit includes the relevant info from the commit message.

c-cube: You can use it with the MIT license as far as I'm concerned, and there's no need to list me as an author. If you start adapting large portions of [oseq](https://github.com/c-cube/oseq/) (which you should :wink:) you can also port it with MIT but list me as a coauthor.

aantron: Thanks, @c-cube!

avsm:
If you rebase this against the `master` branch, the ocaml-ci should be green again.

zshipko:
Just wanted to check in on this and see if there's anything I can do to push it along

raphael-proust:
@zshipko I'll merge it soon (probably tomorrow). I've been a bit busy with other things so I've put the Lwt side on pause, but I'm resuming it this week.

raphael-proust:
@zshipko I pushed two small commits on top of your branch. Please have a look and check whether they are ok. Don't hesitate to rollback some/all changes: they are mere suggestions. (Although I guess the commit for the changelog is not controversial in any way.)

The MR looks generally good. I'll just open a thread to discuss `of_`/`to_seq` which is one part I think decisions can be made.

src/core/lwt_seq.ml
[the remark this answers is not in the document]
raphael-proust: I'm ok with using the `let`-operators. This is mostly a matter of style. Are there any strong opinions on the matter?

I guess the main reason to use `>>=` instead of `let*` would be to keep compatibility with old ocamls, but we're already `>= 4.8` which is when the binds where introduced.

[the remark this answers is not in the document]
CraigFe: My suggestion for removing the `let*` uses in these places was because it avoids introducing a name that is immediately matched on, but I don't have a strong opinion either way :-)

src/core/lwt_seq.mli
```diff
@@ -0,0 +1,86 @@
...
+val of_list : 'a list -> 'a t
+(** Convert a list to a sequence, preserving order. *)
+
+val to_seq : 'a t -> 'a Seq.t Lwt.t
+(** Convert from ['a Lwt_seq.t] to ['a Stdlib.Seq.t] *)
+
+val of_seq : 'a Seq.t -> 'a t
+(** Convert from ['a Stdlib.Seq.t] to ['a Lwt_seq.t].
+  This transformation is lazy, it only applies when the result is
+  traversed. *)
```
raphael-proust, line 86: I'm not sure which `of_seq` we should have: the one currently in the MR, another, both?

Options:
- current code
    - `val of_seq : 'a Seq.t -> 'a t`
- Lwt-flipping
    - `val of_seq : 'a Lwt.t Seq.t -> 'a t`

The idea behind the Lwt-flipping `of_seq` is as follows.

You can already have lwt-sequence with `'a Lwt.t Seq.t`. And as long as you are careful, you can consume the promises sequentially:

```
let rec iter_s f s =
  match s () with
  | Nil -> Lwt.return_unit
  | Cons (p, s) ->
    let* x = p in
    let* () = f x in
    iter_s f s
```

This works but you may accidentally step on your own feet and accidentally forget to wait for one promise to resolve before you start evaluating the next one.

The Lwt-flipping `of_seq` allows you to go from the valid-but-error-prone form into the Lwt-specialised version.

Maybe we need both because this Lwt-flipping `of_seq` is not as much as an inverse to `to_seq`. Maybe we can have `of_seq` as is, and we can have `inject` (or some better name) for the Lwt-flipping `of_seq`.

Ideas? Comments? Suggestions?

zshipko: The current `of_seq` implementation should probably stay as it is, but adding the Lwt-flipping version is a good idea. 

`of_seq_lwt` was mentioned above - that seems a little more clear than`inject` to me.

I will let you know when those changes are pushed.

src/core/lwt_seq.mli
```diff
@@ -0,0 +1,86 @@
...
+val to_list : 'a t -> 'a list Lwt.t
+(** Convert a sequence to a list, preserving order.
+  The traversal happens immediately and will not terminate on infinite
+  sequences. *)
+
+val of_list : 'a list -> 'a t
+(** Convert a list to a sequence, preserving order. *)
+
+val to_seq : 'a t -> 'a Seq.t Lwt.t
+(** Convert from ['a Lwt_seq.t] to ['a Stdlib.Seq.t] *)
```
CraigFe, line 81: This function is going to force all of the sequence cells in order to run the blocking actions, which is perhaps surprising given its name. We could call it [`all`](https://ocaml.janestreet.com/ocaml-core/latest/doc/async_kernel/Async_kernel/Monad_sequence/module-type-S/index.html#val-all) or similar, but personally I think it's odd to support converting this structure back to a `Seq.t` at all. I'd be in favour of removing this function altogether.

c-cube:
I think to_seq doesn't make much sense.
Maybe there could be `to_list: 'a Lwt_seq.t -> 'a list Lwt.t`, which is
clearly materializing the whole sequence; returning a "lazy" view which
is actually fully materialized seems less useful imho.

src/core/lwt_seq.mli
zshipko: That's a good point, I'm not opposed to removing `to_seq`

zshipko:
Replacing `to_seq` with `to_list` sounds good - I will make that update as well.

CraigFe:
(The PR already contains `to_list`, which I think is good :+1:)

zshipko:
Alright, just pushed a commit that removes `to_seq` and adds `of_seq_lwt`

raphael-proust:
I added a small commit that adds a tiny bit of Lwt-specific documentation: it mentions promises and such.

I think there's room to extend the tests and to factor things out there, but as a follow-up MR rather than additional commits in this one, just to make sure it progresses.

@zshipko is it good for you? If so I'll merge and make sure there's the appropriate commit message as mentioned above.

zshipko:
Thank you for adding that addition documentation, @raphael-proust - LGTM!

PR #837, opened 2021-01-22
lwt_ppx.opam: add ocaml-compiler-libs dependency

avsm:
This dependency is present in src/ppx/dune, so it explicitly adds an opam dependency as well. Found with opam-dune-lint, and should fix CI


avsm:
Dup of #834

PR #838, opened 2021-01-22
lwt_ppx: avoid dependency on compiler-libs

CraigFe:
... by replacing uses of `Ast_helper` with `Ast_builder`. Ppxlib provides a stub `Ast_helper` module that [shadows the one from `compiler-libs`](https://ocaml-ppx.github.io/ppxlib/ppxlib/Ppxlib/index.html#module-Ast_helper) and is not intended to be used directly. It's probably best to stick to a single set of AST combinators.

Alternative solution to https://github.com/ocsigen/lwt/pull/834. CC @raphael-proust.

### DIFF
--- a/src/ppx/dune
+++ b/src/ppx/dune
@@ -13,7 +13,7 @@ let () = Jbuild_plugin.V1.send @@ {|
  (public_name lwt_ppx)
  (synopsis "Lwt PPX syntax extension")
  (modules ppx_lwt)
- (libraries ocaml-compiler-libs.common ppxlib)
+ (libraries ppxlib)
  (ppx_runtime_libraries lwt)
  (kind ppx_rewriter)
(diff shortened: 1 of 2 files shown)


avsm (approved):
This looks like the right solution indeed.  We just need to manually check on 4.02 that things still work as its not checked in CI until https://github.com/ocurrent/ocaml-ci/issues/297 is fixed

raphael-proust:
Oh nice! Yes, that's a better solution. Thank you.

avsm:
The ppx is only supported on 4.04+

```
The following dependencies couldn't be met:
  - lwt_ppx → ppxlib >= 0.16.0 → ocaml >= 4.04.1
      base of this switch (use `--unlock-base' to force)
```

So no need for a test of this PR there. The CI tests are sufficient.

PR #843, opened 2021-03-04
Use Windows Unicode API in C stubs

MisterDA:
This was discussed a bit in https://github.com/ocsigen/lwt/pull/790.

I looked for windows functions in `src/unix/{,windows_c}/*.c`.

Two possible changes to this patch:

1. define the macros inside the C files;
2. use the _W_ suffix to `CreateProcess`.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -75,7 +75,7 @@ jobs:
               *.opam
               !lwt_ppx.opam
           - os: windows-latest
-            ocaml-compiler: 4.02.x
+            ocaml-compiler: 4.06.x
             libev: false
             ppx: false
             local-packages: |
--- a/CHANGES
+++ b/CHANGES
@@ -39,6 +39,7 @@
 
   * Code quality improvement: remove an uneeded Obj.magic (#844, Benoit Montagu).
 
+  * On Windows, use the Unicode API in C stubs and functions introduced in OCaml 4.06 to handle Unicode strings. Raise the minimum requirement to OCaml 4.06 (on Windows only). (#843, Antonin Décimo)
 
 ===== 5.4.2 =====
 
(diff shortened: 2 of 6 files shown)


raphael-proust:
I don't have a windows machine to try this on. Anyone has one?

Alternatively, I can merge the Github Actions PR first so we have windows in the CI.

dra27:
> 2\. use the _W_ suffix to `CreateProcess`.

This part isn't strictly necessary - defining `UNICODE` causes the Windows API to `#define` the correct function for `CreateProcess`.

MisterDA:
> 
> 
> > 2. use the _W_ suffix to `CreateProcess`.
> 
> This part isn't strictly necessary - defining `UNICODE` causes the Windows API to `#define` the correct function for `CreateProcess`.

Yes, I was suggesting it for the sake of explicitness, considering that I'm removing the comment below, in which we could also have explicitly called `CreateProcessA`.

>     /* The cast to LPSTR below is only legitimate because we are calling
>       CreateProcessA. See https://github.com/ocsigen/lwt/pull/790. */
>    if (!CreateProcess(NULL, (LPSTR)String_val(cmdline), NULL, NULL, TRUE, 0,

MisterDA:
I noticed that this patch requires OCaml >= 4.06, so changed that on Windows.
I believe it is good to go, but it means that Lwt with OCaml >= 4.02 & < 4.06 isn't buildable on Windows anymore.

MisterDA:
Sorry for all the force-pushes, it's a bit late in the day. Can confirm that this PR works now, it was nicely broken before.

smorimoto:
I just merged #896

smorimoto:
Just needs a CHANGES entry.

MisterDA:
Rebased and added an entry.

smorimoto:
Great thank you

smorimoto (approved):

PR #844, opened 2021-03-30
removed the unnecessary "ugly cast" from function add_cancel_callback

esope:
Removed an unnecessary Obj.magic

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,14 @@
+===== dev =====
+
+====== Additions ======
+
+  * Lwt_seq: a Seq-like data-structure with Lwt delayed nodes (#836, #842, Zach Shipko).
+
+====== Misc ======
+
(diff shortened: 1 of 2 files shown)


raphael-proust:
I pushed a commit, but it's more of a suggestion. The idea is to use the named type alias (`cancel_callback`) instead of the expanded definition (`unit -> unit`) in places where it is used. This is sort of complementary to the removal of the cast so I pushed it here.

If you prefer, I'll push it in its own branch and merge it through there.

esope:
This change is a good idea, indeed. Thanks. Makes things clearer. No need to add this change somewhere else and then merge it.

raphael-proust:
@esope I'll rebase this branch on master so that it goes through the newer CI system that landed there. I'll mere it once it passes that CI.

raphael-proust:
I rebased on master and resolved the conflict (which was only in the changelog)

PR #848, opened 2021-04-16
Migrate to GitHub Actions

smorimoto:
Travis ones are no longer running and worth the migration.


smorimoto:
I think we need to set an environment variable or something for Bisect.

PR #849, opened 2021-04-17
Try removing the RO bit when unlinking files on Windows

MisterDA:
On Windows, some programs may set the read-only bit on files. For
instance, git marks its index files as read-only. Calling
`Unix.unlink` on such a file will result in an
`Unix.Unix_error(Unix.EACCES, "unlink", _)` exception.

`Lwt_io.with_temp_dir` creates a temporary directory in which it is
possible that a read-only file is created. In such case, the
`delete_recursively` function will not be able to clean the temporary
directory and will also raise an exception.

This patch allows by setting the file writable (removing the RO bit)
to delete the file.

This code was copied from Dune:
https://github.com/ocaml/dune/blob/ed361ebc4f37a81d3e6ffc905b0d45f61bc17e9c/otherlibs/stdune-unstable/fpath.ml#L74-L88

### DIFF
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1471,14 +1471,20 @@ let create_temp_dir
   attempt 0
 
 let win32_unlink fn =
-  try Lwt_unix.unlink fn with
-  | Unix.Unix_error (Unix.EACCES, _, _) as e -> (
-    try
-      (* Try removing the read-only attribute *)
-      Lwt_unix.chmod fn 0o666 >>= fun () ->
-      Lwt_unix.unlink fn
-    with
-    | _ -> raise e)
+  Lwt.catch
+    (fun () -> Lwt_unix.unlink fn)
+    (function
+     | Unix.Unix_error (Unix.EACCES, _, _) ->
+       Lwt_unix.lstat fn >>= fun {st_perm; _} ->
+       (* Try removing the read-only attribute *)
+       Lwt_unix.chmod fn 0o666 >>= fun () ->
+       Lwt.catch
+         (fun () -> Lwt_unix.unlink fn)
+         (function exn ->
+            (* Restore original permissions *)
+            Lwt_unix.chmod fn st_perm >>= fun () ->
+            Lwt.fail exn)
+     | exn -> Lwt.fail exn)
 
 let unlink =
   if Sys.win32 then
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1470,6 +1470,22 @@ let create_temp_dir
   in
   attempt 0
 
+let win32_unlink fn =
+  try Lwt_unix.unlink fn with
+  | Unix.Unix_error (Unix.EACCES, _, _) as e -> (
+    try
+      (* Try removing the read-only attribute *)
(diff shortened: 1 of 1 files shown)


src/unix/lwt_io.ml
```diff
@@ -1473,21 +1473,40 @@ let create_temp_dir
...
+    | Unix.Unix_error (Unix.EACCES, _, _) as e -> (
+      try
+        (* Try removing the read-only attribute *)
+        Lwt_unix.chmod fn 0o666 >>= fun () ->
+        Lwt_unix.unlink fn
+      with
+      | _ -> raise e)
+  in
+  let unlink =
+    if Sys.win32 then
```
avsm, line 1488: You probably want to lift up the `Sys.win32` check to be in a toplevel function, as then the inliner can optimise it as a constant.  This may not happen here as you're allocating the `win32_unlink` closure just above (but need to check the lambda). Not a big optimisation given the delete file time will dominate, but it's worth keeping OS-specific functions near the top of the program logic anyway to make it more obvious what's going on.

MisterDA: Well that’s on me, I was supposing that getting the function closer to its callpoint would help inlining or branch prediction in some way.

MisterDA: Applied your suggestion, thanks!

avsm (approved):
This lgtm!

src/unix/lwt_io.ml
```diff
@@ -1470,6 +1470,22 @@ let create_temp_dir
...
 
+let win32_unlink fn =
+  try Lwt_unix.unlink fn with
+  | Unix.Unix_error (Unix.EACCES, _, _) as e -> (
+    try
+      (* Try removing the read-only attribute *)
+      Lwt_unix.chmod fn 0o666 >>= fun () ->
+      Lwt_unix.unlink fn
+    with
+    | _ -> raise e)
```
raphael-proust, line 1481: Maybe it complicates things a bit but is it worth resetting the mod to the original setting in case the second unlink fails?

It does complicate the code to do so (read stats before chmod, catch exceptions of unlink specifically,restore perms before reraising), but at least it'd be more likely to leave the file in its original state.

------------------------------

In addition, I think you should be using `Lwt.catch` to handle exceptions here. (Although I might be missing something so please don't hesitate to let me know if there's a reason that `try`-`with` is appropriate here specifically.)

MisterDA: I applied the two suggestions. The original permissions are fetched with `lstat`, if I'm not mistaken that's the correct syscall in our case.
I hadn't thought about using `Lwt.catch`, I converted the code to it.

raphael-proust: I noticed that you made a subtle change and I'm not sure if it's on purpose: You now reraise the exception from the second `unlink` whereas you previously reraised the original exception.

I don't think one is necessary better than the other (in the specific case of a system programming library for windows), but I'd rather the semantic was not chosen accidentally. Is the change intended or not?

MisterDA:
@raphael-proust Thanks for merging, but I didn't have the time to apply the change to re-raise the first exception!
https://github.com/MisterDA/lwt/commit/8ebc9cc1c312e3d61a6316e6bcbf2a51d76d786d

PR #851, opened 2021-04-20
Provide for setup and teardown hooks

kohlivarun5:
This PR is to evaluate if this is something that would be considered a reasonable addition to Lwt

The goal here is to provide for thread level setup and teardown hooks, which are run everytime the thread becomes 'active'
This is useful where LWT is run in applications where other frameworks (typically within C++) might rely on some global singletons to be active. 

In our usecase, our application relies on some enterprise logging system that runs in C++. The logging system can attach certain 'context' to each log that is emitted, as long as a global singleton is initialized. 

In the past, we have relied on having our own `bind` and `map` wrappers (`Logging.bind`) that would do the setup and teardown for you. It has the unfortunate side effect that even library code that does not care much about the global context needs to start using this specific bind and can't just depend on raw Lwt

The proposal here is to allow for clients to register thread level hooks (similar to implicit values). The hooks are expected to run whenever a thread becomes active and a callback is executed on the thread

**This PR is in draft status right now. If there is enough positive feedback, I can work on cleaning it up to make it ready**

@code-ghalib

### DIFF
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -744,18 +744,12 @@ sig
   val get : 'v key -> 'v option
   val with_value : 'v key -> 'v option -> (unit -> 'b) -> 'b
 
-  type setup = unit -> unit
-  type teardown = unit -> unit
-  val with_setup_teardown : setup:setup -> teardown:teardown -> (unit -> 'c) -> 'c
 
   (* Internal interface *)
   val current_storage : storage ref
 
-  type setup_teardown = {
-    setup : setup;
-    teardown : teardown;
-  }
-
+  type setup_teardown = unit -> (unit -> unit)
+  val with_setup_teardown : setup_teardown -> (unit -> 'c) -> 'c
   val current_setup : setup_teardown option ref
 end =
 struct
@@ -799,16 +793,6 @@ struct
 
   let current_storage = ref Storage_map.empty
 
-  type setup = unit -> unit
-  type teardown = unit -> unit
-
-  type setup_teardown = {
-    setup : setup;
-    teardown : teardown;
-  }
-
-  let current_setup : setup_teardown option ref = ref None
-
   let get key =
     if Storage_map.mem key.id !current_storage then begin
       let refresh = Storage_map.find key.id !current_storage in
@@ -840,19 +824,22 @@ struct
       current_storage := saved_storage;
       raise exn
 
-  let with_setup_teardown ~setup ~teardown (f:(unit -> 'c)) : 'c =
-    let saved_setup = !current_setup in
-    current_setup := Some { setup;teardown };
-    try
-      (* Maybe should call setup and teardown here *)
-      let res = setup () in
-      let result = f () in
-      let () = teardown res in
-      current_setup := saved_setup;
-      result
-    with exn ->
-      current_setup := saved_setup;
-      raise exn
+  type setup_teardown = unit -> (unit -> unit)
+  let current_setup : setup_teardown option ref = ref None
+
+  let with_setup_teardown setup_teardown f =
+      let saved_setup = !current_setup in
+      current_setup := Some setup_teardown;
+      try
+        let teardown = setup_teardown () in
+        (* Should we call setup/teardown here ? *)
+        let result = f () in
+        teardown ();
+        current_setup := saved_setup;
+        result
+      with exn ->
+        current_setup := saved_setup;
+        raise exn
 
 
 end
@@ -1892,17 +1879,25 @@ struct
          depend on how the user obtained [p']. *)
 
       let saved_storage = !current_storage in
-      let saved_setup : setup_teardown option = !current_setup in
+      let saved_setup = !current_setup in
 
       let callback p_result =
         match p_result with
         | Fulfilled v ->
           current_storage := saved_storage;
           current_setup := saved_setup;
-
-          let res_teardown = match saved_setup with | None -> None | Some {setup;teardown} -> Some (setup (),teardown) in
+          let teardown_opt =
+            match saved_setup with
+            | None -> 
+              None
+            | Some setup -> Some (setup ())
+          in
           let p' = try f v with exn -> fail exn in
-          let _ = match res_teardown with | None -> None | Some (res,teardown) -> Some (teardown res) in
+          let _ =
+            match teardown_opt with
+            | None -> None
+            | Some teardown -> Some (teardown ())
+          in
           let Internal p' = to_internal_promise p' in
           (* Run the user's function [f]. *)
 
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -744,19 +744,19 @@ sig
   val get : 'v key -> 'v option
   val with_value : 'v key -> 'v option -> (unit -> 'b) -> 'b
 
-  type 'x setup = unit -> 'x
-  type 'x teardown = 'x -> unit
-  val with_setup_teardown : 'x setup -> 'x teardown -> (unit -> 'c) -> 'c
+  type setup = unit -> unit
+  type teardown = unit -> unit
+  val with_setup_teardown : setup:setup -> teardown:teardown -> (unit -> 'c) -> 'c
 
   (* Internal interface *)
   val current_storage : storage ref
 
-  type 'x setup_teardown = {
-    setup : 'x setup;
-    teardown : 'x teardown;
+  type setup_teardown = {
+    setup : setup;
+    teardown : teardown;
   }
 
-  val current_setup : 'x setup_teardown list ref
+  val current_setup : setup_teardown option ref
 end =
 struct
   (* The idea behind sequence-associated storage is to preserve some values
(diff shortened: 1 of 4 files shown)


raphael-proust:
Your concrete use case is for this logging framework. Is there some other potential use case you are considering? Or even just envisioning?

Have you had a look at the deprecated "key" mechanism? It's different from your proposal, but it has enough similarity that it might be possible to encode one system in the other (most certainly possible to encode keys in your system).

Note that the "key" mechanism is deprecated because it is non-syntactic which makes its use difficult to track and its effect difficult to understand. Roughly, the key mechanism makes sense when considering threads, but not so much when considering promises. (This is a recurring theme of the more difficult and sometimes unpredictable features of Lwt.)

This points to the most specific issue that I can foresee with the integration of this feature:  
Can you explain the feature in terms of promises and binds rather than threads? (Technically Lwt doesn't have threads at all, it's just occasionally convenient to think of specific sequences of bound promises as threads; but it breaks down when considering more advanced control-flow.)

The second important consideration is: what's the performance cost for programs that do not use the feature? There's still some code to traverse even though it is effect-less code.

kohlivarun5:
Thanks for the comments @raphael-proust 

I will try to formalize the details below about my use-case (using some pseudo-code).
We have many modules that determine 'environment' information based on some c++ thread level singletons. Let's take example of the following logging module:

```ocaml
module Logger : sig 
  type context
  val initialize : attributes:(string * string) list -> context
  val cleanup : context -> unit
  
  val log_string : string -> unit
end
```

Given the above, user can do the following:

```ocaml
Logger.log_string "Before";

let context = Logger.initialize ~attributes:[("RequestName","SpecialRequest")] in
(* Some code *)
Logger.log_string "Something Happened";

let sub_context = Logger.initialize ~attributes:[("SubRegion","Critical")] in
Logger.log_string "Critical Log";
Logger.cleanup sub_context;
(* At the end *)
Logger.cleanup context;
Logger.log_string "After";
```

It will produce logs such as:

```
Before
Something Happened RequestName=SpecialRequest
Critical Log RequestName=SpecialRequest SubRegion=Critical
After
```

As you can see, the attributes are automatically attached, no matter how deep in the callstack you might be, as long as the `context` object is alive in the stack. You can have multiple contexts that can also be stacked on top of each other


Currently, we manage this by creating a custom version of `Lwt.bind` and `Lwt.map` by using the key mechanism, where we add the attributes in the key/value and create context on each bind:

```ocaml
let my_bind x f = 
  Lwt.bind x (fun x -> 
         let context = Logger.initialize (Lwt.get_value "ATTRIBUTES") in 
         let res =  f x in 
         Logger.cleanup context;
        res)
```

This is okay but not great because all code needs to be now aware of this special bind, and can't just use Lwt.bind. Sometimes, someone might miss or forget and use `Lwt.bind` and the context is not setup and logs might be incomplete.

Logs is just one concrete example, there might be other things in the context such as user sessions.

When about it in terms of promises and binds, what we are looking for is to have hooks into `map` and `bind` such that:
1) one can set a setup/teardown on a Promise (similar to how we attach key/values)
2) Whenever that promise is `chained` via map or bind, the setup and teardown are appropriately called

I agree in terms of performance, that is something I can measure and see if it causes some slowdown. My expectation would be that if a user does not set any hooks, then it is a no-op and should not impact performance. But I can try to micro-benchmark it

Hopefully the above clarifies the goals here, please let me know if anything is unclear.

code-ghalib:
> Can you explain the feature in terms of promises and binds rather than threads? (Technically Lwt doesn't have threads at all, it's just occasionally convenient to think of specific sequences of bound promises as threads; but it breaks down when considering more advanced control-flow.)

I think this is the friction in using monadic style for modeling effects. In addition to modeling asynchrony, there are plenty of other effects one needs to model in production code. We could probably also model these using monadic style:
- Potential for failure: `'a Result.t Lwt.t`
- Identify the computation to a logging mechanism: `'a Result.t Logged.t Lwt.t`
- Set up a telemetry context (identify users, collect performance, etc.): `'a Result.t Logged.t Metered.t Lwt.t`
- ...  

Each of the `bind`/`return`s would do the teardown (unwrap) and setup (wrap) hooks that are being proposed.  

But this is quite unwieldy, and adds little value in my opinion. While `Result` is an effect we want a programmer to reason about explicitly, the other effects have little impact on control-flow or the logic of the program, so enriching the types does not make much sense to me - the compiler adds no value by checking that everything is wrapped in a `Logged` context for example.  I see the proposed callback registration on scheduler context-switches as using "effect-handler" style instead of monad-style.  

We have used the deprecated `key` mechanism, but it still requires the caller having to use a different `bind` operation as described by @kohlivarun5 .
 
> The second important consideration is: what's the performance cost for programs that do not use the feature? There's still some code to traverse even though it is effect-less code.

I agree we should make sure you only pay for what you use.

raphael-proust:
Thanks for the explanations, I understand what you are trying to do better now.

I don't have strong opinions on the topic, only general thoughts and questions:
- What's the impact for performance?
- What's the impact on readability for code that uses the feature? Or, maybe equivalently, how easy is it to give specifications for code that uses the feature?

I'll let others participate with their additional questions and opinions and such.

--------------------------

I also want to try to implement what you have but only using keys (and not shadowing bind). Mostly I think it'll help me understand where exactly keys become insufficient.

kohlivarun5:
Following are some metrics I obtained for the tests added in the PR:

### With setup_teardown enabled (1 million binds)

##### RUN1
```
Done perf for [pending with noop]. Time taken=86.153000 (ms)
Done perf for [pending with setup]. Time taken=83.360000 (ms)
Done perf for [pending without setup]. Time taken=76.855000 (ms)
Done perf for [already fulfilled with setup]. Time taken=89.596000 (ms)
Done perf for [already fulfilled without setup]. Time taken=81.008000 (ms)
```

##### RUN2
```
Done perf for [pending with noop]. Time taken=107.942000 (ms)
Done perf for [pending with setup]. Time taken=92.081000 (ms)
Done perf for [pending without setup]. Time taken=100.504000 (ms)
Done perf for [already fulfilled with setup]. Time taken=93.724000 (ms)
Done perf for [already fulfilled without setup]. Time taken=86.305000 (ms)
```

### With setup_teardown code removed (as in master) (1 million binds)

##### RUN1
```
Done perf for [pending with noop]. Time taken=97.392000 (ms)
Done perf for [pending with setup]. Time taken=88.960000 (ms)
Done perf for [pending without setup]. Time taken=83.008000 (ms)
Done perf for [already fulfilled with setup]. Time taken=83.150000 (ms)
Done perf for [already fulfilled without setup]. Time taken=83.299000 (ms)
```

##### RUN2
```
Done perf for [pending with noop]. Time taken=91.551000 (ms)
Done perf for [pending with setup]. Time taken=89.912000 (ms)
Done perf for [pending without setup]. Time taken=82.882000 (ms)
Done perf for [already fulfilled with setup]. Time taken=79.039000 (ms)
Done perf for [already fulfilled without setup]. Time taken=88.617000 (ms)
```

As you can see, there is a variability in the results, which I would attribute to machine/cpu variability. 
Please let me know what you think of the above results, or if there are alternate tests you would like to add/measure

src/core/lwt.ml
```diff
@@ -818,6 +823,25 @@ struct
     with exn ->
       current_storage := saved_storage;
       raise exn
+
+  type setup_teardown = unit -> (unit -> unit)
```
code-ghalib, line 827: I realize this is how `with_key` is done, to get around the value restriction, but I prefer the user being oblivious to that technicality, and that the user-facing type be more natural. Does the following code work without imposing any additional overhead?
```ocaml
type 'a context = {
  setup : unit -> 'a;
  teardown : 'a -> unit;
}

type private_context = unit -> unit -> unit
  
let cur_context : private_context option ref = ref None
    
let to_private {setup;teardown} = fun () -> let x = setup () in fun () -> teardown x
  
let with_context c f =
  let saved = !cur_context in
  let setup_teardown = to_private c in
  let _ = cur_context := Some setup_teardown in
  let teardown = setup_teardown () in
  try
    let result = f () in
    let _ = teardown () in
    let _ = cur_context := saved in
    result
  with exn ->
    let _ = teardown () in
    let _ = cur_context := saved in
    raise exn
```

Also wondering whether it makes sense to have:
```ocaml
type 'a context = {
  setup : unit -> 'a;
  teardown : 'a -> unit;
  teardown_exn : 'a -> exn -> unit;
}
(* or *)
type ('a,'b) context = {
  setup : unit -> 'a;
  teardown : 'b state -> 'a -> unit;
}
```
To allow to teardown differently based on whether promise resolved or not.

code-ghalib: Also, do we need either special implementation or a new `with_context` like function for `Lwt_stream.t` or will that just work?

kohlivarun5: Interface updated to 

```ocaml
val with_setup_teardown : setup:(unit -> 'a) -> teardown:('a -> unit) -> (unit -> 'c) -> 'c
```

I will review your comment about Lwt_stream.t

code-ghalib:
fwiw, this logging thing seems to be a common need - came across this today:
https://github.com/ahrefs/devkit/blob/master/lwt_mark.mli
This solution is also based on deprecated `key` mechanism.

src/core/lwt.ml
kohlivarun5: I tested and if the stream map etc are done within a setup_teardown, the same functions are carried for streams

kohlivarun5:
@raphael-proust This should be ready for a review now. Please let me know if anything can be added here

I see `ocaml-ci` check is failing with `Library ppx_let not found`, which I think is an issue with other PRs aswell

PR #852, opened 2021-04-21
Fix Lwt_fmt.stderr to actually point to stderr instead of stdout (#850)

### LINKED ISSUE #850: Lwt_fmt.eprintf prints to stdout instead of stderr
The following test program:

```ocaml
let () =
  Lwt_main.run @@
  Lwt_fmt.eprintf "Test\n%!"
```

which is supposed to print into `stderr` the string "Test", followed by a newline, followed by flushing the stream, behaves as follows on the command line:

```sh
$ ./test >/dev/null
$ ./test 2>/dev/null
Test
```

So it prints into `stdout` instead of `stderr`.

The expected behavior would have been:

```sh
$ ./test >/dev/null
Test
$ ./test 2>/dev/null
```

vog:
Fix Lwt_fmt.stderr to actually point to stderr instead of stdout.

This fixes #850.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,10 @@
 
   * Lwt_seq: a Seq-like data-structure with Lwt delayed nodes (#836, Zach Shipko).
 
+====== Bugfixes ======
+
+  * Fix Lwt_fmt.stderr to actually point to stderr (#852, #850, Volker Diels-Grabsch).
+
 ===== 5.4.0 (2020-12-16) =====
 
 ====== Installability ======
--- a/src/unix/lwt_fmt.ml
+++ b/src/unix/lwt_fmt.ml
@@ -76,7 +76,7 @@ let ifprintf ppft fmt =
   ikfprintf (fun _ t -> t) ppft fmt
 
 let stdout = of_channel Lwt_io.stdout
-let stderr = of_channel Lwt_io.stdout
+let stderr = of_channel Lwt_io.stderr
 
 let printf fmt = fprintf stdout fmt
 let eprintf fmt = fprintf stderr fmt


Drup (approved):
:+1:

(I'm slightly ashamed 🥺)

Maybe add a test ?

raphael-proust:
> Maybe add a test ?

A test might be good, but it's not obvious how to add it. The only tests that touch stdout/stderr are for `Lwt_io` and they are tests for servers and connections.

I'd be ok merging this without a test. I'll leave this open for now so the discussion can continue for a bit on that topic.

raphael-proust:
If you have the time, can you add the following diff? (I couldn't push to your repo/branch directly.) If not it's ok, I'll add it when doing the release.

```
diff --git a/CHANGES b/CHANGES
index 67f9ad454..aa7540df0 100644
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,10 @@

   * Lwt_seq: a Seq-like data-structure with Lwt delayed nodes (#836, Zach Shipko).

+====== Bugfixes ======
+
+  * Fix Lwt_fmt.stderr to actually point to stderr (#852, #850, Volker Diels-Grabsch).
+
 ===== 5.4.0 (2020-12-16) =====

 ====== Installability ======
```

vog:
Will there be an immediate bugfix release of Lwt as soon as this fix ran successfully through CI/CD, or do you collect for fixes before releasing? In the latter case: When will the schedule for the next bugfix release?

raphael-proust:
I'm tempted to make a bugfix release for that. I'll check with the other maintainers and come back to you.

avsm:
Agreed, let's do a bugfix release for this.

raphael-proust:
I'll make the release this week.

PR #853, opened 2021-04-23
Add bisect step to GitHub Actions workflow

smorimoto:


raphael-proust:
I'm not sure I'm reading this right, but it seems that jobs are duplicated: they are run once for the push event and one because we are in ta pull request.

I was checking the errors in the checks and they are the same job for different triggers (`pull_request` and `push`). Interestingly, they fail for different reasons (PR fails with `Error: expected file 'src/core/lwt_condition.ml' is not included in the report`, push fails with false tests for `sleep`, `timeout`, `with_timeout`, and `auto_yield`).

(Side note: the timeout failures are a bit strange: they fail because the promise scheduled to be resolved by the OS clock are woken up "too early". Not sure how this happens. The differences are pretty small so they might be due to clock precision or something along those lines.)

Are the tests repeated? Is there a way in Github actions to run the `push` actions only if there is no PR action or vice versa?

smorimoto:
It looks like a lot of the problems have been resolved, but at this point there is no reason to spend much time on this, so I will close this.

Issue #855, opened 2021-05-20
Deprecate Lwt_main.yield and Lwt_unix.yield?


favonia:
This is a suggestion. After talking to @aantron on Discourse, I feel we should officially deprecate `yield` and related functions using `[@@alert deprecate]`, in hope that one day we will merge the two pools into one.

cyberhuman:
How do you yield then? 🤔

favonia:
`Lwt.pause ()` will do the job.

aantron:
See

https://github.com/ocsigen/lwt/blob/9cbf3a9c21f181dcd78dabf2117f87a62a2d65cc/src/unix/lwt_main.mli#L45-L52

and https://discuss.ocaml.org/t/lwt-pause-versus-yield/7875.

raphael-proust:
See https://github.com/ocsigen/lwt/pull/784 for some earlier discussion of this.

TL;DR:
- `Lwt_main.run` treats paused and yielded promises differently: it resolves paused promises twice as often as yielded promises.
- This is not an intended difference, but it's a significant difference and it may be significant for several binaries.

---------------------------------------------

I think that, before we release this we should:
- run the tests of all the opam packages that depend on lwt with the modified lwt (I think this is possible through the opam CI in some way)
- notify packages that we break or even give them some patches
- advertise the possibility of the change on discuss (and maybe other channels)

Then if all of that goes well, we can release. It'll need a "breaking change" mention in the changelog.


Maybe we could leverage the opam infrastructure to test if we break some of the tests of the opam packages. I think we should do this (if possible), make upstream commits to all possibly broken packages, and we should make a special mention of this in the changelog (and advertise it in advance on discuss too).

favonia:
I just updated `CHANGES` but need help in all other actionable items.

raphael-proust:
I'll ask some opam wizards about the general test of all dependencies.

favonia:
Closing this as #917 will complete every actionable items in this issue.

PR #858, opened 2021-05-23
Deprecate Lwt_main.yield in favor of Lwt.pause.

favonia:
This partially addresses #855. The eventual goal is to remove the second pool completely.

`Lwt_unix.auto_pause` was added to replace `Lwt_unix.auto_yield`.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,8 +1,13 @@
 ===== dev =====
 
+====== Deprecations ======
+
+  * Lwt_main.yield and Lwt_unix.yield are deprecated in favor of the generic Lwt.pause, and Lwt_unix.auto_yield is deprecated in favor of the new Lwt_unix.auto_pause (#855, #858, Favonia). Currently, Lwt.main resolves paused promises more frequently than yielded promises. The difference is unintended but existing applications could depend on it. Once the migration has been completed, Lwt_main.abandon_yielded_and_paused will also be deprecated in favor of Lwt.abandon_paused.
+
 ====== Additions ======
 
   * Lwt_seq: a Seq-like data-structure with Lwt delayed nodes (#836, #842, Zach Shipko).
+  * Lwt_unix.auto_pause: the replacement of Lwt_unix.auto_yield that uses Lwt.pause instead of Lwt_unix.yield (#855, #858, Favonia).
 
 ====== Misc ======
 
--- a/CHANGES
+++ b/CHANGES
@@ -2,11 +2,12 @@
 
 ====== Deprecations ======
 
-  * Lwt_main.yield and Lwt_unix.yield are deprecated in favor of the generic Lwt.pause, and Lwt_unix.auto_yield is deprecated in favor of the new Lwt_unix.auto_pause (#855, #858, Favonia). Currently, Lwt.main resolves paused promises more frequently than yielded promises. The difference is unintended but existing applications could depend on it. Once the migration has been completed, Lwt_main.abandon_yielded_and_paused will also be deprecated in favor of Lwt.abandon_paused.
+  * Lwt_main.yield and Lwt_unix.yield are deprecated in favor of the generic Lwt.pause, and Lwt_unix.auto_yield is deprecated in favor of the new Lwt_unix.auto_pause. Currently, Lwt_main.run resolves paused promises more frequently than yielded promises; the difference is unintended but existing applications could unintentionally depend on it (#855, #858, Favonia).
 
 ====== Additions ======
 
   * Lwt_seq: a Seq-like data-structure with Lwt delayed nodes (#836, #842, Zach Shipko).
+
   * Lwt_unix.auto_pause: the replacement of Lwt_unix.auto_yield that uses Lwt.pause instead of Lwt_unix.yield (#855, #858, Favonia).
 
 ====== Misc ======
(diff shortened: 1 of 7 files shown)


src/unix/lwt_unix.cppo.ml
```diff
@@ -136,6 +136,16 @@ let auto_yield timeout =
     end else
       Lwt.return_unit
 
+let auto_pause timeout =
```
aantron, line 139: I would suggest not to add this function. It seems to me like users can just keep using `auto_yield`. `auto_pause` with timeouts is a little weird in meaning, because the whole point of `pause` is that it does not depend on `Lwt_unix`. However, `auto_pause` inherently depends on `Lwt_unix`, so why not just leave `auto_yield` only?

favonia: The difference is whether eventually you would call `pause` or `yield` when the timeout happens. If secretly changing `yield` to `pause` inside `auto_yield` is okay, then I would argue we should simply silently replacing all `yield` with `pause` without deprecation. I believe our previous discussion has ruled out silent changes, and thus we need a different function is to give the users some warnings. Thus, we cannot leave `auto_yield` only.

The remaining objection seems to be the weirdness of its meaning. I disagree with using `pause` means that it has no place in `Lwt_unix`. In the `Unix` module of the standard library, for example, there's `Unix.in_channel_of_descr` to create `in_channel`. The point of `in_channel` is to be platform-independent, but that does not prevent specific platforms or backends to provide more functionality using the common mechanism. Here, `auto_pause` is extending the common `pause` using features on the specific platform.

favonia: Also, after this PR, there would be no functions with the word `yield` in the lwt library except `Lwt_main.abandon_yielded_and_paused` (which would also be deprecated once `Lwt_main.yield` is completely gone), so it might actually be more consistent to call everything `pause`.

CHANGES
```diff
@@ -1,8 +1,15 @@
 ===== dev =====
 
+====== Breaking ======
```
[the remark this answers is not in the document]
favonia: I am neutral on whether deprecation counts as breaking changes and can work either way. :grin: I see that it could be counted as breaking as we want the users to modify their code, but I also see that it is not breaking because they can choose not to. Happy to update the PR once a consensus is reached.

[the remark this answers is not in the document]
raphael-proust: I guess it breaks the code that is compiled with warning-as-failure.

IIRC, `dune`'s release mode sets this option to false and dune-built opam packages are supposed to be built in release mode. So we shouldn't break too many packages. But I think that deprecation of such a central function will probably break some development versions of some projects.

I don't know to what extent this counts as breaking: it depends on a compilation option that is set differently by different toolchains on different occasions.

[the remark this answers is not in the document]
aantron: > IIRC, `dune`'s release mode sets this option to false and dune-built opam packages are supposed to be built in release mode. So we shouldn't break too many packages.

It's not breaking because of this. Doing this change does not affect the "main" opam-based builds, and anything similar to them. In addition, IIRC, if one uses the various vendoring stanzas of Dune, vendored subprojects are also built with warnings as warnings.

We used to routinely introduce deprecations in Lwt without a major version increment, usually three months or so before such an increment, which would then commit the actual breaking change warned in the deprecation messages.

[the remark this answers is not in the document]
favonia: It seems consensus was reached.

favonia:
Is there something left to be done?

CHANGES
```diff
@@ -1,8 +1,15 @@
 ===== dev =====
 
+====== Deprecations ======
+
+  * Lwt_main.yield and Lwt_unix.yield are deprecated in favor of the generic Lwt.pause, and Lwt_unix.auto_yield is deprecated in favor of the new Lwt_unix.auto_pause (#855, #858, Favonia).
+
+    Currently, Lwt.main resolves paused promises more frequently than yielded promises. The difference is unintended but existing applications could depend on it. Unifying the two pools of promises into one would eliminate possible discrepancies and simplify Lwt_main. As the first step, all yield-related functions are deprecated except for Lwt_main.abandon_yielded_and_paused for applications to work with existing libraries that still use yield. Once the migration has been completed, Lwt_main.abandon_yielded_and_paused will also be deprecated in favor of Lwt.abandon_paused.
```
raphael-proust, line 7: I'm don't know if the markup language used in this file (wikicreole) supports multiple-paragraph bullet-points. A conversion from that to markdown using pandoc seems to suggest it doesn't.

This could be shurnk and inlined. I'd say the two first sentences of the extra paragraph are the most important for the users of the library: it indicates the likely source of behaviour change. The rest is planning for Lwt maintainer (although it could be included in the documentation of `Lwt_main.abandon…`)

favonia: I did not know it's in Creole---thought it's just a nicely formatted plain text file. I can move the text around.

favonia:
So I rebased the PR and cleaned up the commit history a bit.

raphael-proust:
Nothing else to do. I made a small change to the CHANGES file. I'll leave enough time for you to let me know if I have messed up something and then I'll merge.

Thanks for the contribution!

favonia:
LGTM

PR #860, opened 2021-05-24
Lwt_domain: an interface to multicore parallelism

Sudha247:
**Synopsis**: The PR aims to add a module `Lwt_domain` that enables offloading computations to many CPU cores via Multicore's [Domains](https://github.com/ocaml-multicore/ocaml-multicore/blob/4.12%2Bdomains%2Beffects/stdlib/domain.mli).

### `Lwt_domain` interface

This patch aims to add a new module `Lwt_domain`. The API is similar to `Lwt_preemptive`, the major difference being `Lwt_preemptive` runs compuatations in a new systhread and `Lwt_domain` computations in a new domain. 

In effect, the `Lwt_domain` runs OCaml computations in parallel on multiple CPU cores, which is not the case with `Lwt_preemptive`. This is because system threads are run on a single CPU core despite being run concurrently. On the other hand, multicore domains run on the available cores.

### Implementation

A pool of domains is created by the `simple_init` function, with the default number of domains being `4`. The created domains will be alive and waiting for tasks till the time main loop is alive. The number of live domains can be adjusted via the `set_bounds` function.

Unlike `Lwt_preemptive` which creates new systhreads frequently, all live domains are reused here because repeatedly creating and shutting down new domains is an expensive process, which could be avoided.

When `detach` function is called it obtains a new worker from the pool of domains and the function is executed in that worker domain. After the execution is done, the worker domain will again be ready to take tasks.

A separate queue of jobs is maintained that are supposed to be executed on the main domain, being called from `run_in_main`.

### Benchmark

A simple server that returns Fibonacci number of a given input was ported to run in parallel. The server and client code can be found [here](https://github.com/Sudha247/code-samples/tree/master/lwt-parallel). The benchmark was run with 10 clients and 10 requests per client, it's time is shown below. Individual timings can be found in the blog post mentioned below. 

**Time graph of parallel Fibonacci server**

image

#### Side note

`Unix.fork` is not compatible with multiple domains. Hence, the test-suite is modified to run sequentially and all the tests that run `Unix.fork` are run before the `Lwt_domain` test.

---

This module is intended to work with OCaml 5.0 with parallelism support. It can be used now with Multicore OCaml obtained from [multicore-opam](https://github.com/ocaml-multicore/multicore-opam). Any suggestions/feedback welcome, benchmarks from potential applications of this module will be appreciated too!

### Further reading

* [Related blog post](https://sudha247.github.io/2020/10/01/lwt-multicore/)
* [Concurrency and parallelism and design notes](https://github.com/ocaml-multicore/ocaml-multicore/wiki/Concurrency-and-parallelism-design-notes)


raphael-proust:
Hello,

Thanks a lot for this PR! I'm looking forward to multicore-lwt!

I've just had a quick look. (I'll do a longer review later.) I have a few questions/comments:

- You added `Lwt_domain` to the `unix/` domain. Is there a reason for this? Is there a dependency from the multicore part of the runtime on the OS? Or is it by similarity with `Lwt_preemptive` that it ended up there? In general we want to put code in `core/` if it doesn't rely on unix.

    Currently there's a call to lwt_unix, for a notification thing. Is it for convenience (the function with the right signature was there) or intrisically dependent on a unix feature?

- We'll need to grow the test suite a bit just to buff up the coverage. But that's not urgent at all.

avsm:
Lwt_offload should be in core/ indeed -- it's not part of the Unix-specific part of multicore OCaml.

@Sudha247 it would be good to get more of your benchmarks into the test suite here. For example the fib run would be ideal to check that you get the same results and also provide an example.

I'm just discussing with caml-devel the right way to be able to release this with Lwt_offload as an optional module -- I'll update when I find consensus.

Sudha247:
Thanks for the comments @raphael-proust and @avsm!

Agree about moving this module to core. The notification call is just for convenience, I'll work on removing this part after which there won't be any dependency on lwt_unix, so then the module will be ready to be moved to core.

Having more tests for the module would certainly be useful, I'll add the fib example to the testsuite.

Sudha247:
Some updates on the latest commit are:
* Ported the `Lwt_domain` module to domainslib Task pool rather than managing creating domains on the module itself. This reduces duplication with domainslib.
* Added a couple more tests - a test for `run_in_main` and a shorter running version of the Fibonacci benchmark.

With the new update, there's a slight digression from the `Lwt_preemptive` module; since the process of creating and destroying domains is expensive, `Lwt_domain` creates a fixed number of worker domains onto which tasks will be offloaded to, so there's no notion of minimum and maximum number of domains.
### Benchmarks
Some benchmark results with the updates are below:
**Benchmark**: Computes `fib 42` 100 times.

#### Time
[Sequential version](https://github.com/Sudha247/code-samples/blob/master/lwt-parallel/fib_bench_sequential.ml): 205.397s
[Preemptive threads version](https://github.com/Sudha247/code-samples/blob/master/lwt-parallel/fib_bench_preemptive.ml): 210.665s
[Parallel version](https://github.com/Sudha247/code-samples/blob/master/lwt-parallel/fib_bench_parallel.ml) (with `Lwt_domain`)

The results were obtained on an Intel Xeon Gold 5120 processor with 24 isolated cores.

| Cores | Time(s)   | Speedup      |
|-------|--------|--------------|
| 1     | 212.72 | 0.9655744641 |
| 2     | 106.34 | 1.931512131  |
| 4     | 53.15  | 3.864477893  |
| 8     | 27.64  | 7.431150507  |
| 12    | 19.13  | 10.73690538  |
| 16    | 14.89  | 13.79429147  |
| 20    | 10.64  | 19.30422932  |
| 24    | 11.02  | 18.63856624  |

image

Note: There's still a dependency on `Lwt_unix` due to calling notifications. I'm trying to replicate that nicely with Domainslib channels. Will update the PR and move the module to core when it's done.

src/unix/lwt_domain.ml
```diff
@@ -0,0 +1,96 @@
+open Lwt.Infix
+
+module C = Domainslib.Chan
+module T = Domainslib.Task
+
+(* Maximum number of domains: *)
+let max_domains : int ref = ref 0
+
+(* Size of the waiting queue: *)
+let max_domains_queued = ref 128
```
raphael-proust, line 10: This seems to only be used in the setter and getter function. I think it's dead code.

Now, that's the start of an interesting conversation about the queue limit…

In the queue limit as described by the documenation, when the queue is full a new job would not be added and the function would "sleep". The word "sleep" is not clear in the context of Lwt because it's a legacy concept from the olden day of "thread" rather than the modern days of "promise". But considering the type I think it means that the idea is that the promise stays pending until there is space in the queue at which point there the jobs is queued at which point the promise is pending until the job is started at which point the promise is pending until the job is done.

This causes an issue in that it is impossible to tell that the queue is full, that the jobs are piling up. Essentially, the queue is supplemented by an informal queue within the scheduler. Importantly, this scheme doesn't allow to apply back-pressure to whatever part of the program is producing jobs.

There are several approach that we can take, please discuss pros and cons. I'm especially interested in opinions about the way these options interact with the multicore runtime.

- No queue limit. It can be a bit dangerous in that some part of the code might start to create an unreasonable number of jobs which ends up occupying an unreasonable amount of memory.
- Limit via option: `val detach : ('a -> 'b) -> 'a -> 'b Lwt.t option`. So `detach` returns a promise only if there is enough space to store the job that resolves that promise in the queue. It doesn't stand on its own: we need to add a function to test whether there is space and probably a function to wait for a spot `val wait_for_queueu_space : unit -> unit Lwt.t`. It's a complicated interface with non-obvious semantic.
- Limit with some other mechanism: exceptions (probably not), Result, some other such thing. Essentially the same as option.
- Weak limit but allow back-pressure via double promise: `val detach: ('a -> 'b) -> 'a -> 'b Lwt.t Lwt.t`. The value returned by `detach` resolves when it is scheduled (i.e., when it is added to the queue), and it resolves to a promise that resolves when the computation is done . It's a non-trivial interface but the semantic is cleaner.

Ideas? Opinions?

src/unix/lwt_domain.mli
```diff
@@ -0,0 +1,64 @@
...
+        [detach] evaluates to an Lwt promise, which is pending until the
+        domain completes execution.
+        [detach] calls {!simple_init} internally, which means that the number of
+        domains is capped by default at four. If you would like a
+        higher limit, call {!init} or {!set_bounds} directly.
+        Note that Lwt thread-local storage (i.e., {!Lwt.with_value}) cannot be
+        safely used from within [f]. The same goes for most of the rest of Lwt. If
+        you need to run an Lwt thread in [f], use {!run_in_main}. *)
+
+  val run_in_main : (unit -> 'a Lwt.t) -> 'a
```
raphael-proust, line 21: IIUC, if you run `run_in_main f` from a detached computation, then this computation is essentially frozen until the promise `f ()` has resolved. Is this correct understanding?

What happens to this domain in the mean-time? Can it be used by other detached computations?

Is it possible to make a dead-lock by making a promise that waits for a detached computation to finish but the computation waits on the Lwt promise to resolve?

Is it possible to run out of domains by freezing too many computations waiting for promises?

I think this function is useful, but it should be used sparingly. The documentation (I know it's not in its final form rn, just planning ahead) should have a precise description of what happens (a) to the computation, (b) to the domain that the computation runs on, (c) to the promise that's created. In addition, depending on the answer to some of the questions above, the documentation may also need to include warnings, usage notes, and such.

raphael-proust:
This module is very promising (pun intended).

I'll work on packaging: how to ship this module without introducing a hard dependence on OCaml.5.0. I think it'll just need to be a separate opam package `lwt-domain` or something of the sort.

Sudha247:
Thanks a lot for the comments @raphael-proust! I'm doing another pass to update documentation and remove obsolete parts in the PR. I'll think through some of these open-ended questions and respond to them.

avsm:
It looks like there is a real error here in the CI:

```
File "src/unix/lwt_domain.ml", line 24, characters 42-43:
24 | let pool = ref (T.setup_pool ~num_domains:0)
                                               ^
Error: The function applied to this argument has type
         num_additional_domains:int -> Domainslib__Task.pool
```

Is this expected @Sudha247?

Sudha247:
Needed to bump domainslib to the latest version and change the argument to `~num_additional_domains`, it's fixed now.

src/unix/lwt_domain.ml
Sudha247: Indeed this is an interesting conversation! I'll try to highlight some of the pros and cons for each of these proposals.

* No queue limit: No extra work for the programmer, whilst it is possible that large number of jobs end in a stack overflow. This can potentially be avoided by limiting the number of detach operations in the program. Although, it would be useful to preemptively guard against this before an application runs out of memory.

* Option values: The interface is a bit more complicated than the no limit option. I think we can track the number of jobs that are live with a simple atomic counter and deny new entries when it reaches the cap limit.

* Other mechanisms: Exception could work if exceeding queue limit is not expected to happen very often and we'd like to treat it like a fatal error. Furthermore, this will increase the complexity of working around the queue limit in applications. I think using Result is in principle similar to the Option mechanism described above.

* Double promise: Agreed that the interface is non-trivial, but it might be more user-friendly. Only concern I had about this is; would it be possible that an active domain gets blocked waiting for the first promise to resolve?

Regarding the queue size, realistically speaking the queue limit depends on individual jobs' workloads. I suppose it's fine to go with the same queue limit 128 for now - 128 is not too small that we hit the limit too often, is neither too big that we run out of memory frequently.

If this list is any indication, I'm leaning towards Option and double promise, but if people have strong opinions on this or preferences towards other options we've considered above do let us know.

src/unix/lwt_domain.mli
Sudha247: I concur that `run_in_main` needs to be used only sparingly, if at all it is.

>IIUC, if you run run_in_main f from a detached computation, then this computation is essentially frozen until the promise f () has resolved. Is this correct understanding?

Yes that's correct, the calling domain is frozen until the computation is resolved. Other detached computations cannot use that domain since it's blocked, waiting for the promise f () to resolve. This is due to the fact that the function returns the final result and not the promise.

>Is it possible to make a dead-lock by making a promise that waits for a detached computation to finish but the computation waits on the Lwt promise to resolve?

Yes I'd imagine that's possible, and it's not ideal.

>Is it possible to run out of domains by freezing too many computations waiting for promises?

Possible that many domains could be blocked waiting for promises, but I think they'd eventually resolve or max out the queue limit (if we have one).

Updating the documentation is a good idea, will do that!

raphael-proust: Ok. Essentially I think this problem is one of documentation. `run_in_main` should have a biggish warning about freeze and about the possibility of dead-locks.

When finalising this and getting it merged and all, we can then add to the documentation with an example of how to avoid running `run_in_main` and to prefer a stepping mechanism where the parallel computation returns a partial result (similar to what parsers often do when they need to wait for more input). This part really can wait a bit: it's not worth putting some effort in writing this doc until the interface is finalised.

src/unix/lwt_domain.ml
raphael-proust: Another possibility :
- We use `Option` to indicate that the job limit was reached in this low-level interface. And we put together an example (or even a whole separate module) with a higher-level interface that implements the double-promise on top of the option. Basically, we focus on the low-level interface for this module and we build on top of it.

Anyone else has ideas and opinions?

Sudha247: I think we don't have a strict requirement for introducing a job limit currently. To provide some context, there have been some improvements in domainslib (cf. https://github.com/ocaml-multicore/domainslib/pull/29). This has resulted in eliminating some stack overflows that we'd encountered earlier (see https://github.com/ocaml-bench/sandmark/pull/173). This applies to the Lwt_domain interface too, since it uses domainslib internally.

In future, the plan is to introduce fibers in domainslib, which will play well with nested parallelism. Even without fibers, the current implementation is stable enough for our benchmarks. If this makes sense, I propose we don't introduce a job limit for `Lwt_domain`.

src/unix/lwt_domain.ml
```diff
@@ -0,0 +1,82 @@
...
+
+let initialized = ref false
+
+(* Destroys old pool and creates a new pool with `num` domains *)
+let set_num_domains num =
+  initialized := true;
+  max_domains := num;
+  T.teardown_pool !pool;
+  pool := T.setup_pool ~num_additional_domains:(!max_domains - 1);
+  domains_count := !max_domains
```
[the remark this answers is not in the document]
raphael-proust: You set the current threashold at `<= 0`. What happens if I call this function with `1`?

Does it create zero other domains? In this case, what does `detach` do? Can it actually schedule the computation or does it create a promise that's never resolved? Does it schedule it on the same domain as the one running main.run?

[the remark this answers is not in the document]
raphael-proust: from a short experiment, setting max-domains to 1 seems to be blocking any call to `detach`

There are several options here:
- We reject `1` as well, to avoid blockage
- We special case `1` so that `detach` actually runs the computation on the same core
- other?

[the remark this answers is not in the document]
Sudha247: You're right, setting `max_domains` to 1 blocks indefinitely since the bind waits for the detach computation to resolve and for the computation to run it needs the bind to succeed.

Rejecting `1` and making the minimum number of domains to be `2` seems the simpler way. We can still run this on one core, but the program would take longer to execute on one core. This means the program cannot be executed with one domain. If that's not an issue, we could go with the first option.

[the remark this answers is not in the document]
Sudha247: On further consideration, I think we should give provision for running this with one domain. It can be useful for:
* Debugging - we've found debugging easier with one domain. It can also help classify concurrency bugs from other bugs.
* Benchmarking - running on one domain can act as the sequential baseline.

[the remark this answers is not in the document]
avsm: I think `max_domains=1` should definitely be made to work in the Lwt_domains interface. A single core still counts as a multicore setup :) More generally, the domain running the Lwt engine should be able to run detached computations too, since that works already via Lwt_preemptive and systhreads.

avsm:
We need to define whether the [Lwt.async exception hook](https://ocsigen.org/lwt/3.1.0/api/Lwt#VALasync_exception_hook) is supported or not when on a different domain.

src/unix/lwt_domain.mli
```diff
@@ -0,0 +1,70 @@
...
+        inside the promise that called [detach] that called [run_in_main].
+
+        Note that the calling domain will be idle until [f ()] completes
+        execution and returns the result. Thus, heavy use of [run_in_main] may
+        lead to most or all domains being frozen. It's also possible to create a
+        dead-lock when [run_in_main] is called (thus freezing a domain) with a
+        function that calls [detach] (thus needing a domain). Consequently, it
+        is recommended to use this function sparingly. *)
+
+  val set_num_domains : int -> unit
```
kayceesrk, line 49: Instead of `set_num_domains`, I would prefer to have `val setup_pool: int -> unit` and an explicit `val teardown_pool: unit -> unit`. 

I would also like `setup_pool` to raise an exception if a pool is currently active. I am not fond of `set_num_domains` implicitly shutting down the previous pool and creating a new pool. It may be the case that there are active tasks in the pool and implicitly tearing down pool and setting up a new one will silently kill those tasks. At least with an explicit `teardown_pool`, you get to choose to do the expensive operation explicitly. `setup_pool` and `teardown_pool` are what is used by `domainslib`. 

Also, an explict `teardown_pool` gives the chance for the program to cull the domains and not pay the associated cost of participating in the GC. This is useful if the program were to have a specific portion where there is parallelism opportunity and long portions where there isn't parallelism available.

raphael-proust:
I toyed around with replacing the Lwt_unix.condition signaling mechanism with a lower-level, no-Unix version, but I wasn't satisfied with the result. I think that we can move the Lwt_domain module outside of unix/ and into core/ but we might need

- to add some new global data-structure in lwt.ml (similar to `paused` but for detached computation instead),
- add some interaction in the loop of Lwt_main.run to check on those (again, similar to pause),
- add some other mechanism in the js-of-ocaml lwt runtime.

In my opinion, this can wait: we should focus on getting something that works and we can later get it to work better.

src/unix/lwt_domain.ml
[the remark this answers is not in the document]
raphael-proust: I don't think that we need to preemptive or systhread the detach computation if we have a single domain. I think that in this case we can just run the computation directly. If it is well documented I think the interface should be ok:

> If there is a single domain, then `detach f x` is just `return (f x)`. In other words, if there isn't a separate domain to execute the computation on, then the computation is run on the domain that also executes the main Lwt event loop.

I think this is better (than using systhreads) for the following reasons:
- When we eventually move away from Lwt_unix and into Lwt(core) it'll be easier
- The cost model is easier to understand
- The concurrency/parallelism model is easier to understand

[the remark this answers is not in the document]
Sudha247: When we have `max_domains=1`, the computation runs directly. (see 951df84dd181185f3b24f0f4a099df3a441facf1)

src/unix/lwt_domain.mli
Sudha247: Done in c7e92da6ad4d9ab849f7f5b2dc7635d861efb6a5

raphael-proust: I'm wondering if, with this exception discipline, for consistency, we should not auto-initialise on `detach` and instead raise an exception there too. Would that make sense?

raphael-proust: Alternatively, when the pool isn't initialised, we could consider `max_domains = 1` and thus `detach` just runs within the main domain?

kayceesrk: > Alternatively, when the pool isn't initialised, we could consider max_domains = 1 and thus detach just runs within the main domain?

Here is how I thought about the problem. If the user does `Lwt_domain.detach` without setting up the pool, how likely is that they have forgotten to set up the pool? I would think very likely since they want to make their code go faster, but haven't informed the system about how many domains should be used. Hence, I would still prefer `detach` to raise an exception if a pool weren't set up. 

That said, I don't have a strong opinion about this, and it is ok to go for `detach` that auto initialises the pool.

avsm: Alternatively, why not make `Lwt_detach` take an _explicit_ `pool` parameter to every computation, and have the init/teardown functions return a `pool` value?

This makes it possible for the application to assign computations to specific domain pools, which is coming up as a common requirement.

kayceesrk: I like that idea. The issue with that is that you'll have to thread through the `pool` value in your code. Porting your sequential Lwt code to use `Lwt_domain` will require modifying more code than the current API. Alternatively, if `domainslib` were to provide named pools, then the change can be local to where you're adding parallelism.

avsm: Adding named pools into Domainslib does seem better, as it would also allow non-Lwt code to share the underlying domains (e.g. direct-style code)

Sudha247: Passing an explicit pool parameter sounds like a good idea. This will also make managing pools easier when using multiple libraries like say Lwt_domain, domainslib and more that'd be developed in future with multicore features. In that case, can we use setup/teardown functions directly from domainslib? If offering the same API in `Lwt_domain` would make dependencies easier, we can have `Lwt_domain.setup_pool` return a domain pool and pass it around. It will need more code than the API in its current form, though this could be similar to domainslib APIs.

avsm: I think the answer lies in which library 'owns' the scheduling and management of all the domains in an application. Typically it is a bad idea to multiplex schedulers since they will end up competing, so having a single place (like Domainslib) spawn and partition and schedule jobs to domains makes sense.  Once that core interface exists in Domainslib (e.g. via named pools as @kayceesrk suggests), then downstream libraries like Lwt and Async can pass the appropriate invocations to it.  

So we could have something like:

- `Domainslib`:
  * add support for naming pools with a number of sub-domains (the equivalent of `set_num_domains` and `setup_pool` and `teardown_pool` in this PR).
  * ensure there is a default pool with a sensible number of domains so if nothing is setup, there is still a pool to go to globally.
- `Lwt`:
  * add an interface `Lwt.detach ?pool fn` that can dispatch a computation to a pool optionally, but will go to the global Domainslib pool if none is specified.
  * this will be a simple interface by default (no pool required) but specialised applications that need per-pool dispatch (like the Tezos signature checkers, or MirageOS crypto libs) could setup a Domainslib named pool and pass in a `Lwt.detach ~pool` argument through their applications.

Later, other libraries like Async and Eio could also use the same Domainslib interface to create further named pools, but only the Domainslib scheduler would manage those pools, avoiding scheduler multiplexing.

raphael-proust: Note that for Lwt, it would make sense that the pool is created just before the call to `Lwt_main.run` (this is currently recommended in the documentation). This means that the threading of the `pool` value needs to start at the very top of the call graph of the whole program. Moreover, because `detach` is used for non-Lwt functions (its type is doesn't allow Lwt computations) the `pool` value need to be threaded all the way down to the deepest branches of the call graph of the whole program. Basically, it'd be very invasive.

So a default pool and/or some named pools would be a welcome addition.

-------------------------------------------------------

About calling `detach` without a prior `setup_pool`, I'm ok with the exception approach.

There is actually another reason why it might be good to go for the exception approach and it has to do with an under-documented feature(bug? who is to say?) of Lwt. Specifically, before the event-loop is started (when the call to `Lwt_main.run` is made) and after the event-loop has stopped (when the call to `Lwt_main.run` has returned), I/O jobs can be started and can even progress, but their promises can never be resolved.

With `Lwt_domain.detach` you can detach computations even if the event-loop is not running. I am not sure that the current mechanism allows for resolution of these promises when the scheduler is not running. (Currently it uses a feature of Lwt_unix that I am not very familiar with.)

In the future we might want to provide a solution not dependent on the Unix module, to provide solutions for other schedulers, etc. To this end, it makes more sense to have a simple interface with a simple semantic which we can have across all backends. And raising an exception if the pool is not set up is a simple interface.

raphael-proust:
> We need to define whether the [Lwt.async exception hook](https://ocsigen.org/lwt/3.1.0/api/Lwt#VALasync_exception_hook) is supported or not when on a different domain.

The documentation about the exception is explicit. But to make sure the situation re: the async exception hook is clear, we could add a paragraph that explains what happens if we do `Lwt.dont_wait (fun () -> Lwt_domain.detach …) (fun exc -> …)` and if we do `Lwt.async (Lwt_domain.detach …)`.

We can add this later because it can be inferred from the current documentation.

src/unix/lwt_domain.mli
Sudha247: https://github.com/ocaml-multicore/domainslib/pull/45 adds named pools to domainslib. This enables obtaining pools at a later point after their creation with a "name" attached to it during its creation. Essentially this eliminates the need to thread through pool. 

With this, we can eliminate the global pool present in `Lwt_domain` and pass pool as an argument to `detach`. We can also provide wrappers for initializing/tearing down pools in the interface itself, so that it's self contained. I plan to switch to this soon, if there are concerns about this please do let us know.

Sudha247: the PR is rebased to [domainslib.0.3.1](https://github.com/ocaml-multicore/domainslib/releases/tag/0.3.1) (cf. 02ab32f933cf798bc5ffe56b1d7587d90ec1c10c). Major updates are:
* A `pool` is passed explicitly to `detach` computations. This eliminates the need for a global pool we had earlier. If the pool is not active, it raises an `Invalid_argument` exception.

* domainslib `0.3.1` has support for named pools which enables obtaining pools with a name at a later point, and hence we don't need to thread through the pool argument.

* Includes wrappers to some domainslib functions like `setup_pool`, `teardown_pool` etc. to make this module self-contained.

src/unix/lwt_domain.ml
```diff
@@ -0,0 +1,67 @@
+open Lwt.Infix
+
+module C = Domainslib.Chan
+module T = Domainslib.Task
+
+type pool = Domainslib.Task.pool
+
+let setup_pool ?name num_additional_domains =
+    T.setup_pool ?name ~num_additional_domains:num_additional_domains ()
```
yawaramin, line 9:
```suggestion
    T.setup_pool ?name ~num_additional_domains ()
```

src/unix/lwt_domain.ml
```diff
@@ -0,0 +1,67 @@
...
+
+
+let teardown_pool = T.teardown_pool
+
+let lookup_pool = T.lookup_pool
+
+let get_num_domains = T.get_num_domains
+
+let init_result = Result.Error (Failure "Lwt_domain.detach")
```
yawaramin, line 18:
```suggestion
let init_result = Error (Failure "Lwt_domain.detach")
```

`Ok` / `Error` are in `Stdlib` so they are in scope.

src/unix/lwt_domain.ml
```diff
@@ -0,0 +1,67 @@
...
+let run_in_main f =
+  let res = ref init_result in
+  let job () =
+    Lwt.try_bind f
+      (fun ret -> Lwt.return (Result.Ok ret))
+      (fun exn -> Lwt.return (Result.Error exn)) >>= fun result ->
+    res := result;
+    C.send job_done 1;
+    Lwt.return_unit
+  in
```
yawaramin, line 60:
```suggestion
  let job () =
    let open Lwt.Syntax in
    let+ result = Lwt.try_bind f Lwt_result.return Lwt_result.fail in
    res := result;
    C.send job_done ()
  in
```

Taking advantage of some slightly newer features in Lwt, like the [`Lwt_result`](http://ocsigen.org/lwt/latest/api/Lwt_result) module and letops. Also I think this is the only use of `Lwt.Infix` in this file so that open is not needed any more.

Also I am suggesting to change the `job_done` channel to type `unit` as its payload carries no meaning.

yawaramin (commented):
A few suggestions.

src/unix/lwt_domain.ml
```diff
@@ -0,0 +1,67 @@
...
+  C.send jobs job;
+  Lwt_unix.send_notification job_notification;
+  (* blocks calling domain until the job is executed *)
+  ignore @@ C.recv job_done;
```
yawaramin, line 64: If the channel type is `unit` then this becomes `C.recv job_done;`

src/unix/lwt_domain.ml
raphael-proust: It's somewhat annoying that `Lwt_result.catch` lets some exceptions escape by taking a promise rather than a thunk as argument, otherwise it'd have been perfect.

We could use `let+ result = Lwt_result.catch (Lwt.apply f ()) in ..` but that's not more readable.

Sudha247: >It's somewhat annoying that Lwt_result.catch lets some exceptions escape by taking a promise rather than a thunk as argument

Are there any examples of some exceptions escaping `Lwt_result.catch`? The rather simple ones I tried out seemed to work as expected. Also, is this something an user needs to be aware of?

yawaramin:
Thinking about this design a bit more; I think we also need a convenience function:

```ocaml
val Lwt.domain : (* or, detach/blocking/… *)
  ?pool:Lwt_domain.pool ->
  ('a -> 'b) ->
  'a ->
  'b Lwt.t
```

In a typical application, we might want to run a blocking computation at various points in the code. It would become a hassle to set up a domain pool at the application’s top layer and thread that through to all lower layers which needed it. Lwt should set up a domain pool by default, and allow users to override it if necessary.

We can add some docs to the effect that ‘promises are for non-blocking I/O but we have a convenient way to represent blocking computations as promises’.

Sudha247:
>It would become a hassle to set up a domain pool at the application’s top layer and thread that through to all lower layers which needed it.

It sure is a hassle, especially while modifying existing modules. This is why named pools were introduced in [domainslib.0.3.1](https://github.com/ocaml-multicore/domainslib/releases/tag/0.3.1). Pools can be obtained at a later point with the help of an identifier string it is associated to, thereby eliminating the need to thread through the pool across layers.

>Lwt should set up a domain pool by default, and allow users to override it if necessary.

We were doing this earlier. But it could turn out problematic inter-operating with multiple Multicore enabled libraries, in the sense that it gets harder tracking down all the live domain pools. Ideally in Multicore OCaml (and OCaml 5.0) programs; number of domains = number of cores, exceeding that will introduce overheads and could slow down the program. If at all we do end up with a default domain pool, I think the right place for that to reside would be at domainslib.

>We can add some docs to the effect that ‘promises are for non-blocking I/O but we have a convenient way to represent blocking computations as promises’.

This sounds like a good idea. I think I'm not clear about how this convenience function would be useful, it looks like it's exposing `Lwt_domain.detach` to Lwt but I may be misreading here.

Sudha247:
Thanks to @raphael-proust `lwt-domain` is moved to its own package now.

yawaramin:
Thanks for the explanation. I am now understanding a couple of new things. Btw let me preface this comment by saying that, if this `lwt_domain` package were to be merged right now, in my humble opinion it is fine as it is. My suggestion is just to improve developer experience.

> Pools can be obtained at a later point with the help of an identifier string it is associated to,

OK, that is convenient, and then my suggestion becomes:

```ocaml
val Lwt.domain : (* or, detach/blocking/… *)
  string ->
  ('a -> 'b) ->
  'a ->
  'b Lwt.t
```

> But it could turn out problematic inter-operating with multiple Multicore enabled libraries

That's a bummer.

> If at all we do end up with a default domain pool, I think the right place for that to reside would be at domainslib.

Fair enough.

> I think I'm not clear about how this convenience function would be useful,

I am suggesting that in most cases users should not have to think about `Lwt_domain`, and just call `Lwt.domain` to 'magically' parallelize their code. E.g.:

```ocaml
(* main.ml *)

let main () =
  let _ = Lwt_domain.setup_pool ~name:"main" 3 in
  ...

let () = main ()

(* person.ml *)

type t = {
  id : int;
  name : string;
} [@@deriving of_yojson]

let of_yojson = Lwt.domain "main" of_yojson
(* val of_yojson : Yojson.Safe.t -> (t, string) Lwt_result.t *)
```

Now just by wrapping an existing CPU-bound function with `Lwt.domain "main"`, we turn it into a parallelized function in the Lwt context. This convenience will be very beneficial.

src/unix/lwt_domain.mli
raphael-proust: All the points discussed in this conversation have been addressed. I'm resolving the convesation.

src/unix/lwt_domain.ml
raphael-proust: The issue only appears for somewhat corner cases:

```
# let f x = if x < 0 then raise (Invalid_argument "") else Lwt.pause () >>= fun () -> Lwt.return x ;;
val f : int -> int Lwt.t = <fun>
# Lwt_result.catch (f (-1)) ;;
Exception: Invalid_argument "".
```

The issue is that expressions that raise an exception can be given the type of promises. Compare to `catch` which takes a `unit -> 'a t` instead of an `'a t` which allows it to install a `try`-`catch` around it.

But fixing that is out of the scope of this MR.

-----------------

I find the current form of the code satisfying. I'm resolving because this is a somewhat cosmetic suggestion. @yawaramin feel free to open an MR against this branch (or master if this branch is merged) to get this and other improvements merged.

test/unix/main.ml
```diff
@@ -4,8 +4,7 @@
 open Tester
 
 let () =
-  Test.concurrent "unix" [
-    Test_lwt_unix.suite;
+  Test.run "unix" [
```
raphael-proust, line 7: We need to revert this. The tests can now work concurrently.

raphael-proust:
rebased on master to resolve merge conflict

PR #861, opened 2021-05-27
Catch all the possible exceptions during unlinking on win32

raphael-proust:
This intends to catch more of the exceptions (all of them in fact) that may be raised during the work-around needed to remove temporary directories on Win.


avsm:
@MisterDA would you have time to review this before release? xref to your original PR in https://github.com/ocsigen/lwt/pull/849

MisterDA:
LGTM. Very defensive programming, right?

PR #865, opened 2021-05-31
Check that Lwt.bind preserves tail-recursion

yurug:
Lwt should preserve tail-recursion. When a recursive function [f] lives in the `Lwt` monad, a programmer must assume that [f] is only call in tail position in `Lwt.bind`.

Given that robustness to stack overflows cannot be ignored by OCaml programmers, this property should be documented, be part of `Lwt` specification, and should probably never broken in the future.

This test tries to ensure that `Lwt.bind x f` indeed calls `f` only in tail position. For the moment, this MR only includes a single test corresponding to the immediate executions of `f`. We may need a test for each call to `f`, but it is unclear to me because `f` is then probably called under a fresh stack when a callback is deferred.

We checked that the test fails if the following patch is applied:

```ocaml
diff --git a/src/core/lwt.ml b/src/core/lwt.ml
index 877273c81..97ee3ddbe 100644
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -1315,13 +1315,20 @@ struct

     p

+  let something =
+    let r = ref 0 in
+    fun () -> incr r
+
   let run_callback_or_defer_it
       ?(run_immediately_and_ensure_tail_call = false)
       ~callback:f
       ~if_deferred =

     if run_immediately_and_ensure_tail_call then
-      f ()
+      let y = f () in
+      something ();
+      y
+

     else
       let should_defer =
```

### DIFF
--- a/test/core/test_lwt.ml
+++ b/test/core/test_lwt.ml
@@ -4442,3 +4442,35 @@ let callback_list_tests = suite "callback cleanup" [
   end;
 ]
 let suites = suites @ [callback_list_tests]
+
+(*
+
+   Lwt should preserve tail-recursion. When a recursive function [f]
+   lives in the [Lwt] monad, a programmer must assume that [f] is only
+   call in tail position in [Lwt.bind].
+
+   Given that robustness to stack overflows cannot be ignored by OCaml
+   programmers, this property should be documented, be part of
+   [Lwt] specification, and probably never be broken in the future.
+
+   This test tries to ensure that [Lwt.bind x f] indeed calls [f] only
+   in tail position.
+
+*)
+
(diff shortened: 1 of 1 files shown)


yurug:
I guess the failing test is not related to this PR, right?

yurug:
https://github.com/marketplace/actions/shellcheck

raphael-proust:
> I guess the failing test is not related to this PR, right?

That's correct. Your tests are executed in the "Main workflow" tests (because they call `opam exec -- make test`). This means that your test passes on all the supported version of OCaml and on multiple OS and also flamba-musl-static.

PR #869, opened 2021-06-18
Use `is_blocking` in `dup` and `dup2` to fix `ENOTSOCK` on Windows 

MisterDA:
On Windows, calling `Unix.clear_non_block` and `Unix.set_non_block` is
only supported on socket file descriptors. The function `is_blocking`
already deals with the corner cases, but the current implementations
of `dup` and `dup2` don't make use of it. Thus, when the `dup` or `dup2`
wrappers are called with file descriptors, `ENOTSOCK` errors are
raised. The fix is to reuse `is_blocking` in the wrappers.

### DIFF
--- a/src/unix/lwt_unix.cppo.ml
+++ b/src/unix/lwt_unix.cppo.ml
@@ -1242,13 +1242,8 @@ let dup ch =
     set_flags = ch.set_flags;
     blocking =
       if ch.set_flags then
-        lazy(Lazy.force ch.blocking >>= function
-        | true ->
-          Unix.clear_nonblock fd;
-          Lwt.return_true
-        | false ->
-          Unix.set_nonblock fd;
-          Lwt.return_false)
+        lazy(Lazy.force ch.blocking >>= function blocking ->
+               Lazy.force (is_blocking ~blocking fd))
       else
         ch.blocking;
     event_readable = None;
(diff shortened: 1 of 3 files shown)


src/unix/lwt_unix.cppo.ml
```diff
@@ -1224,46 +1224,28 @@ let access name mode =
...
     hooks_writable = Lwt_sequence.create ();
-  }
+  } in
+  set_blocking ~set_flags:ch1.set_flags ch2 ch1.set_flags;
+  ch2
 
 let dup2 ch1 ch2 =
   check_descriptor ch1;
   Unix.dup2 ch1.fd ch2.fd;
-  ch2.set_flags <- ch1.set_flags;
```
raphael-proust, line 1254: This line is not necessary for the establishment of the `blocking` field anymore. However, removing this line also means that `ch2` is in a different state when the function returns. (In the previous behaviour, it was set explicitly now it isn't.)

I'm not sure it can cause issues but maybe? Consider the following case:

```
let cha = … in (* some ch, whatever *)
let chb = … in (* some other ch *)
let dup2ed = dup2 cha chb in (* chb's `set_flag` is unaffected by this *)
… (* some stuff here *)
let duped = dup dup2ed in (* possible issue here *)
…
```

The possible issue comes from the fact that `dup` reads `dup2ed.set_flags` which is set lazily when `dup2ed`'s `blocking field is forced. And so the value of the `set_flags` and `blocking` fields for `duped`, depends on whether `some stuff here` forces this field or not.


----------------------

Am I missing something? Is there a guarantee that the field is forced? Is there another piece of the control-flow I haven't seen?

smorimoto:
Looks good now!

PR #870, opened 2021-06-24
Support `Unix.socketpair` using `PF_UNIX` on Windows and OCaml 4.14 

MisterDA:
A few improvements (imho) to the support of `socketpair` under Windows:
- fix the `socket_domain_table` length;
- add support for IPv6 `socketpair` emulation (I'm rooting for the Universal Deployment of IPv6);
- support Unix domain sockets and `socketpair` emulation using Unix domain sockets to be introduced in OCaml 4.14 (PR 10192).

This allows us to use socketpair and Unix domain sockets to perform IO redirection with processes, which isn't possible with current sockets in OCaml under Windows. Sockets have to be created in non-overlapped mode, with `WSASocket(domain, type, protocol, NULL, 0, 0)` rather than `socket(domain, type, protocol)` (overlapped). Unix domain sockets under Windows don't have this limitation.

### DIFF
--- a/src/unix/lwt_unix.cppo.ml
+++ b/src/unix/lwt_unix.cppo.ml
@@ -1689,7 +1689,11 @@ let shutdown ch shutdown_command =
 external stub_socketpair : socket_domain -> socket_type -> int -> Unix.file_descr * Unix.file_descr = "lwt_unix_socketpair_stub"
 
 let socketpair dom typ proto =
-#if OCAML_VERSION >= (4, 05, 0)
+#if OCAML_VERSION >= (4, 14, 0)
+  let do_socketpair =
+    if Sys.win32 && (dom <> Unix.PF_UNIX) then stub_socketpair
+    else Unix.socketpair ?cloexec:None in
+#elif OCAML_VERSION >= (4, 05, 0)
   let do_socketpair =
     if Sys.win32 then stub_socketpair
     else Unix.socketpair ?cloexec:None in


raphael-proust:
Hi,

Thanks for the contribution. IIUC, this won't be effective until OCaml4.14 is released (or until someone installs it within a custom opam switch or something along those lines). I'm tempted to withhold merging until 4.14 is released or is about to be released. It might be a bit long though.

In the meantime, I'll try to get someone with more C and system-programming chops to do the review.

MisterDA:
>
> I'm tempted to withhold merging until 4.14 is released or is about to be
> released. It might be a bit long though.
>
Completely understandable, but please consider cherry-picking the first two
commits. The first one is a bug-fix, and the second enables socketpair over
IPv6 (because why not).

Let's say I'll bump this PR when OCaml 4.14 alpha is released.

> In the meantime, I'll try to get someone with more C and
> system-programming chops to do the review.
>
Thank you!

>

raphael-proust:
Ok, I'll cherry-pick those after the review. Thanks!

avsm:
I've read through this and it looks fine to me, and matches the upstream PR to OCaml 4.14 that @MisterDA submitted. I'd be ok with the cherrypick. Perhaps @dra27 can have a look if he has time.

dra27:
Looks good to me, too.

raphael-proust:
Ok, I'll open a branch for cherry-picking the prefix of the PR. Thanks for the reviews!

raphael-proust:
Done in https://github.com/ocsigen/lwt/pull/876

I simply cherry-picked the commits and added a changelog entry. I'll merge once the CI passes.

PR #873, opened 2021-07-24
Lwt_stream.of_lwt_seq: 'a Lwt_seq.t -> 'a Lwt_stream.t

raphael-proust:
Goes along with `Lwt_stream.of_seq` but for `Lwt_seq`. Has the same lazy semantic and the same tests.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -25,7 +25,9 @@
 
   * Lwt_stream.return, Lwt_stream.return_lwt: singleton stream constructors (#864, Boning Dong).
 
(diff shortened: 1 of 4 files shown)


raphael-proust:
Inspired by https://github.com/ocsigen/lwt/pull/864

smorimoto:
Just needs a CHANGES entry!

raphael-proust:
> Just needs a CHANGES entry!

Good catch! Added.

smorimoto (approved):

PR #880, opened 2021-09-01
Use CAMLnoreturn_{start,end} instead of Noreturn

nojb:
This fixes compilation under MSVC, see

https://github.com/ocaml/ocaml/blob/4e9eb755e60d0f0f4e458eb3f7af26bce5d8e2ba/runtime/caml/misc.h#L52-L61

### DIFF
--- a/src/unix/lwt_unix.h
+++ b/src/unix/lwt_unix.h
@@ -44,7 +44,9 @@ char *lwt_unix_strdup(char *string);
     (type *)lwt_unix_malloc(sizeof(type) + size)
 
 /* Raise [Lwt_unix.Not_available]. */
-void lwt_unix_not_available(char const *feature) Noreturn;
+CAMLnoreturn_start
+void lwt_unix_not_available(char const *feature)
+CAMLnoreturn_end;
 
 #define LWT_NOT_AVAILABLE_BYTE(prim)             \
     CAMLprim value lwt_##prim(value *a1, int a2) \


nojb:
Friendly ping

smorimoto (approved):

nojb:
Thanks! Sorry for the lag, but this macro does not seem to be available on `4.02`... is that a problem?

smorimoto:
I personally believe that supporting MSVC would be more valuable than keeping `4.02` support, but I don't know why this fails in `4.02`. If I remember correctly, it would have been available even with `4.02`.

nojb:
> I personally believe that supporting MSVC would be more valuable than keeping `4.02` support, but I don't know why this fails in `4.02`. If I remember correctly, it would have been available even with `4.02`.

We could continue to use `Noreturn` for 4.02 (if there is an easy way to test for the OCaml version using the C preprocessor).

smorimoto:
Yes, that sounds good, but is there a way to make a special version constraint only on Windows? Otherwise, I think it would be clean to drop 4.02.

MisterDA:
There is, see https://ocaml.org/manual/intfc.html#ss:c-internal-macros. Just opened a PR to restore `Noreturn` on 4.02.

Some parts of Lwt already don't support 4.02.

```
- lwt_ppx → ppxlib >= 0.16.0 → ocaml >= 4.04.1
    base of this switch (use `--unlock-base' to force)
 ```

In https://github.com/ocsigen/lwt/pull/843 I suggested to raise the minimum requirement on 4.06.

smorimoto:
Oh, my "IIRC" proved wrong again here!
If we raise the downward version constraint on Windows, the #887 seems not necessary. And I would like to +1 to making it happen.

PR #882, opened 2021-09-13
Add Lwt_bytes.{unsafe_,}blit_from_string

hhugo:
Lwt_bytes was missing bliting from string.
- add blit_from_string and unsafe_blit_from_string
- update doc to make distinction between `string` and `bytes`
- add tests for the new functions

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,8 @@
 
 ====== Additions ======
 
+  * Lwt_bytes.blit_from_string: string complement of Lwt_bytes.blit (#882, Hugo Heuzard).
+
   * Lwt_seq: a Seq-like data-structure with Lwt delayed nodes (#836, #842, Zach Shipko).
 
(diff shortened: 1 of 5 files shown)


raphael-proust (changes requested):

raphael-proust:
Rebased on master where the CI is fixed

raphael-proust (approved):

raphael-proust:
I pushed an additional commit which is only style: it replaces the following pattern

```
try
  let () = Lwt_bytes.… in
  Lwt.return_false
with
| Invalid_argument _ -> Lwt.return_true
| _ -> Lwt.return_true
```

with the following pattern

```
match Lwt_bytes.… with
| exception Invalid_argument _ -> Lwt.return_true
| () -> Lwt.return_false
```

(Also, it adds one test.)

We can roll it back if there are any issues with the test. @hhugo can you have a look at the commit?

raphael-proust:
Probably the new pattern for tests is not compatible with old versions of OCaml so I'll revert it if the CI indicates so.

PR #883, opened 2021-09-16
Fix Lwt_unix.lstat calling Unix.stat on Win32

MisterDA:
A nice little bug here ;)

It can become rather nasty with this code (how I found it):

```ocaml
  Lwt_main.run begin
      with_temp_dir @@ fun dir ->
      let link = Filename.(concat dir "link") in
      Unix.symlink ~to_dir:true "C:/Windows" link;
      do_stuff_here ()
    end
  (* If [link] is a directory, Lwt will enter it and try to
     delete its contents *)
```

A little grep in Lwt tells that only `Lwt_io.with_temp_dir` uses `Lwt_unix.lstat`.

PR #886, opened 2021-09-16
Support deleting symlinks in Lwt_io.with_temp_dir on Windows

MisterDA:
If a Windows symlink points to a directory, then rmdir is the correct
function to call. If it points to a file, then unlink is the correct
function. When pointing to a file, the symlink can also have the
read-only attribute, which makes it subject of the little chmod dance.

This depends on the fix in https://github.com/ocsigen/lwt/pull/883.


smorimoto:
@MisterDA Could you rebase this?

MisterDA:
Rebased.

PR #890, opened 2021-10-05
Add OCaml 4.13.x to GitHub Actions workflow

smorimoto:


smorimoto:
Some useful error messages have come up since 4.13: https://github.com/ocsigen/lwt/runs/3812113197?check_suite_focus=true#step:7:1

FYI: @raphael-proust

raphael-proust:
@smorimoto I made the PR https://github.com/ocsigen/lwt/pull/899 to fix those new warnings.

Issue #677, opened 2019-03-21
Add ppx_let integration test to run in Travis


aantron:
Testing Lwt with ppx_let is probably not suitable for ordinary unit tests, because it will greatly bloat the dependencies for ordinary development. We should, however, test that the new `Lwt.Infix.Let_syntax` module works correctly with ppx_let in an optional additional test suite, which can run in Travis.

EDIT: follow-on to #668.

PR #891, opened 2021-10-05
Restore ocaml-ci

dra27:
ocaml-ci began to fail when https://github.com/ocurrent/ocaml-ci/pull/338 was merged, since `dune build @check` requires `ppx_let`.

For testing purposes, `ppx_let` only adds a test dependency on base, which (to me at least) doesn't seem to impose as heavy a dependency burden as suggested in https://github.com/ocsigen/lwt/issues/677.

This approach is far from the only possibility, but I add a dummy opam file whose purpose is to create a package context to house the dependency. This gets picked up by ocaml-ci and also by `opam install . --deps-only`

### DIFF
--- a/lwt_ppx_let.opam
+++ b/lwt_ppx_let.opam
@@ -1,0 +1,27 @@
+opam-version: "2.0"
+
+synopsis: "Dummy package context for ppx_let tests"
+
+version: "5.4.2"
+license: "MIT"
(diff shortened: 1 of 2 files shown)


smorimoto (approved):

smorimoto:
Thank you so much @dra27!

PR #895, opened 2021-10-14
Use exception patterns in pattern matching instead of match try

MisterDA:
That's nicer and also tail-recursive.
https://ocaml.org/manual/coreexamples.html#s%3Aexceptions
https://ocaml.org/manual/patterns.html#sss:exception-match


smorimoto:
Improvable points that have been overlooked for more than 10 years!

PR #903, opened 2021-11-19
 `CreateProcess` needs `CREATE_UNICODE_ENVIRONMENT`

MisterDA:
> If the environment block pointed to by `lpEnvironment` contains
> Unicode characters, be sure that `dwCreationFlags` includes
> `CREATE_UNICODE_ENVIRONMENT`. If this parameter is `NULL` and the
> environment block of the parent process contains Unicode characters,
> you must also ensure that `dwCreationFlags` includes
> `CREATE_UNICODE_ENVIRONMENT`.

From [doc1][].

> If this flag is set, the environment block pointed to by
> `lpEnvironment` uses Unicode characters. Otherwise, the environment
> block uses ANSI characters.

From [doc2][].

This was forgotten from PR #843 that switched to Unicode strings on
Windows (including the environment). I noticed it when reading code
from win32unix.

[doc1]: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw
[doc2]: https://docs.microsoft.com/en-us/windows/win32/procthread/process-creation-flags#CREATE_UNICODE_ENVIRONMENT

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -41,7 +41,7 @@
 
   * Code quality improvement: remove an uneeded Obj.magic (#844, Benoit Montagu).
 
-  * On Windows, use the Unicode API in C stubs and functions introduced in OCaml 4.06 to handle Unicode strings. Raise the minimum requirement to OCaml 4.06 (on Windows only). (#843, Antonin Décimo)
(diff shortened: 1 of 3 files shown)


smorimoto (approved):

smorimoto:
Thanks @MisterDA!

PR #905, opened 2021-11-29
small documentation additions

raphael-proust:


docs/Makefile
```diff
@@ -5,18 +5,20 @@ PKGS=\
...
 	-I ${BLD}/ppx/.ppx_lwt.objs/byte \
 	-I ${BLD}/react/.lwt_react.objs/byte \
+	-I ${BLD}/domain/.lwt_domain.objs/byte \
 	-I ${BLD}/unix/.lwt_unix.objs/byte
 
 MLIS=\
 	$(wildcard ${SRC}/core/*.mli) \
 	$(wildcard ${SRC}/ppx/*.mli) \
-	$(wildcard ${SRC}/react/*.mli) \
+	$(wildcard ${SRC}/domain/*.mli) \
```
smorimoto, line 21: Is it intentional that react is removed here?

smorimoto (approved):

smorimoto:
I'd like to go with this, apart from the one thing I commented on!

docs/Makefile
raphael-proust: That's an error. I wanted to add an entry rather than to replace one. It's fixed now.

smorimoto: Looks good now!

PR #906, opened 2021-11-30
Fix win32_spawn leaking dev_null fd in the parent process

MisterDA:


MisterDA:
Added a change entry.

smorimoto (approved):

smorimoto:
Could you check this @raphael-proust?

smorimoto:
I took a close look, but it looks good, so I will merge this.

PR #909, opened 2021-12-06
Windows: make `create_process` duplicate non-inheritable std handles

MisterDA:
`Unix.create_process` currently always duplicates the handles given to
the child process to make sure they're inheritable (keep-on-exec).  I
have opened the PR 10807 on OCaml to duplicate the handles only if
they're not inheritable (close-on-exec).

This PR mimic this behaviour in Lwt for consistency (assuming the
OCaml PR will be accepted).

See the following PRs:
- https://github.com/ocaml/ocaml/pull/650;
- https://github.com/ocaml/ocaml/pull/10807.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,10 @@
   * In the Lwt_io module, add `?cloexec:bool` optional arguments to functions that create file descriptors (`pipe`). The `?cloexec` argument is simply forwarded to the wrapped Lwt_unix function. (#872, #911, Antonin Décimo)
   * Add Lwt_result.error, Lwt_result.iter, and Lwt_result.iter_error for consistency with Stdlib. (#927, Antonin Décimo)
 
+====== Misc ======
+
+  * On Windows, make Lwt_process.create_process duplicate standard handles given to the child process if they're not inheritable, to mimic the behaviour of Unix.create_process. (#909, Antonin Décimo)
+
 ====== Fixes ======
 
   * Fix win32_spawn leaking dev_null fd in the parent process. (#906, Antonin Décimo)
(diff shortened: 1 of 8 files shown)


MisterDA:
I've just expanded a bit the scope of this PR by switching all internal pipes used in `Lwt_process`  to close-on-exec.

MisterDA:
The patch just got merged in ocaml/ocaml! I think this is ready to go right into Lwt.
The good security model is to always keep file descriptors close-on-exec. On Linux this works fine because we can `dup2` just after the `fork`, which removes the close-on-exec flag. On Windows, we have to either unset the flag or duplicate the file descriptor to unset the flag just before calling `CreateProcess`.
On Windows, `Unix.create_process` used to duplicate the file descriptor. However, if the programmer isn't aware of that, and gives a inheritable (keep-on-exec) handle to `Unix.create_process`, the parent process ends up with a duplicated handle and the child process with a leaked handle. This can cause the parent process to block on itself, or block for a child process, and all kinds of problems follow. To prevent that, we now check whether is the file descriptor is already inheritable or not, and only duplicate it if it's not.
Lwt on Windows didn't do this duplication, so it was exhibiting a different behaviour than win32unix, prone to confuse even relentless programmers. This PR brings Lwt on par with win32unix, with the above bugfix.
As a new change, I've also switched all of Lwt internal pipes to close-on-exec. With my previous patches, this makes the change portable all OCaml version, fixes concurrency and security issues, I've added tests, and it won't change what a sane user would expect (an insane user might loop over all files descriptors of a running process and see slightly less leaked fds after this patch).

raphael-proust (approved):
Once again I can't test the windows side of things, but the code LGTM. Thank you for the added tests.

PR #914, opened 2021-12-07
Enable ocamlformat

smorimoto:


MisterDA:
I'm in favour of using ocamlformat, for consistency and readability.
You left out the files using cppo (e.g. `lwt_unix.cppo.ml`) because ocamlformat doesn't support files using preprocessing, right? Is there a way we could cheat and add a formatted version of these files to the PR?
To ease git-blaming, I also suggest that we add a `.git-blame-ignore-revs` file, containing the commit id of the re-formatting. The file can then be used with `git blame --ignore-revs-file .git-blame-ignore-revs`. See some blog posts explaining the idea: [
Ignoring mass reformatting commits with git blame](https://akrabat.com/ignoring-revisions-with-git-blame/), [Exclude linting & formatting commits when running Git blame](https://tekin.co.uk/2020/09/ignore-linting-and-formatting-commits-when-running-git-blame).

src/core/lwt.ml
```diff
@@ -893,35 +848,33 @@ struct
...
-          let cleanups_deferred = callbacks.cleanups_deferred + 1 in
-          if cleanups_deferred > cleanup_throttle then begin
-            callbacks.cleanups_deferred <- 0;
-            callbacks.regular_callbacks <-
-              clean_up_callback_cells callbacks.regular_callbacks
-          end else
-            callbacks.cleanups_deferred <- cleanups_deferred)
+    ps
+    |> List.iter (fun p ->
+           let (Internal p) = to_internal_promise p in
```
aantron: This is not a helpful comment, but I'm sad about this. Unlike most code, `lwt.ml` is carefully formatted to look good, and ocamlformat is introducing things such as this indentation.

aantron: To make it potentially somewhat helpful, in the past [we applied ocp-indent](https://github.com/ocsigen/lwt/pull/409) (then the go-to formatting tool) to the almost the whole repo, but we avoided touching code that had gotten the full human attention.

Issue #916, opened 2022-01-03
Lwt.pause needs to be used three times


madroach:
Hi,

I'm working with a ssl connection. A reader is permanently watching it for data _or_ `EOF`.
After few seconds of calculations this ssl connection might have been shutdown by the remote end and an `EOF` is waiting to be read. Now I need to force Lwt to `select` so the reader can reopen the connection _before_ writing to the already half-closed connection.

I need to call `Lwt.pause` three times to enforce this:

```
        Lwt.pause () >>= fun () ->
        Lwt.pause () >>= fun () ->
        Lwt.pause () >>= fun () ->
```

I added traces to the `Lwt_main.run` mainloop:

```
Lwt.pause
Lwt_main.run: after Lwt_engine.iter
Lwt.pause
Lwt_main.run: calling leave_hooks
Lwt_main.run: restart run_loop
Lwt.pause
Lwt_main.run: enter_iter_hooks
Lwt_main.run: should_block_waiting_for_io is false
Lwt_engine: after select calling readables
Lwt_ssl.read_bytes: trying to read from SSL
```

Is this meant to work this way? Maybe this should be documented? How am I supposed to force polling of a socket for reads before writing to it?

raphael-proust:
Hello,

I think that `Lwt_unix.sleep 0.` might be a better fit for what you are looking for. (Or possibly, but I don't think you need it, `Lwt_unix.sleep Float.(succ zero)`?)

`Lwt.pause` is about cooperating with other promises rather than cooperating with the OS. Whereas `Lwt_unix.sleep` is the other way around. That's a wishy-washy way of explaining it. I think the issue lies with the lack of documentation of the main loop (`Lwt_main.run`).

madroach:
The problem is, I cannot call into Lwt_unix, only core Lwt.

Also the documentation says:

> In case your callback is just using the CPU for a really long time, you can insert a few calls to Lwt.pause into it, and resume your computation in callbacks of pause. This is basically the same as Lwt_unix.sleep0. – it's a promise that will be resolved by Lwt_main.run after any other I/O resolutions that are already in its queue.

> Lwt.pause () creates a pending promise that is fulfilled after Lwt finishes calling all currently ready callbacks, i.e. it is fulfilled on the next “tick.”
>Putting the rest of your computation into a callback of Lwt.pause () creates a “yield” that gives other callbacks a chance to run first. 

[...]

PR #917, opened 2022-01-04
Remove obsolete second wakeup_paused

madroach:
It was introduced in d5822af9a80b745515e0988541d7a43ed9a6faba.
If I understand correctly this was done to prevent calling select
without timeout while there were pending paused promises.

Since now we explicitely check for this condition and call select with
zero timeout (`should_block_waiting_for_io`). This extra `wakeup_paused`
seems to be obsolete.

Maybe in `Lwt.pause` it should also be documented that Lwt.pause () needs to be used 2 / 3 times to enforce completion of a whole main loop iteration. The current documentation talks about the 'next "tick"'? #916

### DIFF
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -1836,6 +1836,9 @@ val pause : unit -> unit t
     Putting the rest of your computation into a callback of [Lwt.pause ()]
     creates a “yield” that gives other callbacks a chance to run first.
 
+    To wait for one complete iteration of the main loop you need to wait for
+    [Lwt.pause () >>= Lwt.pause]
+
     For example, to break up a long-running computation, allowing I/O to be
     handled between chunks:
 
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -1836,9 +1836,6 @@ val pause : unit -> unit t
     Putting the rest of your computation into a callback of [Lwt.pause ()]
     creates a “yield” that gives other callbacks a chance to run first.
 
-    To wait for one complete iteration of the main loop you need to wait for
-    [Lwt.pause () >>= Lwt.pause]
-
     For example, to break up a long-running computation, allowing I/O to be
     handled between chunks:
 
--- a/src/unix/lwt_main.ml
+++ b/src/unix/lwt_main.ml
@@ -17,7 +17,7 @@ open Lwt.Infix
 let enter_iter_hooks = Lwt_sequence.create ()
 let leave_iter_hooks = Lwt_sequence.create ()
 
-let yield () = Lwt.pause ()
+let yield = Lwt.pause
 
 let abandon_yielded_and_paused () =
   Lwt.abandon_paused ()
--- a/src/unix/lwt_main.ml
+++ b/src/unix/lwt_main.ml
@@ -26,8 +26,6 @@ let abandon_yielded_and_paused () =
 
 let run p =
   let rec run_loop () =
-    (* Fulfill paused promises now. *)
-    Lwt.wakeup_paused ();
     match Lwt.poll p with
     | Some x ->
       x
@@ -40,7 +38,7 @@ let run p =
         Lwt.paused_count () = 0 && Lwt_sequence.is_empty yielded in
       Lwt_engine.iter should_block_waiting_for_io;
 
-      (* Fulfill paused promises again. *)
+      (* Fulfill paused promises. *)
       Lwt.wakeup_paused ();
 
       (* Fulfill yield promises. *)
--- a/src/unix/lwt_main.ml
+++ b/src/unix/lwt_main.ml
@@ -16,12 +16,10 @@ open Lwt.Infix
 
 let enter_iter_hooks = Lwt_sequence.create ()
 let leave_iter_hooks = Lwt_sequence.create ()
-let yielded = Lwt_sequence.create ()
 
-let yield () = (Lwt.add_task_r [@ocaml.warning "-3"]) yielded
+let yield () = Lwt.pause ()
 
 let abandon_yielded_and_paused () =
-  Lwt_sequence.clear yielded;
   Lwt.abandon_paused ()
 
 let run p =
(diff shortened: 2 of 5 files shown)


raphael-proust:
If I remember correctly, the backend-agnostic/unix-free `pause` mechanism was added in order to support `js_of_ocaml`. Does anyone know if the double waking of paused promise is necessary for compatibility (as in, to produce a more similar scheduling) as on js-of-ocaml? I doubt it because I doubt we have enough control about the scheduling of promises on different js backends that we can even dream about scheduling compatibility.

Ping @vouillon @hhugo

raphael-proust:
With this change, the behaviour of `pause` and `yield` becomes identical. I think that we should remove the code for `yield` and simply define `yield` as an alias for `pause`.

We could even remove `yield` altogether because it has been deprecated. But maybe it's worth waiting one or two releases before we remove it to smooth out the transition.

raphael-proust:
There needs to be an entry in the CHANGES file. I can add it later before merging if you prefer.

raphael-proust (commented):
(I don't know how to use github very well and I mostly had comments not related to the chunks anyway, so I left comments but apparently they don't count as a review but I think this does?)

madroach:
While preparing a CHANGES entry I found this:

>Lwt_main.yield and Lwt_unix.yield are deprecated in favor of the generic Lwt.pause, and Lwt_unix.auto_yield is deprecated in favor of the new Lwt_unix.auto_pause. Currently, Lwt_main.run resolves paused promises more frequently than yielded promises; the difference is unintended but existing applications could unintentionally depend on it. (#855, #858, Favonia)

Now I'm wondering in what way existing applications could depend on this behaviour?
I suspect they can't.

raphael-proust:
> Now I'm wondering in what way existing applications could depend on this behaviour?
> I suspect they can't.

Well for example, in your case (https://github.com/ocsigen/lwt/issues/916) you had to use three calls to `pause` where one call to `yield` sufficed. There could be an application that depends on this the same way that you do, but maybe does not know about it. As you pointed out it's not well documented/specified.

madroach:
Yes, resolving paused promises more often may break stuff intending to wait for one whole iteration. But resolving less often? No I don't think so.

madroach:
added a CHANGES entry

src/core/lwt.mli
```diff
@@ -1868,6 +1868,9 @@ val pause : unit -> unit t
     Putting the rest of your computation into a callback of [Lwt.pause ()]
     creates a “yield” that gives other callbacks a chance to run first.
 
+    To wait for one complete iteration of the main loop you need to wait for
+    [Lwt.pause () >>= Lwt.pause]
+
```
raphael-proust, line 1873: As the change of code makes obvious, this is a property of the scheduler, not the `pause` function. I'd move the documentation there because:
- this might be different on different platforms (e.g., js-of-ocaml)
- the notion of main loop is only mentioned in this file as part of a comment about `Lwt_main.run` in the "Execution model" section which is intended more as an introduction / presentation of general concepts

raphael-proust (dismissed):

raphael-proust:
Hi,

I've added a commit to your branch. We can roll it back or amend it or something. It's more of a proposal/suggestion than anything.

The idea behind it is that pause is now equivalent to yield (modulo the fact that yielded promises are resolved after which paused promises are. (So `choose [yield () >|= incr a; pause () >|= incr b; yield () >|= a;]` will actually have an intermediate state where `a=2` and `b=1`.) But I think this difference is
- undocumented
- not something anyone *should* rely on in any way
- not something anyone is likely to rely on in any way

So the commit just makes `Lwt_unix.yield = Lwt.pause`. And then it can eliminate some unneeded code.

Let me know what you think.

madroach:
I think this is sensible. There is indeed no relevant difference between yield and pause.

src/core/lwt.mli
madroach: I added the documentation to `Lwt.main`.

src/unix/lwt_main.mli
```diff
@@ -49,22 +49,20 @@ val yield : unit -> unit Lwt.t [@@deprecated "Use Lwt.pause instead"]
...
-      Unifying the two pools of promises into one in the future would eliminate
-      possible discrepancies and simplify the code. *)
-
-val abandon_yielded_and_paused : unit -> unit
+val abandon_yielded_and_paused : unit -> unit [@@deprecated "Use Lwt.abandon_paused instead"]
 (** Causes promises created with {!Lwt.pause} and {!Lwt_main.yield} to remain
     forever pending.
 
     [yield] is now deprecated in favor of the more general {!Lwt.pause}.
     Once [yield] is phased out, this function will be deprecated as well.
```
favonia, line 59: This line should be updated as well. We broke what we promised earlier, but maybe that's okay. :-)

raphael-proust: > We broke what we promised earlier, but maybe that's okay. :-)

What do you mean?

favonia: Currently it says "Once `yield` is phased out, this function will be deprecated as well.", but `yield` has not been phased out yet.

raphael-proust: Yield is deprecated. I'm not sure what "phase out" means though. So maybe we should wait longer? I'm in favour of deprecating this.

favonia: Oh, I guess that also makes sense. I was thinking about "phasing it out" as "removing it". Personally I'd love to clean this up as soon as possible regardless of what's written there. I guess we agree. :-)

raphael-proust: I updated the comment. It now simply mentions that yield is deprecated.

favonia (approved):

src/unix/lwt_main.ml
```diff
@@ -16,18 +16,14 @@ open Lwt.Infix
 
 let enter_iter_hooks = Lwt_sequence.create ()
 let leave_iter_hooks = Lwt_sequence.create ()
-let yielded = Lwt_sequence.create ()
 
-let yield () = (Lwt.add_task_r [@ocaml.warning "-3"]) yielded
+let yield () = Lwt.pause ()
```
favonia, line 20:
```suggestion
let yield = Lwt.pause
```

PR #918, opened 2022-01-05
Fully initialize custom_operations job_ops

MisterDA:
It was missing an initializer for `.compare_ext` (since 3.12.1), the
default is `custom_compare_ext_default`; and `.fixed_length` (since
4.08.0) for which `NULL` is good.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -28,6 +28,7 @@
   * Use the WSA_FLAG_NO_HANDLE_INHERIT on Windows when creating sockets with WSASocket if the cloexec (non-inheritable) parameter is true. Fixes a Windows problem where a child process would inherit a supposedly non-inheritable socket. (#910, Antonin Décimo)
   * Fix macOS/arm64 tests which have a 16k page. (#932, Kate Deplaix)
   * Fix Lwt_unix.closedir incorrectly checking the return value of closedir(3). (#942, Antonin Décimo)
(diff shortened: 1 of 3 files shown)


raphael-proust:
I don't know enough about this part of the code or the intricacies of the underlying system to meaningfully review this code.

With that in mind, my very generic questions are:
- Do we need to change `src/unix/lwt_libev_stubs.c` where `loop_ops` and `watcher_ops` do use this `struct`?
- Should we add a test with a custom job to the unix test suite?

MisterDA:
I've seen the issue when I added warning flags to the C compiler which reported fields missing initializers. 
The structs `watcher_ops` and `loop_ops` are static, so the fields are implicitly initialized to `NULL` by the compiler, which is not the case for `job_ops` as it isn't static (as far as my knowledge of C goes).
The fact that we never had a crash related to garbage values for these fields in `job_ops` makes me think that they're not used...

> Do we need to change src/unix/lwt_libev_stubs.c where loop_ops and watcher_ops do use this struct?

Seems the best to me, I missed them because the compiler didn't report a warning.

> Should we add a test with a custom job to the unix test suite?

Testing never hurts!

MisterDA:
> * Do we need to change `src/unix/lwt_libev_stubs.c` where `loop_ops` and `watcher_ops` do use this `struct`?

Done.

> * Should we add a test with a custom job to the unix test suite?

I don't think it's necessary because we're already using all the default functions which raise `Failure` when called, and the new one do too. I don't know what this test would look like.

Issue #923, opened 2022-01-23
OCaml 5.00 compatibility


patricoferris:
OCaml 5.00 (at the moment `trunk`) has removed lots of functions that were deprecated including `String.uppercase` that is used in the unix library's discover executable.

https://github.com/ocsigen/lwt/blob/1b123ffa75c48f311cded09e17c7dad690dbee3d/src/unix/config/discover.ml#L67

This needs fixed so lwt can be installed with OCaml 5.00. I have _a fix_ [here](https://github.com/ocsigen/lwt/compare/master...patricoferris:5.00) which aimed to keep things working on `4.02` (I think `uppercase_ascii` was added in `4.03`). By the looks of it, lwt supports `4.02` but isn't tested against it so maybe there's a plan to drop support, in which case `uppercase_ascii` would probably suffice? 

EDIT: I just noticed there are tests for `4.02` and they fail for my "fix" because an older version of dune is used. It probably needs something like https://github.com/janestreet/result? At any rate the issue still stands :))

EduardoRFS:
There is also the problem of Unix.fork on Lwt_process, which I guess could be replaced by create_process.

raphael-proust:
@EduardoRFS Is the issue that we can't use `Unix.fork` at all when we have domains spawned? How is `create_process` handled under the hood in 5.00?

EduardoRFS:
@raphael-proust yes that's the issue, create_process uses spawn which is implemented in C directly https://github.com/ocaml/ocaml/blob/trunk/otherlibs/unix/unix.ml#L894

But that doesn't work as Lwt_process actually cannot be described on create_process because there is no way to close fd's on the child process, so this line cannot be implemented. https://github.com/ocsigen/lwt/blob/master/src/unix/lwt_process.ml#L159

And this changes behavior considerably, so likely lwt will need to implement it's own spawn in C.

Like https://github.com/ocaml/ocaml/blob/trunk/otherlibs/unix/spawn.c#L31

patricoferris:
If I understand correctly, I think something needs done with the `bigarray` library in `lwt.unix` too. The [ocamlfind package is gone from 5.00](https://github.com/ocaml/ocaml/pull/10896) so on a 5.00.0+trunk branch it would seem `lwt.unix` will be hidden because it is `optional` and has unmet dependencies.

raphael-proust:
Lwt is now OCaml.5-compatible. Closing.

EduardoRFS:
[...]

PR #925, opened 2022-01-30
Add support for OCaml 5.00

kit-ty-kate:


kit-ty-kate:
All green and ready to go.

If you really wish to keep 4.02 compatibility see https://github.com/ocsigen/lwt/issues/923 but I personally think this is the perfect time to get rid of compatibility with this 8 years old compiler.

raphael-proust:
Thanks a lot. I had a quick browse and it looks good. I'll have a more thorough read and merge within a week, unless someone beats me to it.

raphael-proust:
I think it is also acceptable to drop 4.02 and the changes look good. It's a shame we need the hacks for <=4.08 but it's not too bad either.

Let's go!

PR #927, opened 2022-02-01
Get `Lwt_result` closer to `Stdlib.Result`

MisterDA:
`Lwt_result` has the double job of mapping fulfilled promises to `Ok` and rejected (with an exception) to `Error`, but also being useful to promises returning results. I had a little piece of code that was missing the `iter` and `error` functions, so I added them.

Add `Lwt_result`, `Lwt_result.iter`, `Lwt_result.iter_error`.
Alias` Lwt_result.map_err` and` Lwt_result.bind_lwt_err` to `Lwt_result.map_error` and `Lwt_result.bind_lwt_error` for consistency with Stdlib. Maybe that part with the deprecation is too much.

Took the liberty of adding `@since 5.6.0` to the new functions but that may not be the proper version number.

### DIFF
--- a/src/core/lwt_result.ml
+++ b/src/core/lwt_result.ml
@@ -81,9 +81,21 @@ let both a b =
     (function
       | Ok x, Ok y -> Ok (x,y)
       | Error _, Ok _
-      | Ok _,Error _ 
+      | Ok _,Error _
       | Error _, Error _ -> some_assert !s)
     (Lwt.both a b)
+
+let iter f r =
+  Lwt.bind r
+    (function
+      | Ok x -> f x
+      | Error _ -> Lwt.return_unit)
+
+let iter_error f r =
+  Lwt.bind r
+    (function
+      | Error e -> f e
+      | Ok _ -> Lwt.return_unit)
 
 module Infix = struct
   let (>>=) = bind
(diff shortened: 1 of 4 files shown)


src/core/lwt_result.ml
```diff
@@ -22,12 +23,13 @@ let map f e =
...
     e
 
-let map_err f e =
+let map_error f e =
   Lwt.map
     (function
       | Error e -> Error (f e)
       | Ok x -> Ok x)
     e
+let map_err f e = map_error f e
```
smorimoto, line 32: Why don't we mark this as deprecated as well?

src/core/lwt_result.ml
```diff
@@ -59,11 +61,12 @@ let bind_result e f =
       | Ok x -> f x)
     e
 
-let bind_lwt_err e f =
+let bind_lwt_error e f =
   Lwt.bind e
     (function
       | Error e -> Lwt.bind (f e) fail
       | Ok x -> return x)
+let bind_lwt_err e f = bind_lwt_error e f
```
smorimoto, line 69: Same

src/core/lwt_result.ml
MisterDA: Isn't the deprecation warning in the interface sufficient?

smorimoto (approved):

PR #928, opened 2022-02-02
Documentation fixes: cross-refs, internal links

MisterDA:
This PR changes all external hyperlinks to the standard API to odoc cross-references. That way, the offline Lwt doc API will correctly refer to the offline OCaml documentation. The PR also adds more cross-references to the `Unix` library, fixes the internal references to never directly use an html anchor, and adds misc. documentation fixes.
All of this is aimed at a better odoc/odig integration.

My final goal is also to convert the manual to mld, to have it installable with the opam package and browsable with odig and odoc, but that'll come in a follow-up PR.

### DIFF
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -1280,7 +1280,7 @@ val map : ('a -> 'b) -> 'a t -> 'b t
 (** [Lwt.map f p_1] is similar to {!Lwt.bind}[ p_1 f], but [f] is not expected
     to return a promise.
 
-    This function is more convenient that {!Lwt.bind} when [f] inherently does
+    This function is more convenient than {!Lwt.bind} when [f] inherently does
     not return a promise. An example is [Stdlib.int_of_string]:
 
 {[
--- a/src/core/lwt.mli
+++ b/src/core/lwt.mli
@@ -54,16 +54,15 @@ let () =
 
     This is all explained in the next sections:
 
-    - {{: #3_Quickstart} Quick start} links these concepts to actual functions
-      in Lwt – the most fundamental ones.
-    - {{: #3_Tutorial} Tutorial} shows how to write examples like the above, and
-      how concurrency happens.
-    - {{: #3_Executionmodel} Execution model} clarifies control flow when using
-      Lwt.
-    - {{: #3_GuidetotherestofLwt} Guide to the rest of Lwt} shows how
-      {e everything} else in Lwt fits into this framework.
+    - {!3_Quickstart} links these concepts to actual functions in Lwt – the most
+      fundamental ones.
+    - {!3_Tutorial} shows how to write examples like the above, and how
+      concurrency happens.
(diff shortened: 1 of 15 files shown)


src/core/lwt_stream.mli
```diff
@@ -45,10 +45,8 @@ val create : unit -> 'a t * ('a option -> unit)
 (** [create ()] returns a new stream and a push function.
 
     To notify the stream's consumer of errors, either use a separate
-    communication channel, or use a
-    {{:https://ocaml.org/api/Stdlib.html#TYPEresult}
-    [result]} stream. There is no way to push an exception into a
-    push-stream. *)
+    communication channel, or use a {!Stdlib.result} stream. There is
```
raphael-proust, line 48: I've had issues with this in some context. Specifically, with links to outer dependencies. Is this expected to works with some specific versions of odoc? With some specific environment? With some specific options?

I just tested locally to run `dune build -p lwt @doc` and I get a lot of warnings including:

```
File "src/core/lwt_stream.mli", line 48, characters 36-52:
Warning: Failed to resolve reference unresolvedroot(Stdlib).result Couldn't find "Stdlib"
```

raphael-proust:
Thanks for the many fixes. I have a significant question about the essence of the MR itself. I've attached it to a diff chunk but it's really a general question. I can do a more thorough review once it is sorted.

src/core/lwt_stream.mli
Julow: That's due to Dune not building the documentation of the dependencies. The documentation can't contain any link to the Stdlib's doc, because it is not built.

I just tried using Odig, which builds the documentation for the stdlib, and I confirm this reference resolves correctly.

src/core/lwt.mli
```diff
@@ -1916,13 +1911,13 @@ val wakeup_paused : unit -> unit
 
 val paused_count : unit -> int
 (** Returns the number of promises that would be fulfilled by calling
-    {!Lwt.wakeup_paused} right now.
+    [Lwt.wakeup_paused] right now.
```
Julow, line 1914: Interesting, the target of the reference as well as the reference itself are removed from the documentation with a stop comment but Odoc still tries to resolve the reference and fails. https://github.com/ocaml/odoc/issues/818

src/core/lwt.mli
```diff
@@ -54,24 +54,23 @@ let () =
 
     This is all explained in the next sections:
 
-    - {{: #3_Quickstart} Quick start} links these concepts to actual functions
+    - {!section-3_Quickstart} links these concepts to actual functions
```
Julow, line 57: All the `section-`, `type:` and `Lwt.` prefixes are not necessary, the references will resolve as long as there's no ambiguity.

Julow (approved):

src/core/lwt.mli
MisterDA: I've removed the unnecessary prefixes, but I'm keeping the module paths: it'd be too long to change at this point.

PR #929, opened 2022-02-04
Don't define `CAML_NAME_SPACE`, use mlvalues macros

MisterDA:
`CAML_NAME_SPACE` has no more effect in 5.00 as all the functions are prefixed with `caml_` and the old names were removed.
The mlvalues macros were introduced in 4.12 and help manipulating option and list values in C.

### DIFF
--- a/src/unix/lwt_config.h
+++ b/src/unix/lwt_config.h
@@ -31,4 +31,12 @@
 #define CAML_NAME_SPACE
 #endif
 
+#if OCAML_VERSION < 41200
+#define Val_none Val_int(0)
+#define Some_val(v) Field(v, 0)
+#define Tag_some 0
+#define Is_none(v) ((v) == Val_none)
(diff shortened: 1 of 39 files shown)


src/unix/lwt_config.h
```diff
@@ -26,4 +26,17 @@
 #define NANOSEC(buf, field) 0.0
 #endif
 
+#include <caml/version.h>
+#if OCAML_VERSION < 50000
+#define CAML_NAME_SPACE
+#endif
+
+#if OCAML_VESION < 41200
```
avsm, line 34:
```suggestion
#if OCAML_VERSION < 41200
```

MisterDA: Thanks.

raphael-proust:
Rebased and added entry to the CHANGES. Will merge once the CI passes.

PR #930, opened 2022-02-15
Require bigarray-compat to enable `lwt.unix` under OCaml 5.00

anmonteiro:
This fixes the [item](https://github.com/ocsigen/lwt/issues/923#issuecomment-1036168918) raised by @patricoferris in https://github.com/ocsigen/lwt/issues/923.

We enable `lwt.unix` if `bigarray-compat` is present, which is a new depopt.

### DIFF
--- a/lwt.opam
+++ b/lwt.opam
@@ -37,6 +37,7 @@ depopts: [
   "base-threads"
   "base-unix"
   "conf-libev"
+  "bigarray-compat"
 ]
 
 conflicts: [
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
This will be fixed by https://github.com/ocaml/dune/pull/5526. I think this can be closed

raphael-proust:
Closing this as per @kit-ty-kate 's recommendation.

PR #931, opened 2022-02-15
poc for await

c-cube:
This is not intended to be merged. It's a super basic proof of concept for `Lwt.await`. This works with `dune exec ./test.exe` on 4.12+domains+effects:

```ocaml
open Lwt.Syntax

let suspend f =
  let* () = Lwt.pause() in
  f()

let f n =
  let+ () = Lwt_unix.sleep (float n) in
  Printf.printf "sleep %d done \n%!" n

let main () =
  let l = List.init 15 f in
  suspend @@ fun () ->
  List.iter Lwt.await l; (* direct style! *)
  Lwt.return()

let () =
  Lwt_main.run @@ main ()
```

### DIFF
--- a/dune
+++ b/dune
@@ -1,0 +1,4 @@
+
+(executable
+  (name test)
+  (libraries lwt lwt.unix))
--- a/src/core/lwt.ml
+++ b/src/core/lwt.ml
@@ -549,7 +549,8 @@ struct
 end
 open Main_internal_types
 
-
+type _ EffectHandlers.eff +=
+  | Await : 'a callbacks -> 'a EffectHandlers.eff
 
 module Public_types =
 struct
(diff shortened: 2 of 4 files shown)


raphael-proust:
Interesting POC :+1:

c-cube:
This is obviously not prod-ready, but I've been wondering if there's
interest in an actual serious implementation for ocaml 5 :)

Not that I have the skill or time to do that.

raphael-proust:
Some experiments mixing events and Lwt (focused more on exploring the limits of what works and what fails when mixing, what interleaving is un/safe, what can/'t I get away with): https://gitlab.com/raphael-proust/lwt-effect-try-out/-/blob/develop/src/

Leaving this out here for future reference.

PR #932, opened 2022-02-16
Fix the tests for macOS/arm64 (which has a 16k page size)

kit-ty-kate:


raphael-proust:
This looks good to me. I'll push a commit to add it to the CHANGES and then I'll merge.

raphael-proust:
Pushed the CHANGES change (and a rebase on top of master at the same time)

PR #933, opened 2022-02-16
Add support for OCaml 5.00 (again)

kit-ty-kate:
The bigarray library was removed from OCaml 5.00 (but now included in the stdlib itself)
This bug was hidden because of a bug in dune:
* https://github.com/ocaml/dune/issues/5420
* https://github.com/ocaml/dune/pull/5421

bigarray-compat is a compatiblity library which allows us to keep compatibility with older versions of the compiler easily.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,8 +2,7 @@ BLD=../_build/default/src
 SRC=../src
 
 PKGS=\
-	-package bytes -package result \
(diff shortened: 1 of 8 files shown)


kit-ty-kate:
no the right fix is https://github.com/ocaml/dune/pull/5526 imo

PR #937, opened 2022-03-14
Fix lwt.opam

fdopen:
`os` instead of `"os"`, `"win32"` instead of `"win32 "`.
The restriction of `mmap` seems wrong, the library is used unconditionally.


smorimoto (approved):

smorimoto:
Thank you @fdopen!

PR #942, opened 2022-04-13
Fix `unix_closedir_job` incorrectly checking return value of job

MisterDA:
`LWT_UNIX_CHECK_JOB` was incorrectly checking the directory stream
instead of the return value of `closedir`.


raphael-proust:
Thanks for this fix! I've grepped for the `LWT_UNIX_CHECK_JOB` and there are a couple of suspicious

raphael-proust:
There are a couple uses of `LWT_UNIX_CHECK_JOB` which could be simplified (they use `job->result != 0` which should be the same as just `job->result` IIUC) and some that are somewhat arbitrarily specific (e.g., the check for `closedir` has  `job->result < 0` but the manpage for `closedir` mentions 0 on success and -1 on error, so we could either leave `job->result`).

Anyway, I didn't find errors in any of these other instances. So I'm merging this. (After rebase and CI.)

PR #943, opened 2022-05-12
Add bind_error to Lwt_result

bn-d:
There is `map` vs `map_error` and there is `bind_lwt` vs `bind_lwt_error`. I think having a `bind_error` would really  make sense. This is very usefully when your processes have retry, backup/alternative loading method etc.

Thanks


raphael-proust:
The changeset looks good to me. The history can use some squashing. Do you want to do it? Otherwise I can squash and push on your branch. Let me know.

bn-d:
Let's just squash and push through GitHub. Thanks

PR #946, opened 2022-05-25
Add cppo as a build dependency for lwt_react.

rizo:
Similar to: https://github.com/ocsigen/lwt/blob/654952b62dccea63f186747be79916dbfdccf48d/lwt.opam#L22

I'm not sure if it should actually be _moved_ from `lwt.opam`.


raphael-proust:
I've added an entry to the CHANGES. I think this MR is ready. @rizo do you agree? If so, I will merge.

rizo:
All good, thanks @raphael-proust!

Issue #939, opened 2022-04-11
Drop support for OCaml < 4.08.0


hannesm:
Hello,

there's quite some craft to support old OCaml releases. I'm wondering what is the reason to support old OCaml versions? Are there users thereof? Are they not fine to upgrade to a newer OCaml version? Or can't they just keep on using old Lwt releases?

Especially with OCaml 5.0.0 on the horizon, I propose to cut backwards compatibility to get rid of various dependencies on compatibility packages and code paths. What do you think? If you like the idea, I can propose a PR for this.

raphael-proust:
I like the idea. I considered it in the past and I think that we should possibly run an informal poll on discuss just to give a chance for potential users to make themselves known.

raphael-proust:
This is now done.

hannesm:
Thanks @raphael-proust for your survey, and your work on reducing the dependency cone. Also thanks for preparing the 5.6 releases :)

PR #947, opened 2022-06-02
Drop support for old ocaml versions

raphael-proust:
This PR removes support for ocaml<=4.07. This is following [a suggestion](https://github.com/ocsigen/lwt/issues/939) (from @hannesm) and [a user poll on discuss](https://discuss.ocaml.org/t/lwt-informal-user-survey/9666).


raphael-proust:
We'll need to get `ocaml-ci` to drop tests for dropped ocaml versions. That might need to wait until we merge or at least we agree to merge.

src/core/dune
```diff
@@ -1,23 +1,10 @@
...
-  in
-  let future_syntax = "(preprocess (future_syntax))" in
-  if older then
-    future_syntax
-  else
-    bisect
+   match Sys.getenv "BISECT_ENABLE" with
+   | "yes" -> "(preprocess (pps bisect_ppx))"
+   | _ -> ""
+   | exception _ -> ""
```
hannesm, line 7: I guess with recent dune (>= 2.7.0) and bisect_ppx (2.5.0) this could be done differently; namely in the `dune` files add the stanza `(instrumentation (backend bisect_ppx)))` (as explained in bisect_ppx README). :)

src/react/lwt_react.cppo.ml
```diff
@@ -5,13 +5,7 @@
...
 open Lwt.Infix
 
-let opaque_identity x =
-  #if OCAML_VERSION < (4, 03, 0)
-    x
-  #else
-    Sys.opaque_identity x
-  #endif
-
+let opaque_identity x = Sys.opaque_identity x
```
hannesm, line 8: maybe just remove this binding entirely, since below you're using `Sys.opaque_identity` explicitly anyways ;)

hannesm:
I removed some more code, see https://github.com/ocsigen/lwt/pull/949 (which is against the branch here). Please have a look :)

src/react/lwt_react.cppo.ml
raphael-proust: It is exported in the API. It is exported after a `(**/**)` so probably not a big issue to remove it. Still we might want to add a deprecation annotation.

It is also used in one of the tests but we could use the fully qualified version too.

raphael-proust:
Thanks for the additional removals! I've left one (1) comment which we may possibly ignore and merge anyway. Otherwise looks good to me.

src/react/lwt_react.cppo.ml
raphael-proust: Resolved from #949

raphael-proust:
After adding all the constraints, the ocaml-ci dropped build attempts on legacy compilers. (Thanks Tmsc!)

However, the only compiler version where all the packages are now buildable is `4.14+domains`. That's because `ocaml-ci` attempts to build all packages, including `lwt_domain.opam`. We should split the `lwt_domain.opam` in its own repository to avoid this. I'll do that in a separate PR.

dinosaure:
> However, the only compiler version where all the packages are now buildable is 4.14+domains. That's because ocaml-ci attempts to build all packages, including lwt_domain.opam. We should split the lwt_domain.opam in its own repository to avoid this. I'll do that in a separate PR.

You probably should make an issue on `ocaml-ci` I think.

tmcgilchrist:
@raphael-proust We have a fix for `ocam-ci` that allows this PR to build.

See https://github.com/tmcgilchrist/lwt/tree/drop-support-for-old-ocaml-versions for a successful build against all supported packages / ocaml versions. Note it skips lwt_domains on unsupported compilers but includes it on ocaml-4.12+domains, and will pickup OCaml 5.0 once that is available.

There are 2 changes I needed to make in LWT:
  1. Add the "base-domains" dependency to lwt_domains.opam
  2. Tag the lwt_domains tests as belonging to that package https://github.com/ocsigen/lwt/compare/drop-support-for-old-ocaml-versions...tmcgilchrist:drop-support-for-old-ocaml-versions

I didn't change the other tests but should be modified to identify which package they belong to.

raphael-proust:
I've made a merge-request (into this branch) of your branch just so I could use the GH UI to merge. Thanks a lot for that fix.

raphael-proust:
I've added one last commit to mark other tests with what packages they correspond to. But that breaks the build because there now several `main.exe` tests for the `lwt` package. So I rolled it back.

I think that a follow-up PR to modernise the `dune` files might be good. We can avoid these `executable`+`alias` in the `test/` and replace by a simple `test` stanza, we can make the change @hannesm suggests above, we can bump the version requirement for dune (to match the new version requirement for ocaml), etc.

raphael-proust:
No one has come forward with some reason to not merge this. And the MR does restore a functioning CI. So I'm merging this.

olafhering:
`lwt_ppx.opam` has now two constraints for `ocaml`.

PR #948, opened 2022-06-02
Fix CI failures

raphael-proust:
The CI has several known failures which are being ignored rn. This PR attempts to fix them all.

Note that some of the failures are triggered in the `ocaml-ci`, the configuration for which is handled separately. I'll post updates about this in comments to this PR.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -102,6 +102,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
+      - name: Hack Git CRLF for ocaml/setup-ocaml issue #529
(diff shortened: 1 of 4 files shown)


raphael-proust:
This still fails on windows+4.06, but it succeeds on windows+4.08 (https://github.com/ocsigen/lwt/runs/6704942176?check_suite_focus=true) where the fix commit appears in conjunction with some other changes which bump the minimal ocaml version (https://github.com/ocsigen/lwt/pull/947).

Possibly we don't want to merge this here MR and we want to merge the commit as part of https://github.com/ocsigen/lwt/pull/947

PR #953, opened 2022-06-13
Support OCaml 5 new Unix API names

dra27:
The compiler is considering ensuring all symbols in the Unix library are prefixed `caml_` for 5.0 (https://github.com/ocaml/ocaml/pull/10926).

This requires:
- Pre-processing to select `unix_exit` or `caml_unix_exit`
- The removal of redundant declarations for functions in `caml/socketaddr.h`
- Version-based checking for the `socket_domain_table` and `socket_type_table` internal Unix symbols

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@
 
   * Lwt is now compatible with OCaml 5.00. Lwt is now incompatible with OCaml 4.02. (#925, #923, Kate Deplaix, Patrick Ferris)
   * Lwt is now incompatible with OCaml.4.07 and earlier. (#947, Hannes Mehnert, Tim McGilchrist)
+  * Lwt-unix is now compatible with OCaml 5.0.0. (#953, David Allsopp)
 
 ====== Additions ======
 
--- a/src/unix/dune
+++ b/src/unix/dune
@@ -9,6 +9,13 @@ let preprocess =
 let () = Jbuild_plugin.V1.send @@ {|
 
 (rule
+ (targets lwt_process.ml)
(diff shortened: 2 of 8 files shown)


dra27:
The upstream OCaml PR was merged - please could someone kick GitHub actions?

tmcgilchrist:
Re-running the analysis step to pickup dune 3.3.0 so this works on OCaml 5.0. Otherwise ocaml-ci should be working for this PR. @dra27

smorimoto:
I just kicked GitHub Actions.

dra27:
The tests require a release of base with https://github.com/janestreet/base/pull/129#issuecomment-1159397380 addressed, although temporarily pinning it also revealed a missing symbol shim.

With https://github.com/dra27/lwt/commit/9d4f2b5b4, this passed ocaml-ci on 5.0

raphael-proust:
Tested locally.

It just needs an entry in CHANGES. I'll push a commit for that.

Additionally, we could also add a check for `5.0.0~alpha0` in the CI. We currently already have `4.12+domains` and `4.14+domains`. Any reason not to do that? There's the fact that we'll want to modify it as the different alphas come in and then as 5.0.0 (non-alpha) is released. But the plus side is we can make sure that further PRs don't break anything.  
I'm pushing a commit to that end. Let me know if you are ok with it.

raphael-proust:
The ci is somewhat finicky because it's an alpha release. I'm merging this now and may address the CI later.

dra27:
I wholeheartedly back adding 5.0 testing to CI 🙂

Issue #956, opened 2022-06-24
EBADF from Lwt_process.with_*


mroch:
I've received reports of `Unix.Unix_error(Unix.EBADF, "close", "")` exceptions that stem from this `process#close`:

https://github.com/ocsigen/lwt/blob/7355895d89e11aed36d2348380029c57d3124aec/src/unix/lwt_process.cppo.ml#L350

Unfortunately I don't have a repro yet to understand what the process we're calling is doing, but we're calling `Lwt_process.with_process_full` here:

https://github.com/facebook/flow/blob/38b41863de17211ae70c8bbc57854fbe7fed1ce3/src/common/lwt/lwtSysUtils.ml#L54

According to [close(2)](https://man7.org/linux/man-pages/man2/close.2.html), `EBADF` in this case means that the file descriptor that Lwt created internally is invalid. Would it make sense to catch and suppress this exception?

Whatever the underlying exception is is being swallowed. Even if close's exception is ignored, it'd be nice to rethrow the original exception with its original backtrace if possible.

mroch:
i don't think it was the root cause because (according to that PR) it broke in 9de01c9f8ddb5f4956f1473b657c055c1974642f which was in 5.6.0, but we're using 5.5.0.

i'm going to use a workaround in Flow to see if I can figure out what the underlying issue that causes the invalid FD is, so that we can hopefully find a repro/write a test.

raphael-proust:
Ok, thanks for the info. I'll try to have a look soon to check whether we could forward the original exception and what other issue might be the root cause.

mroch:
circling back with results from our workaround. these are the two underlying exceptions we've gotten reports of so far (still no actual repro unfortunately):

```
Unix.Unix_error(Unix.EBADF, "read", "")
Raised by primitive operation at Lwt_unix.read_bigarray.(fun) in file "src/unix/lwt_unix.cppo.ml", line 681, characters 8-59
Called from Lwt_unix.retry_syscall in file "src/unix/lwt_unix.cppo.ml", line 500, characters 13-24
Re-raised at LwtSysUtils.command_result_of_process.(fun) in file "flow/src/common/lwt/lwtSysUtils.ml", line 36, characters 2-160
Re-raised at Exception.reraise in file "flow/src/hack_forked/utils/core/exception.ml", line 36, characters 33-76
[...]

PR #955, opened 2022-06-23
Prepare for 5.6.0 release

raphael-proust:
Getting the repository ready for the 5.6.0 release

PR #957, opened 2022-06-27
Fix null file descriptor being closed when used as redirection for standard fd of child processes

MisterDA:
So, this is my mess-up, probably because I ignored the error, and because my tests weren't comprehensive enough.

In my eagerness of cloexec-ing file descriptors, I added one too many `O_CLOEXEC` flag. When one wants to redirect a standard file descriptor of a child process (giving `` `Dev_null`` to `~stdin`, `~stdout`, or `~stderr` arguments of functions in `Lwt_process`, my code introduced in https://github.com/ocsigen/lwt/commit/9de01c9f8ddb5f4956f1473b657c055c1974642f will actually *close* the file descriptor for the child process, which means that any read/write to this standard fd will fail. It is likely that the child process will exit with an error, the opposite of the intent, of course.

I've improved the tests, fixed the bug, simplified a bit the code and used explicit close-on-exec or keep-on-exec everywhere.

### DIFF
--- a/test/unix/dummy.ml
+++ b/test/unix/dummy.ml
@@ -1,6 +1,31 @@
+(* This file is part of Lwt, released under the MIT license. See LICENSE.md for
+   details, or visit https://github.com/ocsigen/lwt/blob/master/LICENSE.md. *)
+
+
+
+let test_input_str = "the quick brown fox jumps over the lazy dog"
+let test_input = Bytes.of_string test_input_str
+let test_input_len = Bytes.length test_input
+
+let read () =
+  let buf = Bytes.create test_input_len in
+  let rec aux n =
+    let i = Unix.read Unix.stdin buf n (Bytes.length buf - n) in
+    if i = 0 || n + i = test_input_len then
+      Bytes.equal buf test_input
+    else aux (n + i)
+  in
+  if aux 0 then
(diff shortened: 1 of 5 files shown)


MisterDA:
Probably related to #956. You'll possibly want to merge #955 before too.

test/unix/dummy.ml
```diff
@@ -1,6 +1,26 @@
...
+  let buf = Bytes.create len in
+  let rec aux n =
+    let i = Unix.read Unix.stdin buf n (Bytes.length buf - n) in
+    if n + i = len then
+      Bytes.compare buf bytes = 0
+    else aux (n + i)
+  in
+  if aux 0 then begin
+    (* try to read a bit more, should raise broken pipe *)
+    ignore (Unix.read Unix.stdin buf 0 1);
```
raphael-proust, line 15: Why do this?

MisterDA: It's not really needed, but I just want to make sure the test is reading only the `test_input_str` and nothing more.

MisterDA:
Thanks for the review, applied and rebased your suggestions.

raphael-proust:
Tested locally: the tests fail when the src changes are reverted.

PR #960, opened 2022-09-02
Add Lwt_preemptive.run_in_main_no_wait

kit-ty-kate:
Useful if you want `Lwt_preemptive.run_in_main` but does not need to wait for the result of the function to finish.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,7 +8,7 @@
 
   * Lwt.reraise an exception raising function which preserves backtraces, recommended for use in Lwt.catch (#963)
 
-  * Lwt_preemptive.run_in_main_no_wait to run a function in the main preemptive thread but without waiting for the result. (Kate Deplaix, #960)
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
Side note: personally i ended up using `Lwt_unix.send_notification` directly instead of using this function because the extra mutex weren't needed and made it a bit too slow and unpredictable but I feel like it might still be useful for someone someday.

raphael-proust:
Thanks for the contribution!

I have just quickly read the diff and it looks good. I will test the feature soon and then merge.

raphael-proust:
Considering the following changes:

- naming it `run_in_main_dont_wait` to be consistent with `dont_wait`
- adding an exception handler (otherwise exceptions raised by the lambda are lost)

I'll do that soon. Unless anyone have comments/remarks/etc. on the matter?

raphael-proust:
I implemented the two things I mention above and also rebased

raphael-proust:
@kit-ty-kate are you ok with the new interface? Does it work for your use-case?

kit-ty-kate:
Looks good by me.

PR #961, opened 2022-09-09
A prototype inclusion of effect handling into the `Lwt_main.run` function

raphael-proust:
Using two schedulers at the same time doesn't really work. Even when they're not really schedulers! And so it's unsurprising that effects and Lwt don't mix well. See [https://gitlab.com/raphael-proust/lwt-effect-try-out/-/blob/develop/src/M01pause.ml](this experiment) for examples of where it breaks down.

This PR is a prototype to allow users to pass an effect handler to the `Lwt_main.run` function which in turn installs it around the appropriate calls.


raphael-proust:
This is an experimental branch. Closing the PR. Anyone interested in these experiments, feel free to ping me.

PR #962, opened 2022-09-28
Fix broken discuss link

cuihtlauac:


raphael-proust:
Thank you for the fix!

PR #963, opened 2022-10-06
Export reraise, document recommended use

raphael-proust:
Better backtrace management primitive and recommendations.

On a dummy example, I manage to obtain accurate backtrace information. Will test on more realistic examples as well.


raphael-proust:
Following the recommendations of this MR, we get backtraces which are quite close to the ppx equivalent. There is still a little bit more noise in the form of additional occurrences of `Lwt.Miscellaneous.poll` than the ppx. But all the information from the ppx is also represent in the reraise backtraces.

A minor difference is that the "Re-raised" entries have the location of the `reraise`, whereas with the ppx they have the location of the `try%lwt`.

(Of course, the ppx is still useful for accessing all the additional keywords like `while%lwt` and such.)

PR #968, opened 2022-10-24
Remove several deprecated values, especially result-related

raphael-proust:
This PR removes some long-deprecated functions. Mostly those related to the result type.


raphael-proust:
I don't feel strongly about whether this PR should be merged. It could break some packages (which will need new constraints). OTOH these functions have been deprecated for a while.

Opinions?

MisterDA:
Regarding `Lwt.result`, it seems a good idea to use `Stdlib.Result` for all cases.

PR #969, opened 2022-10-24
Remove an unused dependency to bytes

kit-ty-kate:
lwt hasn't been supporting OCaml < 4.02 for quite a long time

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,9 @@
 ===== dev =====
 
+===== Build =====
+
+  * Remove unused dependency in dune file. (#969, Kate Deplaix)
+
 ===== 5.6.1 =====
 
 ====== Fixes ======
--- a/src/core/dune
+++ b/src/core/dune
@@ -13,7 +13,6 @@ let () = Jbuild_plugin.V1.send @@ {|
  (synopsis "Monadic promises and concurrent I/O")
  (wrapped false)
  |} ^ preprocess ^ {|
- (libraries bytes)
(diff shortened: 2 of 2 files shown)


raphael-proust:
Thanks!

I think the change should also include:

```
diff --git a/docs/Makefile b/docs/Makefile
index 5485ae43d..6fcaaa67c 100644
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,7 +2,7 @@ BLD=../_build/default/src
 SRC=../src
 
 PKGS=\
-	-package bytes -package result \
+	-package result \
 	-package bigarray -package unix \
 	-package ocaml-migrate-parsetree -package ppx_tools_versioned \
 	-package react \
```

do you concur? I can take care of adding this and the CHANGES entry to the PR before merging.

kit-ty-kate:
Yes indeed, though in this case `ocamlfind` is going to install a `bytes` package so it shouldn't matter. In the dune file it matters because `bytes` has stopped being a builtin with OCaml 5.0 (it only works because it relies on ocamlfind being installed via `ocplib-endian`)

Also I suspect this docs Makefile hasn't been run for a while (ppx_tools_versioned hasn't been updated since OCaml 4.11 i believe)

raphael-proust:
The docs makefile was run not toooooo long ago (I made a special 4.13 switch, pinned a dedicated wikidoc package, and had to set up one or two more things). But I intend to remove it anyway in not too long.

I'll merge this PR after adding a CHANGES entry.

PR #971, opened 2022-10-31
Lwt_domain: update to domainslib.0.5.0

Sudha247:
This updates the `lwt_domain` module to work with the latest version of domainslib. Please note that this is not backwards compatible in its current form.


raphael-proust:
Thanks for the update!! I ran the tests locally and it works.

I removed the workflow with 4.12+domains: it wasn't working. I'll extract the package into a different repo so that the CI runs all the tests. I'll do this in a separate MR.

PR #975, opened 2023-01-17
`Lwt_process`: document that the timeout is in seconds

MisterDA:


raphael-proust:
Thanks a lot for the fix!

PR #976, opened 2023-01-17
Fix a warning when compiling the C stubs with OCaml 5

MisterDA:
See https://github.com/ocaml/ocaml/pull/11255.


raphael-proust:
rebased on master. waiting for CI

PR #982, opened 2023-02-06
Add initial CODE_OF_CONDUCT

raphael-proust:
Adding the ocaml code-of-conduct (https://github.com/ocaml/code-of-conduct).

I've only put my own name down as a contact, but it'd be better to have additional points of contact. Anyone wants to add their name down? I think that the Lwt project is widespread enough in the community that it would make sense to add the Code of Conduct team's contact details too (ping @Sudha247).


raphael-proust:
closed in favour of https://github.com/ocsigen/lwt/pull/992

PR #984, opened 2023-02-10
Expose `Lwt_io.delete_recursively`

MisterDA:
Slight variations of this code are getting duplicated across our codebase, maybe it's best to expose it at the root in Lwt? Would another name better fit the function?

### DIFF
--- a/src/unix/lwt_io.mli
+++ b/src/unix/lwt_io.mli
@@ -543,9 +543,19 @@ val with_temp_dir :
     arguments to it. Once the temporary directory is created at [path],
     [with_temp_dir f] calls [f path]. When the promise returned by [f path] is
     resolved, [with_temp_dir f] recursively deletes the temporary directory and
-    all its contents.
+    all its contents by calling {!Lwt_io.delete_recursively}.
 
     @since 4.4.0 *)
+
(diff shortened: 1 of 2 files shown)


src/unix/lwt_io.mli
```diff
@@ -543,10 +543,20 @@ val with_temp_dir :
     arguments to it. Once the temporary directory is created at [path],
     [with_temp_dir f] calls [f path]. When the promise returned by [f path] is
     resolved, [with_temp_dir f] recursively deletes the temporary directory and
-    all its contents.
+    all its contents by calling {!Lwt_io.delete_recursively}.
 
     @since 4.4.0 *)
 
+val delete_recursively : string -> unit Lwt.t
+(** [delete_recursively path] attemps deleting the directory [path]
```
raphael-proust, line 551:
```suggestion
(** [delete_recursively path] attempts to delete the directory [path]
```

raphael-proust (commented):
LGTM apart from the doc suggestion I made. I can merge with the fix if it works for you.

MisterDA:
I've merged your doc fix into the first commit. All is fine by me!

PR #985, opened 2023-03-20
Enable to read which backend was picked by libev.

mefyl:
My use case for this is letting libev choose its backend freely, but checking it did not fallback to `select` in production as the 1024 file descriptor limit is not fit for purpose.

### DIFF
--- a/src/unix/lwt_libev_stubs.c
+++ b/src/unix/lwt_libev_stubs.c
@@ -91,6 +91,25 @@ CAMLprim value lwt_libev_init(value backend) {
   return result;
 }
 
+CAMLprim value lwt_libev_backend(value loop) {
+  switch (ev_backend(Ev_loop_val(loop))) {
+  case EVBACKEND_SELECT:
+    return Val_int(val_EVBACKEND_SELECT);
+  case EVBACKEND_POLL:
+    return Val_int(val_EVBACKEND_POLL);
+  case EVBACKEND_EPOLL:
+    return Val_int(val_EVBACKEND_EPOLL);
+  case EVBACKEND_KQUEUE:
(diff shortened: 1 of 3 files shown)


mefyl:
I'm at a loss as to why this fails in ocaml-ci with `undefined reference to 'lwt_libev_backend'` on all platforms; it absolutely works locally for me and the function is present in the C stub file 🤔

raphael-proust:
This seems like a reasonable addition. I'll look into the failure.

raphael-proust:
I started the other CI tests (I just noticed they needed an approval start) and they paint a much clearer picture: the error occurs when the code is compiled with libev deactivated.

I've pushed a commit adding some `lazy` because I believe the error occurs as a result of initialisation when libev is absent.

raphael-proust:
It doesn't work, same error. I'll dig further.

src/unix/lwt_libev_stubs.c
bnguyenvanyen, line 1: After line 243, add
```suggestion
LWT_NOT_AVAILABLE1(libev_backend)
```

bnguyenvanyen (commented):
We're still interested in having this upstream.
I don't know how to reproduce the CI failure, but I made a suggestion for the fix in the code: Use the `LWT_NOT_AVAILABLE1` macro when libev is not available, similarly to the other functions in the module.

mefyl:
The remaining CI issues seem to be the same as master AFAICS.

raphael-proust:
Thanks!  
I'm hoping to do a 5.8.1 release soon and I'll include this. (Sorry I only noticed your changes after pushing the 5.8.0 through.

raphael-proust:
Thanks a lot for the contribution <3

PR #986, opened 2023-04-17
[ci] Use the latest version of `actions/checkout`

ilankri:
To avoid the warning about deprecated Node.js 16 actions.  See https://github.com/ocsigen/lwt/actions/runs/8134580776.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -76,7 +76,7 @@ jobs:
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


raphael-proust:
Should we update to v4? Are there release notes or something that would help finding out if it's relevant?

MisterDA:
It's not particularly relevant. It just updates the Node version used to run the actions. There's been critics from the community on whether the version bump was needed, but here we are.

smorimoto (approved):

PR #987, opened 2023-04-17
[doc][ci] Use the latest version of the used actions

ilankri:


raphael-proust:
The CI fails with

```
2023-07-17 09:53.52: No .ocamlformat file found
2023-07-17 09:53.55: Job failed: No opam files found!
```

I'm guessing the first one is a warning (rather than an error). Although maybe not? FWIW, the Lwt project has never had an ocamlformat file. The code evolves relatively little and with relatively few maintainers/contributors so the benefits of automatic formatting are low. Is it required in order to use the latest CI actions?

The second one is strange. There are multiple opam files in the top-level directory of the repository. I'm wondering if there is a bug in the analysis job, maybe a consequence of not having an ocamlformat file?

ilankri:
> The second one is strange. There are multiple opam files in the top-level directory of the repository. I'm wondering if there is a bug in the analysis job, maybe a consequence of not having an ocamlformat file?

This PR is based on the branch `wikidoc` which does not have opam file. It is probably why the CI fails.

ilankri:
@smorimoto Why close this PR?

smorimoto:
I introduced Dependabot and used some tricks of setup-ocaml to speed it up.

ilankri:
> I introduced Dependabot and used some tricks of setup-ocaml to speed it up.

But these changes were made in the main branch `master`, not `wikidoc`, no?

ilankri:
> > I introduced Dependabot and used some tricks of setup-ocaml to speed it up.
> 
> But these changes were made in the main branch `master`, not `wikidoc`, no?

@smorimoto?

raphael-proust:
(I don't use github much (outside of the maintenance on this and one other repo) so I don't necessarily have the whole context for the actions and such. Still here are two cents regarding the current situation:)

Following the conversion of the manual to `mld`, we need to make major changes to the documentation generation and especially the automated part on the CI.

Some of the diff chunks of this PR are likely to be discarded. Others are probably useful (it seems?) but need to be rebased on top of more intrusive work or redone in a different context.

Issue #991, opened 2023-06-05
Need a way to share SIGCHLD with other libraries


talex5:
Currently, Lwt_unix installs a `SIGCHLD` signal handler soon after start-up. It uses this to implement `waitpid`.

This has a couple of problems. The first problem is that, due to #981, the handler corrupts memory and leads to segfaults. However, even when that is fixed it would still be useful to be able to fork processes using both Lwt and e.g. Eio. Currently, eio_posix installs a handler and then Lwt overwrites it. And even with eio_linux (which doesn't use a signal handler), Lwt's handler may reap Eio's processes before it can collect them.

Therefore, I propose adding some kind of process support to `Lwt_engine.t`. Then e.g. [lwt_eio](https://github.com/ocaml-multicore/lwt_eio) can provide an engine that delegates signal handling to Eio.

raphael-proust:
IIUC this is fixed by https://github.com/ocsigen/lwt/pull/993

@talex5 is this correct?

talex5:
Yes, it's fixed now - thanks!

PR #993, opened 2023-06-22
Make signal handling work on OCaml 5 and with Eio

### LINKED ISSUE #981: SIGSEGV on OCaml 5 due to missing SA_ONSTACK
https://github.com/ocaml/ocaml/pull/11497 says:

> First commit makes sure that signal handlers are run on the alternate signal stack. This is required since fiber stacks are generally too small to run signal handlers on.

Test case:

```ocaml
open Lwt.Infix

let main () =
  for _ = 1 to 10000 do
    match Lwt_unix.fork () with
    | 0 -> Unix._exit 1
    | _child -> ()
  done

let main () =
  Lwt_main.run begin
    Lwt.pause () >>= fun () ->
    main ();
    Lwt.return_unit
  end

let () =
  Effect.Deep.match_with main ()
    { retc = ignore;
      exnc = raise;
      effc = (fun _ -> None);
    };
  Gc.full_major ()
```
Crashes with various errors, e.g.
```
$ ./_build/default/bin/main.exe 
corrupted size vs. prev_size
fish: “./_build/default/bin/main.exe” terminated by signal SIGABRT (Abort)
```
(originally reported as https://github.com/ocaml-multicore/lwt_eio/issues/15)

### LINKED ISSUE #994: Race in worker_loop
`worker_loop` starts by masking signals, since if a signal occurs in this thread it will crash the process:

https://github.com/ocsigen/lwt/blob/90e0d6e1d68a80b016f72b0c9dbea65b6c796342/src/unix/lwt_unix_stubs.c#L968-L977

However, there is a slight chance that a signal will occur before this. Here's a test case:

```ocaml
let pid = Unix.getpid ()

let some_worker_job () =
  Lwt_unix.gethostbyname "www.google.com"

let _ =
  Sys.(set_signal sigusr1) @@ Signal_handle (fun _ -> print_endline "Got USR1");
  Lwt_main.run begin
    let job = some_worker_job () in
    print_endline "Mask USR1 in main thread to make sure worker gets it";
    ignore (Thread.(sigmask SIG_BLOCK) [Sys.sigusr1] : _ list);
    print_endline "Sending USR1";
    assert (Unix.system ("kill -USR1 " ^ string_of_int pid) = Unix.WEXITED 0);
    job
  end
```

If I add a small delay then it reliably segfaults:
```ocaml
static void *worker_loop(void *data) {
  lwt_unix_job job = (lwt_unix_job)data;

  for (int64_t i = 1; i < 10000000; i++) {
    __asm__ __volatile__("");
  }

  ...
```

```
Starting new thread                    
Mask USR1 in main thread to make sure worker gets it
Sending USR1
fish: Job 1, './_build/default/test/unix/sign…' terminated by signal SIGSEGV (Address boundary error)
```

I spotted this while trying to track down some segfaults in my program. I'm not sure if this is the cause of those, as the race looks hard to hit without the delay.

Using `pthread_attr_setsigmask_np` to set an initial mask might fix it.

talex5:
The first commit fixes #981 by using `SA_ONSTACK` with OCaml 5 so that signals do not cause memory corruption.

The second commit is a solution to #991. Before installing a signal handler, Lwt now checks first with the engine, which may now prevent Lwt from installing its handler and instead forward signals from some external handler.

This allows e.g. sharing the SIGCHLD handler with Eio.

It would also be possible to do it the other way around (so that Eio uses Lwt's signal handler). However, Lwt only notifies a single domain and if that domain is busy then the signal will be delayed, whereas Eio's signal handler notifies all interested domains immediately.

The third commit fixes #994, which causes the process to segfault if an OCaml signal handler runs in a Lwt worker thread before that thread has set the signal mask. This bug is more likely to be seen when using Lwt with Eio.

These changes should be low-risk:
- The `SA_ONSTACK` change only affects OCaml 5, where Lwt already segfaults.
- The new engine method has a default that just returns `false`, causing the existing code to behave exactly as before.
- The signal masking is essentially the same trick that OCaml itself uses when creating new threads: https://github.com/ocaml/ocaml/blob/686f49870d69f3d2246b2111e54dbe0d042aa681/runtime/domain.c#L1148-L1161

/cc @raphael-proust

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -23,11 +23,20 @@
 
   * Lwt_preemptive.run_in_main_dont_wait to run a function in the main preemptive thread but without waiting for the result. (Kate Deplaix, #960)
 
+  * Lwt_unix.handle_signal and Lwt_engine.forwards_signal to allow other IO libraries (such as Eio) to share signal handlers. (Thomas Leonard, #993, #991)
+
 ====== Build ======
 
   * Remove unused dependency in dune file. (#969, Kate Deplaix)
 
   * Fix some compilation warnings for C stubs with OCaml 5. (#976, Antonin Décimo)
+
+====== Fixes ======
+
+  * Use SA_ONSTACK on OCaml5 to avoid SIGSEGV. (Thomas Leonard, #993, #981)
+
+  * Fix race in worker loop. (Thomas Leonard, #993, #994)
+
 
 ===== 5.6.1 =====
 
--- a/src/unix/lwt_engine.ml
+++ b/src/unix/lwt_engine.ml
@@ -108,6 +108,8 @@ class virtual abstract = object(self)
   method timer_count = Lwt_sequence.length timers
 
   method fork = ()
+
+  method forwards_signal (_signum:int) = false
 end
 
 class type t = object
@@ -433,6 +435,7 @@ let readable_count () = !current#readable_count
 let writable_count () = !current#writable_count
 let timer_count () = !current#timer_count
 let fork () = !current#fork
+let forwards_signal n = !current#forwards_signal n
 
 module Versioned =
 struct
--- a/src/unix/lwt_engine.mli
+++ b/src/unix/lwt_engine.mli
@@ -53,6 +53,13 @@ val fake_io : Unix.file_descr -> unit
 val fork : unit -> unit
   (** Called internally by Lwt_unix.fork to make sure we don't get strange behaviour *)
 
+val forwards_signal : int -> bool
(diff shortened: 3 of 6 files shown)


talex5:
Eio uses a plain OCaml signal handler for SIGCHLD, so this might make #994 more likely.

talex5:
I've added a fix for #994 to this PR too (and updated the PR description).

raphael-proust:
Looks good to me.

PR #995, opened 2023-07-13
Fix `Lwt_io.read_value` on OCaml 5.1

sim642:
The Marshal header size was changed in https://github.com/ocaml/ocaml/pull/12006.


raphael-proust:
Rebased and added a changelog entry

PR #996, opened 2023-07-17
Remove luv

raphael-proust:
The luv engine is experimental and unreleased.


raphael-proust:
The main reason is that it adds noise in the form of warning for various build/test commands and in the form of additional dependencies in the CI jobs. It's not much but it's also not actively worked on so I'd rather it be re-added if/when it's being developed again.

avsm:
That sounds fine; we can always resurrect it.  But why a Changes note if it never made a release?  No users would ever have seen it.

raphael-proust:
Yeah actually the CHANGES is not necessary at all. I'm rolling this back.

aantron:
I personally think it's fine to remove it. Are there any plans to ever use this code?

@ulrikstrid?

ulrikstrid:
The time to work on this never materialised and now that OCaml 5 and Eio has landed I'm just using that. Which means there's no plans from me either.

PR #1004, opened 2023-12-02
Prepare for stricter checking in GCC 14

jamesjer:
GCC 14 will be stricter about such things as implicit function return types, implicit conversions between pointers and integers, and mismatched pointer types.  Some warnings will become errors.  The Fedora project is preparing for these changes:
- https://fedoraproject.org/wiki/Changes/PortingToModernC
- https://fedoraproject.org/wiki/Toolchain/PortingToModernC

The lwt package fails to build with the test compiler mentioned in those links.  The main symptom is missing features due to a gcc run that ends with an error, even though the feature is present.  The cause is mostly implicit conversions between pointers and integers.  In addition, due to header streamlining, `errno.h` is no longer implicitly included in `unix_getprotoby_getservby_job.c`, so include it explicitly.

I await the results of the CI run with fear and trepidation.

### DIFF
--- a/src/unix/config/discover.ml
+++ b/src/unix/config/discover.ml
@@ -652,9 +652,8 @@ struct
 
         int main()
         {
-            struct hostent *he;
-            struct servent *se;
-            he =
+            int x;
+            x =
               gethostbyname_r(
                 (const char*)NULL,
                 (struct hostent*)NULL,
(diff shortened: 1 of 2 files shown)


raphael-proust:
Thanks for the contribution and for the detailed explanations. I don't have the bandwidth to check this in depth rn. I'll try to get to it at some point.

smorimoto:
@MisterDA Do you have any ideas?

smorimoto:
Totally fine to me!

PR #1007, opened 2024-02-02
Constify C constructors and flags tables

MisterDA:
There's little value to this considering the constructors tables are static, used very locally, but this serves nonetheless as an indication to the programmer that the tables are not to be modified. Now these tables will go in the readonly segment, where they belong.

Cf https://github.com/ocaml/ocaml/pull/12223, https://github.com/ocaml/ocaml/pull/12951.

### DIFF
--- a/src/unix/lwt_unix_stubs.c
+++ b/src/unix/lwt_unix_stubs.c
@@ -465,10 +465,11 @@ failure:
   uerror("socketpair", Nothing);
 }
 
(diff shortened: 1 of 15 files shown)


smorimoto (approved):

raphael-proust:
Had a chance to have more of a look. This looks good to me. Also thanks @smorimoto for the review and the updates. I think this is mergeable.

PR #1017, opened 2024-04-30
Add lint-doc job to CI

smorimoto:


raphael-proust:
The CI failure for the doc seems to be that `lwt_sequence` is deprecated… is there a way to exclude this module from the doc? I don't know of a way to do this.

PR #1019, opened 2024-05-09
Make Lwt_seq.of_list lazier, Lwt_seq.to_list tail recursive

samer--:
This PR makes a couple of small changes to make `Lwt_seq.to_list` and `Lwt_seq.of_list` work better with large lists.

`of_list`: current version builds a large term eagerly and can fail with a stack overflow or segmentation fault for large lists. This lazier version defers the creation of `Lwt_seq.node` terms until they are needed.

`to_list`: current version is not tail recursive and can fail with long sequences. This tail recursive version (using a difference list accumulator) seems able to handle longer sequences.

### DIFF
--- a/src/core/lwt_seq.ml
+++ b/src/core/lwt_seq.ml
@@ -283,22 +283,16 @@ let unfold_lwt f u () =
   | None -> return_nil
   | Some (x, u') -> Lwt.return (Cons (x, unfold_lwt f u'))
 
-let rec of_list = function
-  | [] -> empty
-  | h :: t -> cons h (of_list t)
+let rec of_list l () =
+  Lwt.return (match l with [] -> Nil | h :: t -> Cons (h, of_list t))
 
-let rec to_list seq =
-  seq () >>= function
-  | Nil -> Lwt.return_nil
-  | Cons (x, next) ->
-    let+ l = to_list next in
-    x :: l
-let to_list seq =
-  Lwt.apply seq () >>= function
-  | Nil -> Lwt.return_nil
-  | Cons (x, next) ->
-    let+ l = to_list next in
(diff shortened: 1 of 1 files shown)


raphael-proust:
Thanks!

I think there's one regression: if the first call to `seq` fails with an exception, then the call to `to_list` raises an exception (where it used to return a failed promise).

In practice this is probably not too big a deal because the the exception is caught if the call is on the right-on-side of a bind (or inside another closure/lambda). But for now I'd rather we keep this.

Can you add a call to `Lwt.apply` on the first call to the auxiliary function?

samer--:
> Thanks!
> 
> I think there's one regression: if the first call to `seq` fails with an exception, then the call to `to_list` raises an exception (where it used to return a failed promise).
> 
> In practice this is probably not too big a deal because the the exception is caught if the call is on the right-on-side of a bind (or inside another closure/lambda). But for now I'd rather we keep this.
> 
> Can you add a call to `Lwt.apply` on the first call to the auxiliary function?

Hi, thanks for reviewing my PR!
I did try to include a call to `Lwt.apply` when generating the first element of the sequence by using the partially applied `Lwt.apply seq` as the first argument to `aux` on line 295. Will this not have the desired effect?

raphael-proust:
> I did try […]

Ah you're right! I missed this because it's not in the same "shape" as some of the other functions where the recursive function is inlined and the first call has an extra apply.

I think it's all good to go then.

raphael-proust (approved):

raphael-proust:
Thanks for the contribution!

samer--:
> Thanks for the contribution!

Thank you for your time reviewing it and for Lwt! All the best.

PR #1030, opened 2024-10-07
Wrap `caml_convert_flag_list` in `lwt_convert_flag_list` for compat

MisterDA:
See https://github.com/ocaml/opam-repository/pull/26668

### DIFF
--- a/src/unix/lwt_unix.h
+++ b/src/unix/lwt_unix.h
@@ -14,6 +14,9 @@
 #include <caml/socketaddr.h>
 #include <string.h>
 
+// The following macro is for backwards compatibility.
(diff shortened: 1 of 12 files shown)


src/unix/lwt_unix.h
```diff
@@ -15,8 +15,10 @@
 #include <string.h>
 
 #if OCAML_VERSION < 50000
```
raphael-proust, line 17:
```suggestion
// The following macro is for backwards compatibility.
// It is given an `lwt_` prefix to avoid name collisions for code which
// include both this file and alloc.h.
#if OCAML_VERSION < 50000
```

raphael-proust:
Thanks a lot for the fix!

Issue #1028, opened 2024-09-28
Adding an `Lwt_retry` library


shonfeder:
Hello! Thank you for all the work maintaining Lwt, which is a lovely package :)

In our work on the ecosystem infrastructure we have need for retries in many cases, in order to make our systems more robust against transient failures. We have written a tiny library to meet this need, currently called [Lwt_retry](https://github.com/shonfeder/ocluster/tree/master/lwt_retry/lwt_retry.mli). It depends only on `lwt` and `lwt.unix` and it amounts to a handful of combinators over `Lwt.t` and `Lwt_stream.t`. 

Would you be interested in including this library into Lwt? If so, I would be happy to open a PR and make any alterations necessary so that it would be a good fit. If not, we will likely try publish this little utility as a standalone, as we need to use it across several code bases.

To get a sense for how we are using this, you can see an [example of its usage](https://github.com/shonfeder/ocluster/blob/master/worker/cluster_worker.ml#L92-L96) and [the unit tests](https://github.com/shonfeder/ocluster/tree/master/test/test_lwt_retry.ml).

raphael-proust:
I've read the interface and had a glance at the example. It's nice. Few things, of varying level of importance:

There's general reticence in adding things into lwt because of the quite strict backwards compatibility requirement which means maintenance burden tends to accrue. `Lwt_retry` seems small and self-contained enough that we can probably add it to the main Lwt repo. I'd however recommend to add it as a separate package hosted inside the repository. This can help if in the future we need to move it off or more generally make it evolve in some way or another.

-------------------------

The function `map_retry` seems to only be used for `iter`ing for logging. I'm not sure it'd be very useful for much else. Maybe we can leave `map` but maybe we should instead offer `iter`? If a user really needs it, they can always rewrite `map_retry`.

-------------------------

One thing that needs to be made clear in the mli is whether the `on_error` returns an eager stream (i.e., one where the stream is just a data-structure holding all the retry results, where the computation goes on in the background indefinitely whether the stream is consumed or not) or a lazy stream (i.e., one where the stream is a control-flow mechanism where getting an element triggers a retry).

The documentation as-written seem to suggest the former (“The stream will continue until the computation succeeds or produces a fatal error.” kind of suggests that the stream just keeps goign whatever happens). But the type of `with_sleep` and `n_times` suggests the latter.

[...]

PR #1032, opened 2024-10-15
Add Lwt_retry library

shonfeder:
Adds the `Lwt_retry` library discussed in #1028 as a separate package. I believe all guidance suggested there by @raphael-proust has been incorporated. Thanks for that!

I did not add an `iter` function, since I think this doesn't add any value over using `Lwt_stream.iter` directly.

I've tried to make the code, documentation, and tests consistent with the style I found in other locations in the code base.


lwt_retry.opam
```diff
@@ -0,0 +1,31 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "Utilities for retrying Lwt computations"
+maintainer: [
+  "Raphaël Proust <code@bnwr.net>" "Anton Bachin <antonbachin@yahoo.com>"
+]
+authors: ["Jérôme Vouillon" "Jérémie Dimino"]
```
raphael-proust, line 7: I think you can give credit where credit is due (yourself? some colleagues of yours?)

lwt_retry.opam
```diff
@@ -0,0 +1,31 @@
+# This file is generated by dune, edit dune-project instead
+opam-version: "2.0"
+synopsis: "Utilities for retrying Lwt computations"
+maintainer: [
+  "Raphaël Proust <code@bnwr.net>" "Anton Bachin <antonbachin@yahoo.com>"
```
raphael-proust, line 5:
```suggestion
  "Raphaël Proust <code@bnwr.net>" "Shon Feder <shon.feder@gmail.com>"
```

src/retry/lwt_retry.ml
```diff
@@ -0,0 +1,64 @@
...
+let on_error
+    (f : unit -> ('ok, ('retry, 'fatal) error) result Lwt.t)
+  : ('ok, 'retry, 'fatal) attempt Lwt_stream.t
+  =
+  let i = ref 0 in
+  let stop = ref false in
+  Lwt_stream.from begin fun () ->
+    incr i;
+    let+ result = f () in
+    if !stop then None else
```
raphael-proust, line 42: Not sure if I've missed something (shouldn't review in the evening probably) but it seems that you call `f` "one too many time". Like, if the `stop` ref is `true`, you still call `f`, you just don't use the result.

Is this intended?

src/retry/lwt_retry.ml
```diff
@@ -0,0 +1,64 @@
...
+      match result with
+      | Error (`Retry _ as retry) -> Some (Error (retry, !i))
+      | Error (`Fatal _ as fatal) -> stop := true; Some (Error (fatal, !i))
+      | Ok _ as ok -> stop := true; Some ok
+  end
+
+let with_sleep ?(duration=default_sleep_duration) (attempts : _ attempt Lwt_stream.t) : _ attempt Lwt_stream.t =
+  attempts
+  |> Lwt_stream.map_s begin function
+    | Ok ok -> Lwt.return_ok ok
```
raphael-proust, line 52: nano-optimisation

```suggestion
    | Ok _ as ok -> Lwt.return ok
```

Should perform one fewer allocation.

src/retry/lwt_retry.ml
```diff
@@ -0,0 +1,64 @@
...
+    | Error (err, n) ->
+      let* () = Lwt_unix.sleep @@ duration n in
+      Lwt.return_error (err, n)
```
raphael-proust, line 55: Same here

```suggestion
    | Error (_, n) as err ->
      let* () = Lwt_unix.sleep @@ duration n in
      Lwt.return err
```

src/retry/lwt_retry.ml
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+let n_times n attempts =
+  (* The first attempt is a try, and REtries start counting from n + 1 *)
```
raphael-proust, line 59:
```suggestion
  (* The first attempt is a try, and Retries start counting from n + 1 *)
```

src/retry/lwt_retry.ml
```diff
@@ -0,0 +1,64 @@
...
+  let retries = n + 1 in
+  let+ attempts = Lwt_stream.nget retries attempts in
+  match List.rev attempts with
+  | last :: _ -> last
+  | _ -> failwith "Lwt_retry.n_times impossible"
```
raphael-proust, line 64: Have you tried feeding negative numbers of `n`?

I'd recommend adding `if n < 0 then raise (Invalid_argument "Lwt_retry.n_times: n must be positive or null");` at the beginning of the function.

src/retry/lwt_retry.mli
```diff
@@ -0,0 +1,155 @@
...
+  int ->
+  ('ok, 'retry, 'fatal) attempt Lwt_stream.t ->
+  ('ok, 'retry, 'fatal) attempt Lwt.t
+(** [n_times n attempts] is [Ok v] if one of the [attempts] succeeds within [n]
+    retries (or [n+1] attempts), [Error (`Fatal f, n+1)] if any of the attempts
+    results in the fatal error, or [Error (`Retry r, n+1)] if all [n] retries are
+    exhausted and the [n+1]th attempt results in a retry error.
+
+    In particular [n_times 0 attempts] will *try* 1 attempt but *re-try* 0, so
+    it is guaranteed to produce some result.
```
raphael-proust, line 143: It's sort of implied but I think it would be good to say it explicitly somewhere: `n_times` forces the (n first elements of the) on-demand stream of attempts.

Streams are one of the most confusing part of Lwt's core library so better be explicit about everything.

raphael-proust (commented):
Looks good. Some small nitpick-ish comments and a couple of actual suggestions. Feel free to push back though, I could very well have gotten something wrong.

lwt_retry.opam
shonfeder: Whoops! Oversight here. Sorry! Thanks for catching this :)

src/retry/lwt_retry.ml
shonfeder: You are right! Thanks for catching this (embarrassing) mistake. Fixed in fc49af070adc60199866af294607da6342b15737 .

src/retry/lwt_retry.ml
shonfeder: Fixed in c8b1f3977b306a08ae1c04f5730364c89420fcc5 .

src/retry/lwt_retry.ml
shonfeder: Fixed in 0cc9b60db7165839c5ed645c8c5463dcbd6b7e39

src/retry/lwt_retry.ml
shonfeder: Good catch! Fixed in ade04564e371f8b0d4c9fa769ba4a6077b0f2418

src/retry/lwt_retry.mli
shonfeder: LMK if you think 34aefc14bde3d66f1640b3c51684defdec6456a7 addresses this!

shonfeder (commented):
Thank you for the careful review! I think I've addressed all the suggestions and corrections.

lwt_retry.opam
shonfeder: Ah, this is just generated based on the dune-project, which is why I missed.

shonfeder: Hmm, this is actually trickier than I'd expect, as dune doesn't support `maintainers` or `authors` fields on `package` stanzas until dune 2.0. But there are several other aspects of the lwt dune config that are incompatible with dune 2.0. Most are trivial so far, but some have to do with the C FFI:

```
File "test/retry/dune", lines 8-9, characters 1-33:
8 |  (action
9 |   (run %{exe:main.exe})))
File "_build/.dune/default/src/unix/dune", line 178, characters 1-43:
178 |  (c_flags -I. (:include unix_c_flags.sexp))
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: 'c_flags' was deleted in version 2.0 of the dune language. Use the
(foreign_stubs ...) field instead.
Had 1 error, waiting for filesystem changes...
```

So I think our options here are:

1. Leave the maintainer/authors package as is, generated from the data in the dune-project file.
2. use an opam.template file for the retry package.
3. Update lwt to use dune 2.0, and put the data in the `package` stanza.

3 seems like the right (tm) way to do this to me. If you agree, I can make that change in a separate PR then rebase this on top. WDYT?

raphael-proust: For 3 I need to first check the dependency cone for dune.2+ (minimal ocaml version and other such minimal version constraints). One of the thing is we try to keep maximum compatiblity because we treat Lwt as a kind of foundational library.

Depending on the compatibility, I'd go for either 2 or 3. I can check that some times in the coming week.

src/retry/lwt_retry.ml
raphael-proust: :+1: nice adding a test too

src/retry/lwt_retry.ml
raphael-proust: ah I see. Either is fine then.

src/retry/lwt_retry.mli
raphael-proust: LGTM

raphael-proust:
Thanks! I'll have a look at the dune version thing this week

lwt_retry.opam
MisterDA: Dune doesn't have external dependencies. It requires (at least) ocaml 4.08 and has a fallback through a secondary ocaml compiler for earlier versions. I don't think there's anything to be concerned here with.

shonfeder: Thank, @MisterDA . Opened a PR for the update at https://github.com/ocsigen/lwt/pull/1035

shonfeder:
Rebased on #1035 so I could update the package author and maintainer data. I'll need to rebase on master once that is merged in.

The only CI failures remain the FreeBSD and MacOS failures that are occurring in the trunk:

```
Test 'getgrgid and Unix.getgrgid' in suite 'lwt_unix' raised 'Not_found'
Test 'getgrnam and Unix.getgrnam' in suite 'lwt_unix' raised 'Not_found'
```

raphael-proust:
> I'll need to rebase on master once that is merged in

you can go ahead @shonfeder

shonfeder:
Rebased! Should be good for a final review and/or merge! Thanks again for the reviews and guidance along the way, @raphael-proust :)

shonfeder:
Anything else you'd like to see here @raphael-proust ?

dune-project
```diff
@@ -12,6 +12,17 @@
...
+(package
+ (name lwt_retry)
+ (synopsis "Utilities for retrying Lwt computations")
+ (authors "Shon Feder")
+ (maintainers
+   "Raphaël Proust <code@bnwr.net>"
+   "Shon Feder <shon.feder@gmail.com>")
+ (depends
+  (ocaml (>= 4.08))
+  lwt))
```
raphael-proust, line 24:
```suggestion
  (lwt (>= 5.3.0))))
```

because of `Lwt.Syntax`

lwt_retry.opam
```diff
@@ -0,0 +1,31 @@
...
+]
+authors: ["Shon Feder"]
+license: "MIT"
+homepage: "https://github.com/ocsigen/lwt"
+doc: "https://ocsigen.org/lwt"
+bug-reports: "https://github.com/ocsigen/lwt/issues"
+depends: [
+  "dune" {>= "2.0"}
+  "ocaml" {>= "4.08"}
+  "lwt"
```
raphael-proust, line 15:
```suggestion
  "lwt" {>= "5.3.0"}
```

matching change

test/retry/main.ml
```diff
@@ -0,0 +1,158 @@
...
+        let attempts = Retry.(operation |> on_error) in
+        try
+          let _ = Retry.(attempts |> n_times invalid_negative_retries) in
+          false (* We failed to raise the invalid argument exception *)
+        with
+          Invalid_argument _ -> true);
+
+    (* test that the sleeps actually throttle computations as desired *)
+    test "with_sleep really does sleep"  (fun ()  ->
+      let duration _ = 0.1 in
```
raphael-proust, line 147:
```suggestion
      let duration _ = 0.01 in
```

This is the bottleneck for the test suite and it doesn't make sense to go for tenth of a seconds order of magniture.

raphael-proust (changes requested):

raphael-proust (approved):

raphael-proust:
Ready to merge unless the CI shows I messaed up something with my suggested+applied changes

shonfeder:
Thank you for the fixes!

raphael-proust:
Thanks for the contribution!!

PR #1033, opened 2024-10-21
Prepare for 5.2 AST bump

patricoferris:
This PR is a patch for an upcoming release of ppxlib where the internal AST is [bumped from 4.14 to 5.2](https://github.com/ocaml-ppx/ppxlib/pull/514). Until that is merged and released, there is no reason to merge this PR.

To test these changes, you can use the following opam-based workflow. I've made releases of most of these patches to an opam-repository overlay.

```shell
opam switch create ppxlib-bump --repos=ppxlib=git+https://github.com/patricoferris/opam-repository#5.2-ast-bump
opam install <your-package>
```

The following describes the most notable changes to the AST.

Note: no update has been made of the `opam` file, but `ppxlib` will likely need a lower-bound before merging/releasing.

### Functions
---

#### Currently 

In the parsetree currently, functions like:

```ocaml
fun x y z -> ...
```

Are represented roughly as

```ocaml
Pexp_fun(x, Pexp_fun (y, Pexp_fun(z, ...)))
```

Functions like:

```ocaml
function A -> ... | B -> ...
```

Are represented roughly as

```ocaml
Pexp_function ([ case A; case B ])
```

#### Since 5.2

All of these functions now map to a single AST node `Pexp_function` (note, this is the same name as the old cases function). The first argument is a list of parameters meaning:

```ocaml
fun x y z -> ...
```

Now looks like:

```ocaml
Pexp_function([x; y; z], _constraint, body)
```

And the `body` is where we can either have _more_ expressions (`Pfunction_body _`) or cases (`Pfunction_cases _`). That means:

```ocaml
function A -> ... | B -> ...
```

Has an empty list of parameters:

```ocaml
Pexp_function([], _, Pfunction_cases ([case A; case B]))
```

### Local Module Opens for Types

Another feature added in 5.2 was the ability to locally open modules in type definitions. 

```ocaml
module M = struct
  type t = A | B | C
end

type t = Local_open_coming of M.(t)
```

This has a `Ptyp_open (module_identifier, core_type)` AST node. Just like normal module opens this does create some syntactic ambiguity about where things come from inside the parentheses.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -54,11 +54,20 @@ jobs:
       - run: opam install conf-libev
         if: ${{ matrix.libev == true }}
 
-      - run: opam install . --deps-only --with-test
+      - run: opam install ./lwt.opam ./lwt_react.opam ./lwt_retry.opam --deps-only --with-test
 
-      - run: opam exec -- dune build
+      - run: opam install ./lwt_ppx.opam --deps-only --with-test
+        if: ${{ contains(fromJSON('["5.1","5.2","5.3"]'), matrix.ocaml-compiler) }}
 
-      - run: opam exec -- dune runtest
+      - run: opam exec -- dune build -p lwt,lwt_react,lwt_retry
+
+      - run: opam exec -- dune build -p lwt_ppx
+        if: ${{ contains(fromJSON('["5.1","5.2","5.3"]'), matrix.ocaml-compiler) }}
+
+      - run: opam exec -- dune runtest -p lwt,lwt_react,lwt_retry
+
+      - run: opam exec -- dune runtest -p lwt_ppx
+        if: ${{ contains(fromJSON('["5.1","5.2","5.3"]'), matrix.ocaml-compiler) }}
 
   lint-opam:
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


raphael-proust:
I've tested after

- rebased on top of recent master
- installed ppxlib.0.36.0

The only issue is that ppxlib.0.36.0 is incompatible with ppx_let (which is used in tests, to make sure that lwt exports the values that ppx_let needs).

Ultimately, the ppx_let test could be run in a different, separate environment. It's a test of lwt only. But this is a bit difficult to do from a purely technical pov…

Maybe we can wait until ppx_let releases a version compatible with ppxlib.36.0

patricoferris:
PR for ppx_let https://github.com/janestreet/ppx_let/pull/16 (which also needs the PR to ppxlib_jane https://github.com/janestreet/ppxlib_jane/pull/5).  Hopefully these get reviewed/merged promptly and we can get a release of the library.

kit-ty-kate:
Now that the janestreet packages have been updated and merged into opam-repository, could this PR be looked at again?
This is required for lwt_ppx to be compatible with OCaml 5.4

smorimoto:
We have to send a patch to `bisect_ppx` as well, I guess.

patricoferris:
I'm taking a look at bisect_ppx today

kit-ty-kate:
Is bisect_ppx really necessary to merge this? I really don't think such CI tool should block the release of such a key ecosystem package.

PR #1036, opened 2024-10-23
Lwt_stream: non-Lwt version of junk_old

mebsout:
`junk_old` does not need to be in the Lwt monad, it is replaced by `junk_available`. `junk_old` is kept for API backward compatibility.


raphael-proust:
Thanks for the PR :heart: . It looks good at a quick glance. I'll review soon.

raphael-proust:
I added some tests for junk_available and junk_old. Seems like there weren't any before. I made something very simple.

mebsout:
Thank you!

PR #1037, opened 2024-10-29
Remove unused opam templates

smorimoto:


lwt.opam
```diff
@@ -31,7 +31,7 @@ depends: [
 depopts: ["base-threads" "base-unix" "conf-libev"]
 dev-repo: "git+https://github.com/ocsigen/lwt.git"
 build: [
-  ["dune" "subst"] {dev}
+  ["dune" "subst"] {pinned}
```
shonfeder, line 34: afaik, this should always be

```suggestion
  ["dune" "subst"] {dev}
```

see https://github.com/ocaml/dune/pull/3647

smorimoto: dune 3.0 seems to fix that.

shonfeder: Nice. Thanks :)

Issue #839, opened 2021-02-01
Dead links in the readme


Drup:
There are deadlinks in the readme, notably some documentation ones.

Also, since Discord is mentioned in the readme, it would be good if one of the maintainers hanged around there. :)

aantron:
Regarding Discord, another option is to remove it instead. I put it in there because I monitored Discord.

arvidj:
See #1038

PR #1038, opened 2024-11-01
Some clean up in CONTRIBUTING.md and fix broken links & redirects

arvidj:
This PR:
 - updates the recommendation for how to set up a switch for lwt to recommend ocaml 4.08 (lowest supported OCaml version IIUC)
 - removes mentions of appveyor and travis-ci as these are replaced by github actions
 - fixes broken links and redirects, and reference-style markdown links that were not used


arvidj:
Fixes #839

test/unix/test_lwt_engine.ml
```diff
@@ -6,28 +6,6 @@
...
-          with Not_found -> false
-        in
-
-        let in_appveyor =
-          try ignore (Sys.getenv "APPVEYOR_REPO_COMMIT"); true
-          with Not_found -> false
-        in
-
-        if not (in_travis || in_appveyor) then Lwt.return_true
-        else Lwt.return Lwt_config.libev_default);
```
MisterDA, line 26: Instead of deleting this, you could check whether the `CI` env var is `true` or `True`, or check whether `GITHUB_ACTIONS` is `true`.

MisterDA (changes requested):

test/unix/test_lwt_engine.ml
arvidj: I wasn't quite sure of the meaning of this test. Since `in_travis` and `in_appveyor` would always be false, it's basically a noop. I can do what you propose, and then the meaning of the test would be "if we're in github actions, then check that `Lwt_config.libev_default` is true". Is that what we want the test to check (if so, why do we want to test that?)?

raphael-proust:
Thanks a lot for the long overdue fixes

test/unix/test_lwt_engine.ml
raphael-proust: It's probably left-over from special-casing for different CIs and I think we can remove it.

PR #1040, opened 2024-11-15
Add `Lwt_result.collect` without deps

seprov:
# What

This PR adds `Lwt_result.collect` without adding ocaml-containers as a dependency.

# Why

To handle the common case where a list of results need to be collected to a single meaningful result.

# Notes

`partition_filter_map` and `split_result` are available in ocaml-containers, but to avoid adding the dependency, I've recreated them here.

Also, I'm not sure how y'all like to mention future releases in documentation comments. Let me know about that, thanks.


src/core/lwt_result.ml
```diff
@@ -104,6 +104,29 @@ let iter_error f r =
...
+let split_result results =
+  partition_filter_map
+    (fun x -> match x with Ok o -> `Left o | Error e -> `Right e)
+    results
+
+let collect x =
+  x |> Lwt.all
+  |> Lwt.map (fun r ->
+          let oks, errors = split_result r in
+          match errors with [] -> Ok oks | _ -> Error errors)
```
raphael-proust, line 128: On a quite specific note: This seems like a simple yet costly way to implement collect. E.g., the `oks` list is reversed (line 110) whether there are errors or not.

On a more general note: the fact that the Lwt part and the result part of this function are entirely independent makes me wonder whether it actually belongs in Lwt.

seprov: Thank you for the feedback.

> This seems like a simple yet costly way to implement collect

You're right. If we wanted to add `collect` to `Lwt_result`, then I'd be happy to optimize this.

> the fact that the Lwt part and the result part of this function are entirely independent makes me wonder whether it actually belongs in Lwt

You're also right that the Lwt and result parts are distinct. If, for example, `collect` were in `CCResult` as 

```ocaml
let collect x =
  let oks, errors = split_result x in
  match errors with [] -> Ok oks | _ -> Error errors
```

then usage might look like

```ocaml
my_data_list
|> request_for_each
|> Lwt.all |> Lwt.map CCResult.collect
```

which isn't much different from or less convenient than

```ocaml
my_data_list
|> request_for_each
|> Lwt_result.collect
```

If you think that, even with a more efficient implementation of `result list` filtering and partitioning, this code does not belong in Lwt, then I'd be happy to close this PR.

raphael-proust: I'm not too keen on adding features to Lwt so I'm undecided. It seems like it could be useful but also not really a core functionality… And making it a core functionality is potentially going to make it more confusing (e.g., what happens with failures/cancelation) than encouraging users to just call `Lwt.all` explicitely (better documented and tested) to pipe into a collection function.

I'm tending towards a "weak no", as in, I could be convinced otherwise but as it stands I don't think it belongs in Lwt itself.

seprov: Alright, I've gone ahead and closed this PR. Thanks for considering this change.

PR #1053, opened 2025-03-07
dune-project fixes from 5.[89].1

raphael-proust:
- add explicit version field
- ~incompatible with ppxlib.0.36.0~


dune-project
```diff
@@ -26,10 +26,11 @@
 
 (package
  (name lwt_ppx)
+ (version dev)
```
smorimoto, line 29: I don't think this is needed.

raphael-proust: I added it mostly because then it's easier to remember to set the `version` field when doing a release. Otherwise I have to remember to add the `version` field which I'm likely to forget.

Probably should just write a release script, but in the meantime I'd rather have that.

PR #1054, opened 2025-03-29
Fix opam constraint

hhugo:
following https://github.com/ocaml/opam-repository/pull/27693

lwt_ppx uses `Lwt.reraise` which was introduced in Lwt 5.7


smorimoto:
Somewhat this change is included in the latest master, so I will close this PR.

PR #1060, opened 2025-07-08
add the `lwt_direct` package, for direct-style control flow

c-cube:
This is a WIP, it needs testing, but the fundamentals are here.

### DIFF
--- a/src/direct/lwt_direct.mli
+++ b/src/direct/lwt_direct.mli
@@ -44,7 +44,10 @@ val spawn : (unit -> 'a) -> 'a Lwt.t
 
     When [f ()] terminates (successfully or not), the promise
     [spawn f] is resolved with [f ()]'s result, or the exception
-    raised by [f ()]. *)
+    raised by [f ()].
+
+    The promise returned by [spawn f] is not cancellable. Canceling it
+    will have no effect. *)
 
 val spawn_in_the_background :
   (unit -> unit) ->


src/direct/lwt_direct.mli
```diff
@@ -0,0 +1,33 @@
...
+  unit
+(** [run_in_the_background f] is similar to [ignore (run f)].
+    The computation [f()] runs in the background in the event loop
+    and returns no result.
+    @param on_uncaught_exn if provided, this is called when [f()]
+    raises an exception. *)
+
+val yield : unit -> unit
+(** Yield to the event loop.
+    Can only be used inside {!run} or {!run_in_the_background}. *)
```
raphael-proust, line 26:
```suggestion
    calling [yield] outside of {!run} or {!run_in_the_background} will raise an exception,
    crash your program, or otherwise cause errors. It is a programming error to do so. *)
```

Or something like that. I want it to be non-commital on what happens but also avoid "cannot do that" in documentation (it's never very clear).

src/direct/lwt_direct.mli
```diff
@@ -0,0 +1,33 @@
...
+
+val await : 'a Lwt.t -> 'a
+(** [await prom] returns the result of [prom], or re-raises the
+    exception with which [prom] failed if it failed.
+    If [prom] is not resolved yet, [await prom] will suspend the
+    current task and resume it when [prom] is resolved.
+    Can only be used inside {!run} or {!run_in_the_background}. *)
```
raphael-proust, line 33:
```suggestion
    calling [yield] outside of {!run} or {!run_in_the_background} will raise an exception,
    crash your program, or otherwise cause errors. It is a programming error to do so. *)
```

src/direct/lwt_direct.mli
raphael-proust: Maybe we can put `yield` after `await` and mention something along the lines of

> `yield` is similar to `await (Lwt.pause ())` but more efficient.

src/direct/lwt_direct.ml
```diff
@@ -0,0 +1,95 @@
...
+    ready but [tasks] contains ready tasks *)
+  if not (Queue.is_empty tasks) then ignore (Lwt.pause () : unit Lwt.t)
+
+let setup_hooks =
+  let already_done = ref false in
+  fun () ->
+    if not !already_done then (
+      already_done := true;
+      let _hook1 = Lwt_main.Enter_iter_hooks.add_first run_all_tasks in
+      let _hook2 = Lwt_main.Leave_iter_hooks.add_first run_all_tasks in
```
raphael-proust, line 38: I wonder if it should be both enter and leave hooks. Doing both makes the yield different from Lwt_unix.yield.

This will take n loops of the scheduler
```
(* Lwt_unix.yield is just an alias for Lwt.pause *)
let rec vanilla_yield_n_times n () = if n < 0 then Lwt.return_unit else Lwt.bind (Lwt.pause ()) (unix_yield_n_times (n-1))
```

and this will take n/2
```
let rec direct_yield_n_times n = if n < 0 then () else (Lwt_direct.yield (); direct_yield_n_times (n-1))
```


Is there a reason you added both?

In a very handwavy way the enter hook is good for tasks that await on promises that are about to be resolved this run of the scheduler (e.g., `pause ()`, but also whatever promise depends on the I/O that has completed and triggered the shceuler run), whilst the leave hook is good for tasks that await on actual I/O.

c-cube:
I added some tests!

c-cube:
```
5 |  (build_if (>= %{ocaml_version} "5.0"))
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: 'build_if' is only available since version 3.9 of the dune language.
```
:cry: not sure how to make CI happy on OCaml 4 .xx at this rate…

src/direct/lwt_direct.ml
```diff
@@ -0,0 +1,117 @@
...
+let[@inline] push_task f : unit = Queue.push f tasks
+
+let default_on_uncaught_exn exn bt =
+  Printf.eprintf "lwt_task: uncaught task exception:\n%s\n%s\n%!"
+    (Printexc.to_string exn)
+    (Printexc.raw_backtrace_to_string bt)
+
+let run_all_tasks () : unit =
+  let n_processed = ref 0 in
+  let max_number_of_steps = min 10_000 (2 * Queue.length tasks) in
```
raphael-proust, line 32: I think the literal 10_000 is ok for the alpha release. But we'll want to think more about what's the right value at some point and document it. In the meantime we should probably leave a placeholder for a comment and all that jazz just in preparation.

```suggestion
let absolute_max_number_of_steps =
  (* *)
  10_000

let run_all_tasks () : unit =
  let n_processed = ref 0 in
  let max_number_of_steps = min absolute_max_number_of_steps (2 * Queue.length tasks) in
```

src/direct/lwt_direct.ml
```diff
@@ -0,0 +1,117 @@
...
+    let t = Queue.pop tasks in
+    incr n_processed;
+    try t ()
+    with exn ->
+      let bt = Printexc.get_raw_backtrace () in
+      default_on_uncaught_exn exn bt
+  done;
+  (* make sure we don't sleep forever if there's no lwt promise
+    ready but [tasks] contains ready tasks *)
+  if not (Queue.is_empty tasks) then ignore (Lwt.pause () : unit Lwt.t)
```
raphael-proust, line 43: The idea here is to put a trivial promise in the shceduler just so that the scheduler has a reason to kick-in again so that it doesn't wait for IO before it processes the tasks again?

I'll rewrite the comment to be a bit more explicit (I don't trust my future self to grasp this quickly).

raphael-proust (commented):
Very nice work.

I'll want to do a light editing pass with some renaming and such to make the code more homogeneous (e.g., in run "resolve" and "promise" should have consistent name). But I'll do that after we discuss the points that need discussing and we resolve some questions.

src/core/lwt.mli
```diff
@@ -2061,3 +2061,14 @@ val backtrace_try_bind :
 val abandon_wakeups : unit -> unit
 
 val debug_state_is : 'a state -> 'a t -> bool t
+
+module Private : sig
```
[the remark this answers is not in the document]
c-cube: That's a great point, it could be `Private_lwt_keep_away` with an alert.

lwt_direct.opam
```diff
@@ -0,0 +1,34 @@
...
+authors: ["Jérôme Vouillon" "Jérémie Dimino"]
+license: "MIT"
+homepage: "https://github.com/ocsigen/lwt"
+doc: "https://ocsigen.org/lwt"
+bug-reports: "https://github.com/ocsigen/lwt/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "5.0"}
+  "base-unix"
+  "lwt" {>= "3.0.0"}
```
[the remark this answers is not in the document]
c-cube: Ah, but with the changes related to storage we'll need 6.0 or more actually :(

src/direct/lwt_direct.ml
```diff
@@ -0,0 +1,117 @@
...
+
+let run_inside_effect_handler_and_resolve_ (type a) (promise : a Lwt.u) f () : unit =
+  let run_f_and_set_res () =
+    let res = ref (Error (Failure "not resolved")) in
+    Storage.reset_to_empty();
+    (try
+       let r = f () in
+       res := Ok r
+     with exn -> res := Error exn);
+    Lwt.wakeup_result promise !res
```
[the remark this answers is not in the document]
c-cube: Ah yes, fair enough.

src/direct/lwt_direct.mli
```diff
@@ -0,0 +1,94 @@
...
+
+(** Local storage.
+
+    This storage is the same as the one described with {!Lwt.key},
+    except that it is usable from the inside of {!run} or
+    {!run_in_the_background}.
+
+    Each task has its own storage, independent from other tasks or promises. *)
+module Storage : sig
+  type 'a key = 'a Lwt.key
```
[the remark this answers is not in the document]
c-cube: - I think it's good to be able to use existing keys
- it's deprecated but why? Local storage is important!! For logging and tracing at least. Eio has it too :)

lwt_direct.opam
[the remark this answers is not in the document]
raphael-proust: ahh well true

unless we take the 6.0 release as a reason to remove the storage altogether… it's been deprecated for a while now

let's put 6+ for now and we'll lower it if we decide to remove keys and local storage

src/direct/lwt_direct.mli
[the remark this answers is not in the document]
raphael-proust: yes I think the stoarge makes more sense for direct-style, because of the `run` wrapper gives it a more explicit scope. let's keep it for now and we'll see later

kit-ty-kate:
I tested alpha01 and got a segfault (it didn't in alpha00 but cohttp didn't seem to work):
```
Thread 2 "opam-health-serve-ng" received signal SIGSEGV, Segmentation fault.                                   
[Switching to LWP 4945]                                                                                   
0x00005555561aa77e in lwt_unix_send_notification (domain_id=0, id=2147482650) at lwt_unix_stubs.c:577     
warning: 577    lwt_unix_stubs.c: No such file or directory                                               
(gdb) bt                                                                                                  
#0  0x00005555561aa77e in lwt_unix_send_notification (domain_id=0, id=2147482650) at lwt_unix_stubs.c:577 
#1  0x00005555561aae7b in execute_job (job=job@entry=0x7fffe78c1410) at lwt_unix_stubs.c:983              
#2  0x00005555561aaf0d in worker_loop (data=<optimized out>) at lwt_unix_stubs.c:1071                     
#3  0x00007ffff7fbf9d2 in start (p=0x7fffe78b9ad8) at src/thread/pthread_create.c:207                     
#4  0x00007ffff7fc1314 in __clone () at src/thread/x86_64/clone.s:22                                      
Backtrace stopped: frame did not save the PC                                                              
(gdb)                                                                                                     
```
In case it's my fault you can find the code in https://github.com/kit-ty-kate/opam-health-check-ng/pull/6

src/direct/lwt_direct.mli
```diff
@@ -0,0 +1,95 @@
...
+    uppercase version to another channel.
+
+    This task is itself a [unit Lwt.t], which is resolved when the function
+    returns. It is possible to use
+    {!Lwt_direct.run_in_the_background} to ignore the result and
+    let the task run in the background instead.
+
+    *)
+
+val spawn : (unit -> 'a) -> 'a Lwt.t
```
kit-ty-kate, line 38: I'm not sure `spawn` is a very understandable name. Maybe `promise` (like javascript, …) or `async` (like rust's tokio or miou) would be a bit better?

src/direct/lwt_direct.mli
```diff
@@ -0,0 +1,95 @@
...
+    promises and use {!await} to wait for them. Like any promise
+    in Lwt, [f ()] can starve the event loop if it runs long computations
+    without yielding to the event loop.
+
+    When [f ()] terminates (successfully or not), the promise
+    [spawn f] is resolved with [f ()]'s result, or the exception
+    raised by [f ()].
+
+    The promise returned by [spawn f] is not cancellable. Canceling it
+    will have no effect. *)
```
kit-ty-kate, line 50: is that the case for everything inside `f` or just for non-lwt code in `f`?

kit-ty-kate: I see, thanks!

PR #1061, opened 2025-07-10
Lwt_io: options for network functions, client and server

c-cube:
The idea is that we should be able to use the high level API and still have
basic functionality such as TCP_NODELAY being set automatically. For more
advanced cases, optional callbacks can be used to customize the socket FDs.

### DIFF
--- a/src/unix/lwt_io.mli
+++ b/src/unix/lwt_io.mli
@@ -572,10 +572,16 @@ val open_connection :
 
       @raise Unix.Unix_error on error.
 
-      @param set_tcp_nodelay if true, [TCP_NODELAY] is set on the socket FD
+      @param set_tcp_nodelay if true, [TCP_NODELAY] is set on the socket FD. This
+        avoids a surprising 40ms delay in some situations.
+        See for example https://brooker.co.za/blog/2024/05/09/nagle.html for why.
 
       @param prepare_fd is a custom callback that can be used to modify the socket FD
-      before it is turned into high level channels
+      before it is turned into high level channels.
+
+      For example passing
+      [~prepare_fd:(fun fd -> Lwt_unix.setsockopt_int fd SO_SNDBUF 65_536)]
+      will set the socket's send buffer's size to 64kiB.
   *)
 
 val with_connection :
@@ -588,7 +594,8 @@ val with_connection :
       connection to the given address and passes the channels to
       [f]
 
-      See {!open_connection} for more details. *)
+      See {!open_connection} for more details about [set_tcp_nodelay]
+      and [prepare_fd]. *)
 
 (**/**)
 
@@ -661,6 +668,9 @@ f client_address client_socket
     The returned promise (a [server Lwt.t]) resolves when the server has just
     started listening on [listen_address]: right after the internal call to
     [listen], and right before the first internal call to [accept].
+
+    See {!open_connection} for more details about [set_tcp_nodelay]
+      and [prepare_fd].
 
     @since 4.1.0 *)
 


c-cube:
I added some tests. I'm not sure why the CI fails on freeBSD.

c-cube:
I'm not sure why the tests fail on freeBSD.

c-cube:
After some thoughts, I think `TCP_NODELAY` should still be set by default. In the super rare case where it must be unset, people can either specify the option or go to `Lwt_unix`; for the rest this will save perplexity over these mysterious 40ms that keep popping up.

dinosaure:
Sorry for arriving late (and after the merge), but it is generally advisable to take a conservative approach when creating a socket and, perhaps, document such an option, which is an optimisation in itself.

The apparent optimisation provided by `TCP_NODELAY` is often quite specific to the fact that the tests/benchmarks are done on a local network. But when it comes to communicating with the Internet (where there are a variety of implementations of the TCP/IP stack), `TCP_NODELAY` can be counterproductive (especially with BSD systems that can send delayed ACKs for small packets). What is certain is that `TCP_NODELAY` changes the behaviour of the sender but not the reception (on the peer side, which may have a different implementation of the TCP/IP stack).

What is certain is that this type of optimisation is quite relative to the network you are on. It is also specific to what we want to do: some would recommend UDP rather than TCP+TCP_NODELAY, for example. This does not systematically invalidate the use of `TCP_NODELAY`, but neither does it systematically confirm the use of `TCP_NODELAY`. This is why the option still exists and why the Nagle algorithm is still applied by default.

I am not against `TCP_NODELAY`, but it would be more prudent to continue offering a socket that behaves like the one we could create with the `Unix` module and to offer, via the documentation and the option, a possible optimisation explaining the implications.

On another note, there is much more documentation on the Nagle algorithm as a bottleneck when writing than on congestion problems because `TCP_NODELAY` has been enabled. For a user, it might be more interesting to encounter one rather than the other (and discover that `lwt` has a subtly different way of creating sockets).

c-cube:
I disagree. Nagle is the wrong default in 99% of use cases, and this still allows the user to disable it. Performance issues caused by its disabling are also easy to find (too many small packets) and fixed by buffering, which Lwt_io already does!

A user who knows they do want Nagle (who? When?) will know to disable the new option. Other users will have fewer mysterious performance issues.

hannesm:
I understand that enabling or not TCP_NODELAY is controversial, as it is as well controversially discussed on the Internet. In the end, some programming languages (such as Go) enable TCP_NODELAY by default, others don't. I haven't seen an OS that enables TCP_NODELAY by default.

For what it is worth, I'd at least expect the default in Lwt_io being documented. At the moment, the docstring is:
``OCaml
      @param set_tcp_nodelay if true, [TCP_NODELAY] is set on the socket FD. This
        avoids a surprising 40ms delay in some situations.
        See for example https://brooker.co.za/blog/2024/05/09/nagle.html for why.
```

It may be worth to refer to (a) the default, (b) maybe Nagle's algorithm (and the nice wikipedia article https://en.wikipedia.org/wiki/Nagle%27s_algorithm -- instead of a random rant on the Internet.

Also, this should then be highlighted in the Changes of a potential new lwt release.

c-cube:
I'm assuming your HTTP server sets it. Mine does. Go does by default. Rust's [axum does](https://github.com/tokio-rs/axum/issues/2521).

Generally this doesn't affect a user of lwt_unix, the low level interface. And for a non expert user of lwt_io I am indeed very confident TCP_NODELAY is the correct default. I had to set it once afain recently when writing a basic TCP server for tests! Congestion, if it happens, happens later and I haven't seen it be as bad as the 40ms tax of forgetting TCP_NODELAY.

c-cube:
> It may be worth to refer to (a) the default, (b) maybe Nagle's algorithm (and the nice wikipedia article https://en.wikipedia.org/wiki/Nagle%27s_algorithm -- instead of a random rant on the Internet.

Started in https://github.com/ocsigen/lwt/pull/1065 , good point!

PR #1075, opened 2025-09-09
Improve CI job names

hhugo:


hhugo:
I've rebased the branch and gave a better name to the 32bit job

PR #1078, opened 2025-09-13
Add back documentation on sequencing PPX

yawaramin:
This was mistakenly removed in #516.

### DIFF
--- a/src/ppx/ppx_lwt.mli
+++ b/src/ppx/ppx_lwt.mli
@@ -51,6 +51,24 @@ code
 
    Not using parentheses will confuse the OCaml parser.
 
+   - sequencing:
+
+   {|<expr>;%lwt|}
+
+   For example:
+
+   {|
+Lwt_io.write file "hello, ";%lwt
+Lwt_io.write_line file "world!"
+  |}
+
+  is expanded to:
+
+  {|
+bind (Lwt_io.write file "hello, ") (fun () ->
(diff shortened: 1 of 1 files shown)


raphael-proust:
thanks

i'll make some changes so that both the ppx and the `Syntax` module are advertised in the manual. i don't think we should remove one to make space for the other: both should be shown.

yawaramin:
Sure. The manual already mentions the let-operators: https://github.com/ocsigen/lwt/blob/f0a4c42e6d00af8e2d5a427859ddef40a3068e41/docs/manual.mld#L456

Maybe that's the place to add more details.

src/core/index.mld
```diff
@@ -15,30 +15,28 @@ Here is a simplistic Lwt program which requests the Google front page, and fails
...
 
   match Lwt_main.run (Lwt.pick [request; timeout]) with
   | Some response -> print_string response
   | None -> prerr_endline "Request timed out"; exit 1
 
-(* ocamlfind opt -package lwt.unix -linkpkg example.ml && ./a.out *)
+(* ocamlfind opt -package lwt.unix,lwt_ppx -linkpkg example.ml && ./a.out *)
 ]}
 
 In the program, functions such as [Lwt_io.write] create promises. The
```
yawaramin, line 42: Also note that some of these examples actually were using the PPX originally, as you can see in the next line:

PR #1087, opened 2025-12-09
Update dune-project with maintenance-intent latest

balat:
This is a request from the opam-repository maintainers.

The ultimate goal is to push package versions that are not maintained
(and thus irrelevant) to a opam-repository-archive - which will make
every client have to perform fewer work, every CI system to compute
less. Apart from CPU and memory savings, there's bandwidth being saved.

policy document
https://github.com/ocaml/opam-repository/blob/master/governance/policies/archiving.md

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,4 +1,4 @@
-(lang dune 3.15)
+(lang dune 3.18)
 
 (name lwt)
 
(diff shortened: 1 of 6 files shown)


raphael-proust:
> Do we want another policy than latest for Lwt?

not sure…

so just to make things clear: `latest` means that for each compiler version from 4.08 to 5.4, we keep the latest lwt package that is installable for that version

and also every lwt package that is needed by any package with a maintenance intent that makes it stay… meaning if a package foo depends on lwt<3.4, and package foo is not archived (typically bc it has a maintenance intent) then lwt.3.3 would be kept

I think `latest` is ok unless someone needs a specific version to stay too??

PR #1096, opened 2026-01-17
eta expand partial applications

avsm:
this is a noop in OCaml, but fixes the build in OxCaml (which infers stack allocation locality without an explicit annotation in the case of partial application)

### DIFF
--- a/src/unix/lwt_unix.cppo.ml
+++ b/src/unix/lwt_unix.cppo.ml
@@ -1549,7 +1549,7 @@ let recv ch buf pos len flags =
   if pos < 0 || len < 0 || pos > Bytes.length buf - len then
     invalid_arg "Lwt_unix.recv"
   else
-    let do_recv = if Sys.win32 then Unix.recv else stub_recv in
+    let do_recv fd buf pos len fl = if Sys.win32 then Unix.recv fd buf pos len fl else stub_recv fd buf pos len fl in
     wrap_syscall Read ch (fun () -> do_recv ch.fd buf pos len flags)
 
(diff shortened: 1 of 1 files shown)


src/unix/lwt_unix.cppo.ml
```diff
@@ -1549,7 +1549,7 @@ let recv ch buf pos len flags =
   if pos < 0 || len < 0 || pos > Bytes.length buf - len then
     invalid_arg "Lwt_unix.recv"
   else
-    let do_recv = if Sys.win32 then Unix.recv else stub_recv in
+    let do_recv fd buf pos len fl = if Sys.win32 then Unix.recv fd buf pos len fl else stub_recv fd buf pos len fl in
     wrap_syscall Read ch (fun () -> do_recv ch.fd buf pos len flags)
```
raphael-proust: if we're eta-expanding i think we should also inline, very little point in defining the intermediary identifier

```suggestion]
    wrap_syscall Read ch (fun () -> if Sys.win32 then Unix.recv ch.fd buf pos len flags else stub_recv ch.fd buf pos len flags)
```

or maybe even better
```
else if Sys.win32 then
  wrap_syscall Read ch (fun () -> Unix.recv ch.fd buf pos len flags)
else
  wrap_syscall Read ch (fun () -> stubs_recv ch.fd buf pos len flags)
```
because it becomes very readable, easy to scan the two lines and spot the difference

raphael-proust (changes requested):

avsm:
I won't have the time to do deeper changes here I'm afraid.

PR #1106, opened 2026-03-13
Prepare 5 10 0

raphael-proust:
This is a version meant to ease the transition to lwt.6. Specifically, for projects which define some custom `Lwt_engine` or handle `Lwt_unix` notification identifiers, they can now write code that's compatible with both 5.9 and 6.


raphael-proust:
not meant for merging into master

should really have a 5.9 branch though…

raphael-proust:
probs need to be `5.10.0` bc it's a breaking change

PR #1107, opened 2026-04-20
Fix pool_mutex initialization order on Windows

raphael-proust:
Move initialize_threading() to run before pool_mutex is accessed in lwt_unix_start_job, rather than only inside the DETACH/SWITCH case.

On Windows, CRITICAL_SECTION must be explicitly initialized via InitializeCriticalSection before use. Unlike pthreads where a zero-initialized mutex (PTHREAD_MUTEX_INITIALIZER) is valid, a zero-initialized CRITICAL_SECTION is invalid. The previous code could access pool_mutex before initialize_threading() was called, causing crashes or undefined behavior on Windows when the first job used async_method != NONE.

initialize_threading() is idempotent (guarded by threading_initialized flag), so calling it earlier is safe and has no effect on Unix platforms.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,9 @@
+===== 6.1.2 =====
+
+====== Fixes ======
+
+  * fix initialisation of lock on Windows. (Mark Elvers, David Allsopp, #1103, #1107, #1104)
+
 ===== 6.1.1 =====
 
 ====== Fixes ======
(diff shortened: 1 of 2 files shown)


raphael-proust:
this is the first commit of https://github.com/ocsigen/lwt/pull/1103 from @mtelvers

PR #1110, opened 2026-06-18
Enhance README with Ocsigen project details

balat:
Updated the README to include project affiliation and documentation links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,6 @@ repo for "Releases only" is also an option.
 
 ## Libraries to use with Lwt
 
-- [eliom](https://ocsigen.org/eliom), [Server](https://ocsigen.org/ocsigenserver) and the full Ocsigen Web suite
 - [alcotest](https://github.com/mirage/alcotest/) —
 unit testing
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -211,3 +213,4 @@ distributed computing
 web framework
 - [lwt_domain](https://github.com/ocsigen/lwt_domain) — domain parallelism when
   using Lwt with OCaml 5
+- ...
```
raphael-proust: the elipsis is a bit weak

maybe something like

```suggestion
- and many other projects, see "Used by" section of the [lwt package page](https://ocaml.org/p/lwt/latest)
```

README.md
```diff
@@ -191,6 +192,7 @@ repo for "Releases only" is also an option.
 
 ## Libraries to use with Lwt
 
+- [eliom](https://ocsigen.org/eliom), [Server](https://ocsigen.org/ocsigenserver) and the full Ocsigen Web suite
```
raphael-proust: I think the list was sorted alphabetically. We should probably keep that.

PR #1114, opened 2026-06-20
doc: list the full Ocsigen family in the hosted table

balat:
The `(hosted …)` table now lists the **whole** Ocsigen family (eliom, toolkit, start, ocsigenserver, js_of_ocaml, tyxml, lwt, reactiveData), not only the projects this doc currently links to. This way, adding a cross-project link later is automatically rewritten for ocsigen.org instead of silently staying an unresolved reference. Same canonical table as eliom/toolkit/start.

### DIFF
--- a/doc/wodoc
+++ b/doc/wodoc
@@ -46,3 +46,17 @@
     (link "Lwt_react" lwt_react/Lwt_react/index.html)
     (link "Lwt_ppx"   lwt_ppx/Lwt_ppx/index.html)
     (link "Lwt_retry" lwt_retry/Lwt_retry/index.html)))
+
+;; Sibling Ocsigen projects, so wodoc rewrites their ocaml.org cross-references
+;; to relative links into their deployed docs on ocsigen.org. The full family is
+;; listed (not only the ones this doc currently links to) so that adding a link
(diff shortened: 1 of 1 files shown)


doc/wodoc
```diff
@@ -46,3 +46,17 @@
...
+;; later never silently breaks. (pkg deploy-dir layout wrapper-module)
+(hosted
+  (eliom           eliom           true  Eliom)
+  (ocsigen-toolkit ocsigen-toolkit true  Ot)
+  (ocsigen-start   ocsigen-start   true  Os)
+  (ocsigenserver   ocsigenserver   false Ocsigen)
+  (js_of_ocaml     js_of_ocaml     subdir)
+  (tyxml           tyxml           subdir)
+  (lwt             lwt             subdir)
+  (reactiveData    reactiveData    root))
```
raphael-proust, line 62: @balat  I'm guessing this info is duplicated on each of those projects so each of their documentation is fixed.

If that's indeed the case, the information probably lives somewhere else. Like in an `ocsigen/projects-documentation` repo or something.

Not urgent to address but worth thinking about for the future.

PR #1118, opened 2026-06-30
ci(doc): cache the opam switch to skip recompiling the dependency tree

balat:
## Problem

The `Documentation` workflow recompiles the dependency tree plus wodoc and odoc from source on **every push** (several minutes); `setup-ocaml@v3` only caches the base switch (the compiler), so none of that was cached.

## Fix

Cache the whole `_opam` switch with `actions/cache`, keyed on the `*.opam` files (the dependency set) and wodoc's resolved HEAD commit. On a cache hit the install step is skipped; the project itself is still rebuilt from the checkout by `dune build @doc` in the doc step, so the docs stay current. The cache rebuilds once when an `*.opam` file or wodoc's HEAD move.

Same change validated on **ocsigen.github.io** (4 min → ~1 min) and **ocsigenserver** (5 min → ~1 min, cache hit confirmed). First run on this branch is a cache miss (populates the cache); the speed-up shows on the next run.

> The `release` job (manual, infrequent) is left as-is.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -34,7 +34,26 @@ jobs:
         with:
           ocaml-compiler: "5.4.0"
 
+      # The bottleneck of this job is compiling the dependency tree plus wodoc and
+      # odoc from source; setup-ocaml only caches the base switch (the compiler),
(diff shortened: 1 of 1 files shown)


raphael-proust:
thanks

Issue #1081, opened 2025-10-10
lwt.5.9.2 doesn't compile anymore on Windows with the Cygwin backend


kit-ty-kate:
Looks like some assumption about the linker were made that doesn't hold on Cygwin.
```
#=== ERROR while compiling lwt.5.9.2 ==========================================#
# context              2.5.0~alpha1~dev | cygwin/x86_64 | ocaml-system.4.14.2 | git+https://github.com/ocaml/opam-repository.git#38a1469dbbc69c770e534e26f79c97256a442b71
# path                 ~/.opam/default/.opam-switch/build/lwt.5.9.2
# command              ~/.opam/default/bin/dune build -p lwt -j 3 @install
# exit-code            1
# env-file             ~/.opam/log/lwt-623-ad3ec6.env
# output-file          ~/.opam/log/lwt-623-ad3ec6.out
### output ###
# Usage:
# [...]
# 
# Notes:
# * The -I, -l and -L options do not need to be separated from their argument.
# * An option like /linkXXX is an abbrevation for '-link XXX'.
# * An option like -Wl,-XXX is an abbreviation for '-link -XXX'.
# * FlexDLL's object files are searched by default in the same directory as
#   flexlink, or in the directory given by the environment variable FLEXDIR
#   if it is defined.
# * Extra argument can be passed in the environment variable FLEXLINKFLAGS.
# 
# Homepage: http://alain.frisch.fr/flexdll.html
```

kit-ty-kate:
The full failure was:
```
+ /home/runneradmin/.opam/default/bin/dune "build" "-p" "lwt" "-j" "3" "@install" (CWD=/home/runneradmin/.opam/default/.opam-switch/build/lwt.5.9.2)
- File "src/unix/dune", lines 52-196, characters 0-3016:
-  52 | (library
-  53 |  (name lwt_unix)
-  54 |  (public_name lwt.unix)
- .....
- 194 |   (:include unix_c_library_flags.sexp))
- 195 |  (instrumentation
- 196 |   (backend bisect_ppx)))
[...]

PR #1119, opened 2026-07-02
fix cygwin properly

raphael-proust:
addressing https://github.com/ocsigen/lwt/issues/1081


src/unix/config/discover.ml
```diff
@@ -261,14 +261,16 @@ struct
...
+    | "Win32" ->
       let unicode = ["-DUNICODE"; "-D_UNICODE"] in
       if Configurator.ocaml_config_var_exn context "ccomp_type" = "msvc" then
         extend unicode ["ws2_32.lib"]
       else
         extend unicode ["-lws2_32"]
-    else
-      extend ["-fPIC"; "-pthread"] []
+    | "Cygwin" -> ()
+    | "Unix" -> extend ["-fPIC"; "-pthread"] []
```
kit-ty-kate: it works fine as it is currently with fpic and pthread in the ccopts on cygwin
```suggestion
    | "Cygwin" | "Unix" -> extend ["-fPIC"; "-pthread"] []
```

PR #1121, opened 2026-07-09
lwt_pool: fix starvation bug

koonwen:
[BUG] Pool waiters starve when in-use members fail and there are no new pool use attempts.

Lwt_pool.use has two failure paths, but only one wakes waiters:
1. Free member fails validate → dispose + replace_disposed (wakes a waiter with a fresh member, or hands it the creation error).
2. In-use member's operation raises → dispose only. No waiter is woken, no replacement created.

Waiters can only be woken by a `release` or `replace_disposed`. If all members are disposed via the failed-use path (2) and there are waiters because of pool oversubscription, they sleep indefinitely.

### DIFF
--- a/test/core/test_lwt_pool.ml
+++ b/test/core/test_lwt_pool.ml
@@ -183,18 +183,15 @@ let suite = suite "lwt_pool" [
     in
     let gen = (fun () -> Lwt.return_unit) in
     let p = Lwt_pool.create 1 ~check:(fun _ is_ok -> is_ok false) gen in
-    let holder = use p (fun _ ->
-      Lwt.bind
-        (Lwt_unix.sleep 1.)
-        (fun () -> failwith "op timeout"))
-    in
+    let yielder, stop_yielding = Lwt.wait () in
+    let holder = use p (fun _ -> Lwt.bind yielder (fun () -> failwith "op timeout")) in
     let waiter = use p (fun _ -> Lwt.return_unit) in
-    let no_starvation = Lwt.bind
-      (Lwt.join [holder; waiter])
-      (fun () -> Lwt.return_true)
-    in
-    let hanging = Lwt.bind (Lwt_unix.sleep 3.0) (fun () -> Lwt.return_false) in
-    Lwt.pick [no_starvation; hanging]
+    let no_starvation = Lwt.join [holder; waiter] in
+    Lwt.wakeup stop_yielding ();
+    Lwt.bind (Lwt.pause ())
+      (fun () -> match Lwt.state no_starvation with
+      | Lwt.Return _ -> Lwt.return_true
+      | Lwt.Fail _ | Lwt.Sleep -> Lwt.return_false)
   end;
 
-  ]
+]
--- a/test/core/test_lwt_pool.ml
+++ b/test/core/test_lwt_pool.ml
@@ -176,4 +176,25 @@ let suite = suite "lwt_pool" [
     Lwt.return (v = 1))
   end;
 
+  test "no starvation for waiters if pool member fails" begin fun () ->
+    let use p f = Lwt.catch
+      (fun () -> Lwt_pool.use p (fun _ -> f ()))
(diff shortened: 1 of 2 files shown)


test/core/test_lwt_pool.ml
```diff
@@ -176,4 +176,25 @@ let suite = suite "lwt_pool" [
...
+  test "no starvation for waiters if pool member fails" begin fun () ->
+    let use p f = Lwt.catch
+      (fun () -> Lwt_pool.use p (fun _ -> f ()))
+      (fun _ -> Lwt.return_unit)
+    in
+    let gen = (fun () -> Lwt.return_unit) in
+    let p = Lwt_pool.create 1 ~check:(fun _ is_ok -> is_ok false) gen in
+    let holder = use p (fun _ ->
+      Lwt.bind
+        (Lwt_unix.sleep 1.)
```
raphael-proust: because of this, the test will take a full second to run. it's hard to find a right amount of time to make sure that even on low-spec hardware with all the tests running concurrently, you have a consistent scheduling.

the trick is to introduce a promise and resolve it at the right time during the test. i cam up with the following:

```
diff --git a/test/core/test_lwt_pool.ml b/test/core/test_lwt_pool.ml
index 1a0ed3e94..eebf82cb4 100644
--- a/test/core/test_lwt_pool.ml
+++ b/test/core/test_lwt_pool.ml
@@ -183,9 +183,10 @@ let suite = suite "lwt_pool" [
     in
     let gen = (fun () -> Lwt.return_unit) in
     let p = Lwt_pool.create 1 ~check:(fun _ is_ok -> is_ok false) gen in
+    let yielder, stop_yielding = Lwt.wait () in
     let holder = use p (fun _ ->
       Lwt.bind
-        (Lwt_unix.sleep 1.)
+        yielder
         (fun () -> failwith "op timeout"))
     in
     let waiter = use p (fun _ -> Lwt.return_unit) in
@@ -193,8 +194,13 @@ let suite = suite "lwt_pool" [
       (Lwt.join [holder; waiter])
       (fun () -> Lwt.return_true)
     in
-    let hanging = Lwt.bind (Lwt_unix.sleep 3.0) (fun () -> Lwt.return_false) in
-    Lwt.pick [no_starvation; hanging]
+    Lwt.wakeup stop_yielding ();
+    Lwt.bind (Lwt.pause ())
+      (fun () -> match Lwt.state no_starvation with
+      | Lwt.Fail _ -> Lwt.return_false
+      | Lwt.Sleep -> Lwt.return_false
+      | Lwt.Return true -> Lwt.return_true
+      | Lwt.Return false -> assert false)
   end;
 
   ]
```

essentially
1. set everything up with a blocking promise that prevents progress
2. resolve the blocking promise
3. wait one `pause` to make sure that the resolve has propagated through
4. check the current state of things

it's a generic recipe for writing tests that would otherwise have some sleep in them

double check that i haven't messed up your test though

raphael-proust (dismissed):

test/core/test_lwt_pool.ml
koonwen: Semantically I want the test to express:

1. `holder` create resource and take pool slot but yield before failure so it occupies the slot
2. `waiter` try to use pool but find it used and put to sleep
3. `holder` unblocked and it's operation fails.

This works if the scheduler always run the promises following the order they are defined in the code. Is this part of the contract?

raphael-proust: defining a promise is the same as starting it. there's no way to define a promise without starting it.

when you want to delay something you have to be explicit:
- `let p = let* () = pause () in <rest of the code here, delayed by one scheduler loop>`
- `let p = let* () = sleep t in <rest of the code, delayed by t>`
- `let p = let* () = w in <rest of the code, delayed by however long it takes for w to resolve>`

--------

in the present case i think that it does what you describe. also the test fails if we remove your fix.

raphael-proust (commented):
looks good now!

PR #1123, opened 2026-07-27
lwt_io: add ?max_connections to establish_server*

### LINKED ISSUE #1111: Lwt_io.establish_server*: max number of connections
(I received by email from a user who doesn't use github)

By now, Lwt_io.establish_server* accept all connections, with no limit.
Such a server will crash (with a Unix_error EMFILE) when accepting one
connection too many. This makes such servers very vulnerable to simple
attacks.

Would it be possible to add an optional parameter to establish_server*
functions to be able to set a limit to current alive connections ?

This parameter could be:
- an integer, and the establish_server* functions would have to count
  current connections and only call Lwt_unix.accept if a new connection
  can be accepted.

- or a function (unit -> unit Lwt.t) which would block until
  Lwt_unix.accept can be called; it would be up to the user code to
  handle counting of connections.

(I have a preference for the second solution, which is more flexible
and will not complexify Lwt_io's code).

rodibozman:
Attempt to fix https://github.com/ocsigen/lwt/issues/1111.

WIP not satisfied with the current code, opened the MR just for diff readability.

### DIFF
--- a/src/unix/lwt_io.ml
+++ b/src/unix/lwt_io.ml
@@ -1671,7 +1671,8 @@ let establish_server_generic
 
       optionally_set_tcp_nodelay set_tcp_nodelay client_socket;
       prepare_client_fd client_socket;
-      connection_handler_callback client_address client_socket;
+      Lwt.async (fun () ->
+        connection_handler_callback client_address client_socket);
 
       accept_loop ()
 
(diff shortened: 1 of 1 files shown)


rodibozman:
Closing for now, just testing stuff, sorry for the noise.

